PR #1, opened 2015-11-27
`logback-access.xml` can be served from outside of jar.

minyk:
Current state, `logback-access.xml` is contained inside of livy-assembly JAR. This commit introduces `livy.server.logback-access.conf` key for specifying `logback-access.xml` from any other locations.


harishreedharan:
In #2 I am removing logback completely, to be compatible with the SparkClient code which is being added.

minyk:
Cool, then I close this PR. :smile:

PR #3, opened 2015-12-15
CDH-35359, CDH-35360. Fix RSC pom, skip tests that require a Spark installation.

vanzin:
Make sure the RSC pom integrates with the rest of the build and uses the
same dependencies; exclude some dependencies that were causing test
failures.

And for tests that need to fork Spark, check whether SPARK_HOME is set,
and skip them when it's not available.

Finally, SparkInterpreter was using internal APIs that were removed in 1.6;
added code to detect which method to use via reflection, although now we
have two cases of using internal APIs...

### DIFF
--- a/livy-core/pom.xml
+++ b/livy-core/pom.xml
@@ -22,7 +22,7 @@
 
     <dependency>
       <groupId>com.cloudera.hue.livy.client</groupId>
-      <artifactId>livy-remote-spark-client</artifactId>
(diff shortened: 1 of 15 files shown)


vanzin:
I tested this both with SPARK_HOME (using a build off of the master Spark branch) and without.

vanzin:
Merged manually.

PR #4, opened 2015-12-15
Rename packages and artifacts.

vanzin:
This is a mechanical change that:

- Removes "hue" from the package names, since it was a needless
  extra path.
- Adds "_2.10" to all artifacts where it was missing, since we need
  that for Scala-based projects.
- Makes sure all groups IDs are "com.cloudera.livy".
- Rename livy-remote-spark-context to livy-client-local.
- Add empty livy-api and livy-client-http modules.


vanzin:
This includes #3 which will be merged separately.

vanzin:
Merged manually.

PR #5, opened 2015-12-16
Add a public api (livy-api).

vanzin:
This PR moves the public API out of livy-client-local and into the
livy-api module, making package name adjustments as necessary. In the process,
the counters feature was dropped (mostly because it relies on accumulators
which are a bad idea for this).

The livy-api supports a service model for instantiating clients; right
now, there are no implementations, but a subsequent change will make
livy-client-local available as a livy-api implementation.


vanzin:
Merged manually.

PR #6, opened 2015-12-17
Refactor livy-client-local on top of livy-api.

vanzin:
This change makes the local client into an implementation of
LivyClientFactory / LivyClient. This allows it to be used solely
by using the public livy-api interfaces (livy-client-local still
needs to be in the classpath, obviously).

A lot of the changes were to un-break the mess that became the
configuration code of the RSC. Now there's a single class used
by everybody. The changes to use the public API instead of the
old SparkClientFactory were very trivial.

Minor build change: maven-compiler-plugin was disabled, since the
Scala plugin also compiles Java classes.


vanzin:
Note this also contains #5 for now; they'll be pushed separately.

The code can use a lot more cleanup, but that can also be done separately. Currently is passes all tests when SPARK_HOME is set.

PR #7, opened 2015-12-17
Cleanup some of the public API.

vanzin:
This change removes some methods from the public API that don't serve
a clear purpose, and can be implemented by users on top of the API if
they really need them.

The whole "MetricsCollection" part kinda falls into that category, but
I decided to leave it in for now, pending more discussions about the
API.

I also cleaned up unchecked warnings and turned on the flag to generate
them during compilation.


vanzin:
Built on top of #6, standard disclaimer applies.

PR #8, opened 2015-12-18
Generate a tarball for the Livy server.

vanzin:
The tarball contains all needed code to run Livy plus a separate
packaging for the local client, used to run the remote Spark
context; those jars are not yet used because we'll need to fix
a bug in the code.

Adjustments in the bin/ scripts were made (including removing
unneeded scripts), and a default logging config was added (to
avoid ugly messages during startup).


vanzin:
As usual, built on patches that I haven't merged yet.

PR #13, opened 2015-12-18
Add config for location of local client jars, don't use "jarOfClass".

vanzin:
This allows Livy client jars to be cached somewhere (e.g. HDFS) instead
of always being uploaded to the cluster. It also removed the use of sketchy
API to figure out what jar to use (since it didn't include dependencies).


vanzin:
Merged manually.

PR #18, opened 2016-01-06
Add support to bypass pre-serialized jobs in client-local.

vanzin:
To implement the HTTP-based protocol for the Spark client API, the
Livy server needs to be able to bypass data serialized in the
client to the remote driver, without trying to deserialize it
in the process, because it doesn't have the necessary client
classes to do that.

This change adds a private api in client-local and some shared
helper code to implement this bypass; the remote driver will
deserialize these jobs as part of executing them. The code
running inside the Livy server becomes a dumb bypass for the
job and its result.

- Removed "Serializable" from API, since it only adds noise when
  using Kryo.
- Added a "client-common" package with shared serde code to implement
  bypass jobs.
- Added "private" API to client-local to send already serialized jobs
  to the remote context.

### DIFF
--- a/api/src/main/java/com/cloudera/livy/Job.java
+++ b/api/src/main/java/com/cloudera/livy/Job.java
@@ -22,7 +22,7 @@ import java.io.Serializable;
 /**
  * Interface for a Spark remote job.
  */
-public interface Job<T extends Serializable> extends Serializable {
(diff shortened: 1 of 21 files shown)


vanzin:
I'll merge this once the tests pass (again) since it's blocking some other stuff I want to send for review.

PR #19, opened 2016-01-09
Add servlet to handle livy client sessions.

vanzin:
This is not complete; it allows creating / tearing down sessions and
submitting synchronous / asynchronous jobs, but there's currently no
way to retrieve a job's results (or even monitoring it's state, since
the backend code for that is not implemented).

Also includes a new base test class for the servlets that does away
with a lot of copy & paste.

### DIFF
--- a/api/src/main/java/com/cloudera/livy/LivyClientBuilder.java
+++ b/api/src/main/java/com/cloudera/livy/LivyClientBuilder.java
@@ -57,6 +57,13 @@ public final class LivyClientBuilder {
     return this;
   }
 
+  public LivyClientBuilder setIfMissing(String key, String value) {
+    if (!config.containsKey(key)) {
+      setConf(key, value);
+    }
+    return this;
+  }
+
   public LivyClient build() {
     ClassLoader cl = Thread.currentThread().getContextClassLoader();
     if (cl == null) {
--- a/server/src/main/scala/com/cloudera/livy/server/client/ClientSessionServlet.scala
+++ b/server/src/main/scala/com/cloudera/livy/server/client/ClientSessionServlet.scala
@@ -1,0 +1,91 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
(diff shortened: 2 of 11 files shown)


vanzin:
This is code I picked up from @harishreedharan, cleaned up, fixed up to match recent changes, and added unit tests. It's not complete; notable, you can't get job results nor monitor their status yet. Also there are some missing unit tests. But the change is already not small, so let's do things in parts.

api/src/main/java/com/cloudera/livy/LivyClientBuilder.java
```diff
@@ -57,6 +57,13 @@ public LivyClientBuilder setAll(Properties props) {
     return this;
   }
 
+  public LivyClientBuilder setIfMissing(String key, String value) {
```
harishreedharan: Not sure how likely it is that two threads would call this method at the same time, we might want to `synchronize` this.

vanzin: This builder is not thread-safe; no point in synchronizing just this method.

server/src/main/scala/com/cloudera/livy/server/client/ClientSessionServlet.scala
```diff
@@ -0,0 +1,91 @@
...
+  post("/:id/submit-job") {
+    sessionManager.get(params("id").toInt) match {
+      case Some(session) =>
+        val req = parsedBody.extract[SerializedJob]
+        if (req.job != null && req.job.length > 0) {
+          Created(session.submitJob(req.job))
+        } else {
+          BadRequest("No job provided.")
+        }
+      case None =>
```
harishreedharan: We should return a `BadRequest` here too, since the session does not exist

vanzin: I'm cleaning this up as part of a separate change.

harishreedharan:
This looks good, but one of the tests timed out (an older test though), not sure if it was due to this PR

vanzin:
That test times our every once in a while. I'll re-run it.

PR #21, opened 2016-01-11
Break RemoteDriver into multiple source files.

vanzin:
RemoteDriver was kind of a mess, and because I'll need to add more features
to it, was becoming harder and harder to change. This change breaks the file
into multiple sources, which doesn't necessarily makes it easier to understand,
but at least makes it easier to add new things and eventually do a proper
cleanup.


vanzin:
I'm gonna merge this since I have changes built on top of it.

PR #25, opened 2016-01-13
Implement client methods to retrieve job status, results.

vanzin:
Use the new APIs in the local-client module to fetch job status and
send results back to the requester. A lot of the changes here are
cleanup so that code doesn't need to be repeated all the time.


vanzin:
This includes PR #22 for now but they'll be pushed separately.

vanzin:
I'm gonna merge this to unblock more work.

PR #27, opened 2016-01-13
Add helper class for config data, clean up unused public API.

vanzin:
I'll need something like LocalConf for the HTTP client also, so
break the useful parts into an abstract class and modify LocalConf
to inherit from it; the new class does not inherit from Hadoop's
Configuration, so some code needed to be reimplemented.

Also removed "clientId()" from the public API since that is not
really important for users. At least not now.


vanzin:
Merging this to unblock further work.

PR #28, opened 2016-01-14
Garbage collect timed-out sessions via SessionManager's logic. Close …

harishreedharan:
…LivyClient when session is closed.


spark/src/main/scala/com/cloudera/livy/spark/client/ClientSession.scala
```diff
@@ -19,31 +19,46 @@ package com.cloudera.livy.spark.client
 import java.net.URI
 import java.nio.ByteBuffer
 import java.util.concurrent
-import java.util.concurrent.{ExecutionException, TimeUnit}
+import java.util.concurrent.TimeUnit
 import java.util.concurrent.atomic.AtomicLong
 
+import org.apache.spark.SparkConf
```
vanzin: We should standardize on one import order... I've been following Spark's style, but the rest of the code is a mess of different styles.

Using Spark's style, this import is in the wrong place.

spark/src/main/scala/com/cloudera/livy/spark/client/ClientSession.scala
```diff
@@ -19,31 +19,46 @@ package com.cloudera.livy.spark.client
...
-class ClientSession(val sessionId: Int, createRequest: CreateClientRequest) extends Session {
+class ClientSession(val sessionId: Int, createRequest: CreateClientRequest)
+  extends Session with Logging {
   implicit val executionContext = ExecutionContext.global
+
   var sessionState: SessionState = SessionState.Starting()
-  SessionClientTracker.createClient(
-    sessionId, createRequest.sparkConf.asJava, createRequest.timeout)
+
+  private val _timeout = createRequest.timeout
```
vanzin: You can declare `timeout` in the parent class as a `val` and avoid the indirection here.

spark/src/main/scala/com/cloudera/livy/spark/client/ClientSession.scala
```diff
@@ -19,31 +19,46 @@ package com.cloudera.livy.spark.client
...
+
   sessionState = SessionState.Running()
 
   private val operations = mutable.Map[Long, java.util.concurrent.Future[_]]()
   private val operationCounter = new AtomicLong(0)
 
-  def getClient(): Option[LocalClient] = {
-    SessionClientTracker.getClient(id).map(_.asInstanceOf[LocalClient])
-  }
+  @volatile private var _lastActivity = System.currentTimeMillis()
```
vanzin: We should really use `System.nanoTime()` for these things.

spark/src/main/scala/com/cloudera/livy/spark/client/ClientSession.scala
```diff
@@ -54,14 +69,17 @@ class ClientSession(val sessionId: Int, createRequest: CreateClientRequest) exte
...
 
   def addFile(uri: URI): Unit = {
-    getClient.foreach(_.addFile(uri))
+    _lastActivity = System.currentTimeMillis()
+    client.addFile(uri)
   }
 
   def addJar(uri: URI): Unit = {
-    getClient.foreach(_.addJar(uri))
+    _lastActivity = System.currentTimeMillis()
```
vanzin: This would probably be cleaner if the parent class instead had a `recordActivity` method.

spark/src/main/scala/com/cloudera/livy/spark/client/ClientSession.scala
harishreedharan: I don't mind - it changes current behavior (session manager uses currentTimeMillis, but we can easily change that too). Will update in next iteration

Tagar: Good point. nano resolution would allow to use this for some sort of profiling too. It's not possible / not accurate enough to do with ms resolution.

vanzin: That's not really the reason for the change; the main difference for this use case is that `currentTimeMillis` is not monotonic.

harishreedharan:
@vanzin Any more comments?

core/src/main/scala/com/cloudera/livy/sessions/Session.scala
```diff
@@ -34,9 +39,15 @@ trait Session {
     }
   }
 
+  val timeout: Long = TimeUnit.MILLISECONDS.toNanos(1000 * 60 * 60)
```
vanzin: You could also say `TimeUnit.HOURS.toNanos(1)`.

core/src/main/scala/com/cloudera/livy/sessions/SessionState.scala
```diff
@@ -61,19 +61,19 @@ object SessionState {
     override def toString = "shutting_down"
   }
 
-  case class Error(time: Long = System.currentTimeMillis()) extends SessionState {
+  case class Error(time: Long = System.nanoTime()) extends SessionState {
```
vanzin: Is `SessionState` exposed to clients? If it is, using `nanoTime` here is kinda useless, since `nanoTime` only makes sense on the machine where it was called.

vanzin:
LGTM aside from the two comments above.

core/src/main/scala/com/cloudera/livy/sessions/SessionState.scala
harishreedharan: The `SessionState` itself is, but the timestamp is not. The `/:id/state` API will return the state's string form, which is basically "started", "dead" etc.

core/src/main/scala/com/cloudera/livy/sessions/Session.scala
harishreedharan: Done

harishreedharan:
Closed via https://github.com/cloudera/livy/commit/4218a47a7cdacfc9684f713a72aac9278e66ed71

PR #31, opened 2016-01-14
Fix license headers.

vanzin:
Everything but the code under api/ and client-local/ is Cloudera code,
so should have a Cloudera header, not the ASF one. Code from those two
modules comes primarily from Hive, so should retain the ASF headers.

I also removed a few files that were not used.


assembly/pom.xml
```diff
@@ -1,4 +1,21 @@
 <?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Licensed to Cloudera, Inc. under one
```
harishreedharan: Why the `~` in all of these xml files?

vanzin: I'm just standardizing the style. Some existing headers have the leading `~`, some don't, I chose to use it.

harishreedharan:
LGTM

PR #34, opened 2016-01-20
Code refactor in preparation for HTTP-based client.

vanzin:
This change includes three different refactorings that will help with
the implementation of the HTTP client for Livy.

- Move declarations of messages expected by the client servlet into
  the client-common package.

This is probably the nastiest change, not because it's controvertial,
but because json4s, which is used heavily by Scalatra, completely
ignores Java types, and in fact has serious issues with them. So a few
hacks are needed on the livy-server module to work around limitations of
json4s. I'd really prefer an alternate API where json4s was completely
bypassed, but that will be left for another day.

As part of that, a few messages were slightly modified or merged with
others, to simplify the handling of the on-the-wire types.

- Move shared code in JobHandleImpl into an abstract class.

A lot of that class needs to be replicated in the HTTP client, so the
parts that could be shared now live in an abstract class in client-common.

- Slight changes to the poms to automatically include scala testing code.

For reasons that will be clear later, the tests for the HTTP client are
written in Scala, even though the library is in Java. The changes in the
poms make more of the boilerplate test setup code shared in the root pom.


vanzin:
I'm going to push this since I have two PRs ready to be sent, built on top of this one.

PR #38, opened 2016-01-22
An HTTP client for Livy Client sessions.

vanzin:
The HTTP client talks to the client servlet in the Livy server
to control client sessions. The library tries to keep a single
connection to the server, and implements a background polling
mechanism to fetch data from the server while implementing the
asynchronous interface of the Livy API.

### DIFF
--- a/client-http/src/main/java/com/cloudera/livy/client/http/HttpClient.java
+++ b/client-http/src/main/java/com/cloudera/livy/client/http/HttpClient.java
@@ -1,0 +1,156 @@
+/*
+ * Licensed to Cloudera, Inc. under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  Cloudera, Inc. licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.cloudera.livy.client.http;
+
+import java.net.URI;
+import java.nio.ByteBuffer;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.Callable;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.Future;
+import java.util.concurrent.ThreadFactory;
+
+import com.cloudera.livy.Job;
+import com.cloudera.livy.JobHandle;
+import com.cloudera.livy.LivyClient;
+import com.cloudera.livy.client.common.BufferUtils;
+import com.cloudera.livy.client.common.Serializer;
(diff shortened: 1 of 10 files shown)


client-http/src/main/java/com/cloudera/livy/client/http/HttpClient.java
```diff
@@ -0,0 +1,156 @@
...
+      executor.shutdownNow();
+      try {
+        conn.delete(Map.class, "/%s", sessionId);
+      } catch (Exception e) {
+        throw propagate(e);
+      } finally {
+        try {
+          conn.close();
+        } catch (Exception e) {
+          // Ignore.
```
harishreedharan: We should log this exception.

client-http/src/main/java/com/cloudera/livy/client/http/HttpClient.java
```diff
@@ -0,0 +1,156 @@
...
+    }
+  }
+
+  @Override
+  public Future<?> addJar(final URI uri) {
+    return addResource("add-jar", uri);
+  }
+
+  @Override
+  public Future<?> addFile(URI uri) {
```
harishreedharan: `uri` can be final here too.

client-http/src/main/java/com/cloudera/livy/client/http/HttpClient.java
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  private <T> JobHandleImpl<T> sendJob(final String command, Job<T> job) {
+    final ByteBuffer serializedJob = serializer.serialize(job);
+    JobHandleImpl<T> handle = new JobHandleImpl<T>(config, conn, executor, serializer);
+    handle.start(sessionId, command, serializedJob);
+    return handle;
+  }
+
+  private RuntimeException propagate(Exception cause) {
```
harishreedharan: There are a couple of other places in the existing codebase where I felt it might be useful to add a `LivyException` - that might be beneficial here too.

client-http/src/main/java/com/cloudera/livy/client/http/HttpClient.java
vanzin: Actually the `final` is leftover from code I deleted, it's not needed.

client-http/src/main/java/com/cloudera/livy/client/http/HttpClient.java
vanzin: It's not a really interesting exception. Since the library doesn't do any logging at all (instead choosing to throw exceptions when errors occur), I'd rather not add it for this.

client-http/src/main/java/com/cloudera/livy/client/http/HttpClient.java
vanzin: Not against adding that, but separate change.

client-http/src/main/java/com/cloudera/livy/client/http/LivyConnection.java
```diff
@@ -0,0 +1,148 @@
...
+      Class<V> retType,
+      String uri,
+      Object... uriParams) throws Exception {
+    req.setURI(new URI(server.getScheme(), null, server.getHost(), server.getPort(),
+      uriRoot + String.format(uri, uriParams), null, null));
+    req.setHeader(HttpHeaders.ACCEPT, APPLICATION_JSON);
+    req.setHeader(HttpHeaders.CONTENT_TYPE, APPLICATION_JSON);
+    req.setHeader(HttpHeaders.CONTENT_ENCODING, "UTF-8");
+
+    CloseableHttpResponse	res = client.execute(req);
```
harishreedharan: Unneeded spaces

harishreedharan:
LGTM

PR #39, opened 2016-01-25
LIVY-39

GayathriMurali:
Modified LocalClient.java to use Spark 1.6 launcher.

### DIFF
--- a/client-local/src/main/java/com/cloudera/livy/client/local/LocalClient.java
+++ b/client-local/src/main/java/com/cloudera/livy/client/local/LocalClient.java
@@ -36,6 +36,7 @@ import java.util.Properties;
 import java.util.UUID;
 import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.HashMap;
 
 import com.google.common.base.Charsets;
 import com.google.common.base.Joiner;
@@ -61,6 +62,8 @@ import com.cloudera.livy.client.local.driver.RemoteDriver;
 import com.cloudera.livy.client.local.rpc.Rpc;
 import com.cloudera.livy.client.local.rpc.RpcServer;
 import static com.cloudera.livy.client.local.LocalConf.Entry.*;
+
+import org.apache.spark.launcher.SparkLauncher;
 
 public class LocalClient implements LivyClient {
   private static final Logger LOG = LoggerFactory.getLogger(LocalClient.class);
@@ -238,10 +241,11 @@ public class LocalClient implements LivyClient {
       String executorJavaOpts = Joiner.on(" ").skipNulls().join(
           osxTestOpts, conf.get(EXECUTOR_OPTS_KEY));
 
-      // Create a file with all the job properties to be read by spark-submit. Change the
+
+      // Create a file with all the job properties to be read by client process. Change the
       // file's permissions so that only the owner can read it. This avoid having the
       // connection secret show up in the child process's command line.
-      File properties = File.createTempFile("spark-submit.", ".properties");
+      File properties = File.createTempFile("client-props.", ".properties");
       if (!properties.setReadable(false) || !properties.setReadable(true, true)) {
         throw new IOException("Cannot change permissions of job properties file.");
       }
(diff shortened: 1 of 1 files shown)


GayathriMurali:
@vanzin  Any update on this?

vanzin:
Hi @GayathriMurali , sorry for the delay (legal can be fun).

Can you write a proper title for the PR? We use that as the commit message. Also, you need to rebase the code since it now has conflicts.

client-local/src/main/java/com/cloudera/livy/client/local/LocalClient.java
```diff
@@ -284,73 +288,66 @@ public void run() {
       // SparkSubmit will take care of that for us.
       String master = conf.get("spark.master");
       Preconditions.checkArgument(master != null, "spark.master is not defined.");
-
-      List<String> argv = Lists.newArrayList();
-
+      //---Modifying argv to be a map
+       HashMap<String,String> argv=new HashMap<String,String>();
```
vanzin: The indentation is wrong here. It's also missing spaces around `=`.

Also, I'm not sure what you're gaining by putting all arguments in a hash map. You're just creating more coupling - instead of just using the methods and constants defined in `SparkLauncher` to identify the options to set, you're now creating a different string representation of each option for no benefit I can see.

client-local/src/main/java/com/cloudera/livy/client/local/LocalClient.java
```diff
@@ -372,18 +369,30 @@ public void run() {
...
-      argv.add("--remote-host");
-      argv.add(serverAddress);
-      argv.add("--remote-port");
-      argv.add(serverPort);
-
-      LOG.info("Running client driver with argv: {}", Joiner.on(" ").join(argv));
-      final Process child = new ProcessBuilder(argv.toArray(new String[argv.size()])).start();
+      //LOG.info("Running client driver with argv: {}", Joiner.on(" ").join(argv));
+      //-----Comment existing livy launcher
+     // final Process child = new ProcessBuilder(argv.toArray(new String[argv.size()])).start();
```
vanzin: Don't leave debug code commented out. Either properly have log statements, or delete these.

client-local/src/main/java/com/cloudera/livy/client/local/LocalClient.java
```diff
@@ -372,18 +369,30 @@ public void run() {
...
+        final SparkLauncher launcher= new SparkLauncher();
```
vanzin: Indentation here is incorrect. Also, missing space before `=`. And, for the following lines, see previous comment about the hash map you're creating. Just create the launcher first thing, and then set things directly in the launcher.

client-local/src/main/java/com/cloudera/livy/client/local/LocalClient.java
```diff
@@ -393,6 +402,7 @@ public void run() {
         @Override
         public void run() {
           try {
+            //modify line below to use spark launcher.waitFor()
```
vanzin: There's no such thing as `SparkLauncher.waitFor()`.

client-local/src/main/java/com/cloudera/livy/client/local/LocalClient.java
```diff
@@ -372,18 +369,30 @@ public void run() {
...
+        launcher.setConf("spark.executor.memory",argv.get("executor-memory"));
+        launcher.setConf("spark.executor.instances",argv.get("num-executors"));
+        launcher.setConf("spark.driver.extraClassPath",argv.get("ClassPath"));
+        launcher.setConf("spark.driver.extraLibraryPath",argv.get("LibraryPath"));
+        for(String filepath : jars) { launcher.addJar(filepath); }
+        launcher.addAppArgs("--properties-file" + properties.getAbsolutePath());
+        launcher.addAppArgs("--remote-host" + serverAddress);
+        launcher.addAppArgs("--remote-port" + serverPort);
+
+        final Process child = launcher.launch();
```
vanzin: This is fine for now, especially since it doesn't require changing the Spark version (this particular API was added in 1.4). But eventually we want to actually move to 1.6 and use `launcher.startApplication()` instead, which gives us better control and more information about the running Spark app.

client-local/src/main/java/com/cloudera/livy/client/local/LocalClient.java
GayathriMurali: I added the HashMap because it was easier to add all parameters in one place along with the launcher creation. Modified  this with the latest PR

vanzin:
I'm gonna close this since you opened a new PR (#79).

PR #40, opened 2016-01-26
Automatically load default configuration from the classpath.

vanzin:
To make it easier to manage client configurations, allow LivyClientBuilder
to load configuration from files present in the classpath, mimicking what
other Hadoop APIs also do. The class will look for two files (livy-client.conf
and spark-defaults.conf) and load their contents if they're present.

On the server side, the ClientSession implementation was changed to
use this instead of trying to load Spark configuration from the system
properties. This should allow server-specific defaults to be configured
while allowing clients to override individual configs using the
session create request.

In the client-local implementation, I broke Spark and Livy configurations
into two separate files, to (arguably) simplify the code a bit. This is
not strictly related to the client configs change; but I also changed the
fact that client-local was loading spark-defaults.conf on its own, and
that's not needed anymore.

### DIFF
--- a/client-local/src/main/java/com/cloudera/livy/client/local/LocalClient.java
+++ b/client-local/src/main/java/com/cloudera/livy/client/local/LocalClient.java
@@ -29,13 +29,17 @@ import java.io.Writer;
 import java.net.URI;
 import java.net.URL;
 import java.nio.ByteBuffer;
+import java.nio.file.Files;
 import java.util.ArrayList;
+import java.util.EnumSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 import java.util.UUID;
 import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicInteger;
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static java.nio.file.attribute.PosixFilePermission.*;
 
 import com.google.common.base.Charsets;
 import com.google.common.base.Joiner;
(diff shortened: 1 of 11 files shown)


client-local/src/main/java/com/cloudera/livy/client/local/LocalClient.java
```diff
@@ -345,7 +302,7 @@ public void run() {
       }
 
       argv.add("--properties-file");
```
harishreedharan: This can be done in a separate change, but I think we should change the name of this config var. `properties-file` is not descriptive enough. We should call it something like `--spark-conf-file` or something

client-local/src/main/java/com/cloudera/livy/client/local/LocalClient.java
```diff
@@ -381,6 +338,8 @@ public void run() {
       argv.add(serverAddress);
       argv.add("--remote-port");
       argv.add(serverPort);
+      argv.add("--config-file");
```
harishreedharan: Perhaps make this `--livy-conf-file` to make it clear that it is Livy conf?

harishreedharan:
Suggested a couple of minor changes, otherwise LGTM.

client-local/src/main/java/com/cloudera/livy/client/local/LocalClient.java
vanzin: `--properties-file` is a spark-submit option. You cannot change it.

client-local/src/main/java/com/cloudera/livy/client/local/LocalClient.java
vanzin: It's a `RemoteDriver` option, so it means "the config file for the RemoteDriver" already. This will all become clearer when we're using the SparkLauncher API...

client-local/src/main/java/com/cloudera/livy/client/local/LocalClient.java
harishreedharan: hmm, never used that one - didn't know! New lessons everyday.

PR #42, opened 2016-01-26
Fix packaging, REPL.

vanzin:
After some directory renaming, the final package was missing some jars.
Also, the REPL was not working after changes were made to stop using
fat jars.

To fix the REPL, I'm using a similar approach to that used by the local
client: copy the needed REPL jars to a repl-specific directory, and add
all jars to the Spark application using "--jars". The list of jars can
be configured so that they can point at cached locations (e.g. local: or
hdfs:).

Since setup-classpath was not useful for the livy-repl script, I inlined
it into livy-server; in fact, the livy-repl script is not used anywhere,
so maybe we could remove it.


harishreedharan:
LGTM

PR #44, opened 2016-01-28
LIVY-15. Added support for API versioning.

tc0312:
- Added a new trait ApiVersioningSupport to provide versioning support for Scalatra Servlet.
- Added an enum to define Livy's API versions.

### DIFF
--- a/server/src/main/scala/com/cloudera/livy/server/ApiVersioningSupport.scala
+++ b/server/src/main/scala/com/cloudera/livy/server/ApiVersioningSupport.scala
@@ -1,0 +1,94 @@
+/*
+ * Licensed to Cloudera, Inc. under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  Cloudera, Inc. licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.cloudera.livy.server
+
+import javax.servlet.http.HttpServletRequest
+
+import org.scalatra.{NotAcceptable, ScalatraBase}
+
+/**
+ * Livy's servlets can mix-in this trait to get API version support.
+ *
+ * Example: {{{
+ * import ApiVersions._
+ * class FooServlet
+ *   ...
+ *   with ApiVersioningSupport
+ *   ...
+ * {
+ *   get("/test") {
+ *     ...
+ *   }
+ *   get("/test", apiVersion <= v0_2) {
+ *     ...
+ *   }
+ *   get("/test", apiVersion <= v0_1) {
+ *     ...
+ *   }
+ * }
+ * }}}
+ */
+trait ApiVersioningSupport extends AbstractApiVersioningSupport {
+  this: ScalatraBase =>
+  // Link the abstract trait to Livy's version enum.
+  override val apiVersions = ApiVersions
+  override type ApiVersionType = ApiVersions.Value
+}
+
+trait AbstractApiVersioningSupport {
+  this: ScalatraBase =>
+  protected val apiVersions: Enumeration
+  protected type ApiVersionType
+
+  /**
+   * Before proceeding with routing, validate the specified API version in the request.
+   * If validation passes, cache the parsed API version as a per-request attribute.
+   */
+  before() {
+    request(AbstractApiVersioningSupport.ApiVersionKey) = request.getHeader("Accept") match {
+      case acceptHeader @ AbstractApiVersioningSupport.AcceptHeaderRegex(apiVersion) =>
+        try {
+            apiVersions.withName(apiVersion).asInstanceOf[ApiVersionType]
+        } catch {
+          case e: NoSuchElementException =>
+            halt(NotAcceptable(e.getMessage))
+        }
+      case _ =>
+        // Return the latest version.
+        apiVersions.apply(apiVersions.maxId - 1).asInstanceOf[ApiVersionType]
+    }
+  }
+
+  /**
+   * @return The specified API version in the request.
+   */
+  def apiVersion: ApiVersionType = {
+    request(AbstractApiVersioningSupport.ApiVersionKey).asInstanceOf[ApiVersionType]
+  }
+
+}
+
+object AbstractApiVersioningSupport {
+  // Get every character after "application/vnd.livy.v" until hitting a + sign.
+  private final val AcceptHeaderRegex = """application/vnd\.livy\.v([^\+]*).*""".r
+
+  // AbstractApiVersioningSupport uses a per-request attribute to store the parsed API version.
+  // This is the key name for the attribute.
+  private final val ApiVersionKey = "apiVersion"
+}
--- a/server/src/test/scala/com/cloudera/livy/server/ApiVersioningSupportSpec.scala
+++ b/server/src/test/scala/com/cloudera/livy/server/ApiVersioningSupportSpec.scala
@@ -1,0 +1,123 @@
+/*
+ * Licensed to Cloudera, Inc. under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  Cloudera, Inc. licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.cloudera.livy.server
+
+import javax.servlet.http.HttpServletResponse
+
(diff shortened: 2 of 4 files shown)


tc0312:
Updated PR after fixing styling issues.

server/src/main/scala/com/cloudera/livy/server/ApiVersioningSupport.scala
```diff
@@ -0,0 +1,98 @@
+package com.cloudera.livy.server
```
vanzin: hi @tc0312,

You're missing copyright headers in some files. I have a pending style check to make sure they're there, but in the meantime can you add them to this PR?

server/src/main/scala/com/cloudera/livy/server/ApiVersioningSupport.scala
```diff
@@ -0,0 +1,98 @@
...
+
+import javax.servlet.http.HttpServletRequest
+
+import org.scalatra.{NotAcceptable, ScalatraBase}
+
+/**
+  * Livy's servlets can mix-in this trait to get API version support.
```
vanzin: There aren't many examples in the code base, but the existing docs use javadoc-style indentation. Would you mind using that? It looks slightly different:

    /**
     * The docs go here.
     */

Wonder if I can write a style checker for that.

server/src/main/scala/com/cloudera/livy/server/ApiVersioningSupport.scala
```diff
@@ -0,0 +1,98 @@
...
+    // Get every character after "application/vnd.livy.v" until hitting a + sign.
+    val acceptHeaderRegex = """application/vnd\.livy\.v([^\+]*).*""".r
+
+    acceptHeader match {
+      case acceptHeaderRegex(v) => Option(v)
+      case _ => None
+    }
+  }
+
+  private def parseForApiVersion(request: HttpServletRequest): ApiVersionType = {
```
vanzin: It took me a while to follow this code and understand what it's doing in which cases... I'd like to suggest a shorter and (to me) clearer version:

    private val acceptHeaderRegex = """application/vnd\.livy\.v([^\+]*).*""".r
    
    private def parseApiVersion(request: HttpServletRequest): Unit = {
      val version = header("Accept") match {
        case Some(accept @ acceptHeaderRegex(v)) =>
          try {
            apiVersions.withName(v).asInstanceOf[ApiVersionType]
          } catch {
            case e: Exception =>
              halt(s"Unknown API version: $accept")
          }
    
        case _ =>
          apiVersions.apply(apiVersions.maxId - 1).asInstanceOf[ApiVersionType]
      }
      request(AbstractApiVersioningSupport.ApiVersionKey) = version
    }

This may or may not compile - I did some basic testing on the Scala shell. But mainly, it would be nice to have a comment saying what's going on; basically, no Accept header means latest version, and so does an accept header that does not have a Livy version in it.

server/src/main/scala/com/cloudera/livy/server/ApiVersioningSupport.scala
```diff
@@ -0,0 +1,98 @@
...
+    } catch {
+      case e: Exception =>
+        halt(NotAcceptable(e.getMessage))
+    }
+  }
+
+  /**
+    * @return The specified API version in the request.
+    */
+  def getApiVersion(): ApiVersionType = {
```
vanzin: Can we call this `def apiVersion: ApiVersionType`? Then the call sites become a little cleaner.

server/src/test/scala/com/cloudera/livy/server/ApiVersioningSupportSpec.scala
```diff
@@ -0,0 +1,103 @@
...
+    shouldReturn("/test", "application/vnd.livy.v0.2+", v0_2.toString)
+    shouldReturn("/test", "application/vnd.livy.v0.1+bar", v0_1.toString)
+    shouldReturn("/test", "application/vnd.livy.v0.2+foo", v0_2.toString)
+    shouldReturn("/test", "application/vnd.livy.v0.1+vnd.livy.v0.2", v0_1.toString)
+    shouldReturn("/test", "application/vnd.livy.v0.2++++++++++++++++", v0_2.toString)
+    shouldReturn("/test", "application/vnd.livy.v1.0", LatestVersionOutput)
+  }
+
+  it("should return error when the specified API version does not exist") {
+    shouldFail("/test", "application/vnd.livy.v", 406)
```
vanzin: minor, but I really like the constants defined in `HttpServletResponse`. No idea what "406" means otherwise.

server/src/main/scala/com/cloudera/livy/server/ApiVersioningSupport.scala
tc0312: I'm using Intellij. That's the default comment syntax it generates.
I fixed it manually but I've some concerns on a coding style that's so strict that they are incompatible with popular IDEs.

vanzin: I don't use IntelliJ, but lots of Spark developers do and the Spark style is to use javadoc-style indents. So there must be a setting somewhere.

server/src/main/scala/com/cloudera/livy/server/ApiVersioningSupport.scala
tc0312: Sorry I'm new to Scala and haven't grasped how powerful pattern matching is.

I fixed the pattern matching. But I want to keep parseApiVersion() purely a parsing function so I left error reporting unchanged and kept `request(AbstractApiVersioningSupport.ApiVersionKey) = version` out of parseApiVersion().

server/src/main/scala/com/cloudera/livy/server/ApiVersioningSupport.scala
tc0312: Fixed

server/src/main/scala/com/cloudera/livy/server/ApiVersioningSupport.scala
```diff
@@ -0,0 +1,101 @@
...
+  /**
+   * @return The specified API version in the request.
+   */
+  def getApiVersion(): ApiVersionType = {
+    request(AbstractApiVersioningSupport.ApiVersionKey).asInstanceOf[ApiVersionType]
+  }
+
+  private def parseApiVersion(request: HttpServletRequest): ApiVersionType = {
+    // Get every character after "application/vnd.livy.v" until hitting a + sign.
+    val acceptHeaderRegex = """application/vnd\.livy\.v([^\+]*).*""".r
```
vanzin: This should be a class field to avoid having to compile the regex on every request.

server/src/main/scala/com/cloudera/livy/server/ApiVersioningSupport.scala
tc0312: Fixed

server/src/test/scala/com/cloudera/livy/server/ApiVersioningSupportSpec.scala
tc0312: Fixed

server/src/main/scala/com/cloudera/livy/server/ApiVersioningSupport.scala
```diff
@@ -0,0 +1,101 @@
...
+    request.getHeader("Accept") match {
+      case acceptHeader @ acceptHeaderRegex(apiVersion) =>
+        try {
+          apiVersions.withName(apiVersion).asInstanceOf[ApiVersionType]
+        } catch {
+          case e: NoSuchElementException =>
+            throw new Exception(s"Unknown API version: $acceptHeader", e)
```
vanzin: So this is why I suggested the other approach... You need quite a lot of extra code just to throw this exception and catch it, while my suggestion does everything with much less code and without using exceptions for control flow.

You just move the `halt()` call here, since that will stop further processing (by throwing an exception, but well, you can't get everything...), and then your "before" implementation is much simpler.

server/src/main/scala/com/cloudera/livy/server/ApiVersioningSupport.scala
tc0312: As long as the comment is in `/** */` Javadoc will parse it just fine. Whitespaces don't matter to the parser.
Unfortunately Intellij doesn't offer an option to switch. It's following the recommended style of Scaladoc. Which is compatible with Javadoc.

server/src/main/scala/com/cloudera/livy/server/ApiVersioningSupport.scala
tc0312: I want to keep parseApiVersion() purely a parsing function.
If we move halt() in parseApiVersion(), parseApiVersion() becomes a part of the control flow.  It makes the code non-linear and harder to follow. Also parseApiVersion() cannot be re-used elsewhere.

vanzin: I guess what I don't understand is why you want that. There's a single call site for that function. The call site, ignoring exception handling, is a single line of code. That screams "inline me!".

If you had unit tests that called `parseApiVersion` directly, I'd sort of understand the desire. But you don't, so your desire to separate concerns just complicates the code for no reason. The whole method could be inlined in the `before()` handler and nothing would change, except you'd have less code, and you wouldn't have this weird contract "let me throw a very generic exception just because the caller is not interested in the exception, but in its message".

tc0312: Merged parseApiVersion() into before().

server/src/main/scala/com/cloudera/livy/server/ApiVersioningSupport.scala
```diff
@@ -0,0 +1,94 @@
...
+ * import ApiVersions._
+ * class FooServlet
+ *   ...
+ *   with ApiVersioningSupport
+ *   ...
+ * {
+ *   get("/test") {
+ *     ...
+ *   }
+ *   get("/test", getApiVersion() <= v0_2) {
```
vanzin: minor: need to update this to match new api (`apiVersion`).

vanzin:
Could you also change the PR title to "LIVY-15. Added support for API versioning."? I have some tooling to automate some administrative tasks when merging PRs and that would help a lot. Thanks!

vanzin:
Thanks, merging this (assuming my merge script works).

PR #45, opened 2016-01-28
Added error handling for startup timeout in repl.

tc0312:
Before this fix, if repl didn't transit from STARTING to another state within 30 seconds, session.waitForStateChange() will throw a TimeoutException. Thus repl will not call server's callback. The exception is swallowed by ExecutionService as it's in a Future. repl will keep running.
As server's callback wasn't called, and repl's YARN process is in RUNNING state, the session is stuck in "STARTING" state on server. If user attempts to delete the session, server would stuck.

With this fix, if repl startup times out, repl will quit with a non zero exit code. YARN will restart repl. If all restart attempts fail, repl's YARN state will transit to FAILED. server will see that and delete won't stuck.

### DIFF
--- a/repl/src/main/scala/com/cloudera/livy/repl/Main.scala
+++ b/repl/src/main/scala/com/cloudera/livy/repl/Main.scala
@@ -127,20 +127,27 @@ class ScalatraBootstrap extends LifeCycle with Logging {
     info(s"Notifying $callbackUrl that we're up")
 
     Future {
-      session.waitForStateChange(SessionState.Starting(), Duration(30, TimeUnit.SECONDS))
+      try {
+        session.waitForStateChange(SessionState.Starting(), Duration(1, TimeUnit.MINUTES))
 
-      // Wait for our url to be discovered.
-      val replUrl = waitForReplUrl()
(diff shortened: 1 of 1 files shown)


tc0312:
I cannot reproduce the test failure locally. Is that test stable?

vanzin:
Yeah that test fails sometimes here, not sure why yet. I restarted the test run.

repl/src/main/scala/com/cloudera/livy/repl/Main.scala
```diff
@@ -127,20 +127,28 @@ class ScalatraBootstrap extends LifeCycle with Logging {
...
+
+        info(s"Calling $callbackUrl...")
+        val rep = Http(req OK as.String)
+        rep.onFailure {
+          case _ => System.exit(1)
+        }
+
+        Await.result(rep, Duration(10, TimeUnit.SECONDS))
+      } catch {
+        case e: Throwable => {
```
vanzin: nit: braces are not necessary

vanzin:
LGTM aside from style nit.

tc0312:
Fixed

PR #46, opened 2016-01-29
Support file/jar upload.

harishreedharan:
- Modified the servlet to add upload support
- Add 2 new APIs: `uploadFile` and `uploadJar` which upload the files from the client to Livy
Todo: tests.

### DIFF
--- a/api/src/main/java/com/cloudera/livy/LivyClient.java
+++ b/api/src/main/java/com/cloudera/livy/LivyClient.java
@@ -17,6 +17,7 @@
 
 package com.cloudera.livy;
 
+import java.io.File;
 import java.net.URI;
 import java.util.concurrent.Future;
 
@@ -58,6 +59,13 @@ public interface LivyClient {
   void stop();
 
   /**
+   * Upload a jar to be added to the Spark application classpath
+   * @param jar The local file to be uploaded
+   * @return A future that can be used to monitor this operation
+   */
+  Future<?> uploadJar(File jar);
+
+  /**
    * Adds a jar file to the running remote context.
    *
    * Note that the URL should be reachable by the Spark driver process. If running the driver
@@ -68,6 +76,13 @@ public interface LivyClient {
    * @return A future that can be used to monitor the operation.
    */
   Future<?> addJar(URI uri);
+
+  /**
+   * Upload a file to be passed to the Spark application
+   * @param file The local file to be uploaded
+   * @return A future that can be used to monitor this operation
+   */
+  Future<?> uploadFile(File file);
 
   /**
    * Adds a file to the running remote context.
--- a/client-http/pom.xml
+++ b/client-http/pom.xml
@@ -69,6 +69,11 @@
       <version>${project.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.apache.httpcomponents</groupId>
+      <artifactId>httpmime</artifactId>
+      <version>4.5.1</version>
+    </dependency>
   </dependencies>
 
   <build>
--- a/client-http/src/main/java/com/cloudera/livy/client/http/LivyConnection.java
+++ b/client-http/src/main/java/com/cloudera/livy/client/http/LivyConnection.java
@@ -18,6 +18,7 @@
 
 package com.cloudera.livy.client.http;
 
+import java.io.File;
 import java.io.IOException;
 import java.net.URI;
 import java.util.concurrent.TimeUnit;
@@ -25,17 +26,17 @@ import java.util.concurrent.TimeUnit;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.apache.http.HttpEntity;
 import org.apache.http.HttpHeaders;
-import org.apache.http.HttpHost;
 import org.apache.http.HttpStatus;
 import org.apache.http.client.config.RequestConfig;
 import org.apache.http.client.methods.CloseableHttpResponse;
 import org.apache.http.client.methods.HttpDelete;
-import org.apache.http.client.methods.HttpEntityEnclosingRequestBase;
 import org.apache.http.client.methods.HttpGet;
 import org.apache.http.client.methods.HttpPost;
 import org.apache.http.client.methods.HttpRequestBase;
 import org.apache.http.client.protocol.HttpClientContext;
 import org.apache.http.entity.ByteArrayEntity;
+import org.apache.http.entity.mime.MultipartEntityBuilder;
+import org.apache.http.entity.mime.content.FileBody;
 import org.apache.http.impl.DefaultConnectionReuseStrategy;
 import org.apache.http.impl.client.CloseableHttpClient;
 import org.apache.http.impl.client.HttpClientBuilder;
@@ -96,11 +97,11 @@ class LivyConnection {
   }
 
   synchronized <V> V delete(Class<V> retType, String uri, Object... uriParams) throws Exception {
-    return sendRequest(new HttpDelete(), retType, uri, uriParams);
+    return sendJSONRequest(new HttpDelete(), retType, uri, uriParams);
   }
 
   synchronized <V> V get(Class<V> retType, String uri, Object... uriParams) throws Exception {
-    return sendRequest(new HttpGet(), retType, uri, uriParams);
+    return sendJSONRequest(new HttpGet(), retType, uri, uriParams);
   }
 
   synchronized <V> V post(
@@ -111,7 +112,31 @@ class LivyConnection {
     HttpPost post = new HttpPost();
     byte[] bodyBytes = mapper.writeValueAsBytes(body);
     post.setEntity(new ByteArrayEntity(bodyBytes));
+    return sendJSONRequest(post, retType, uri, uriParams);
+  }
+
+  synchronized <V> V post(
+      File f,
+      Class<V> retType,
+      String paramName,
+      String uri,
+      Object... uriParams) throws Exception {
+    HttpPost post = new HttpPost();
+    MultipartEntityBuilder builder = MultipartEntityBuilder.create();
+    builder.addPart(paramName, new FileBody(f));
+    post.setEntity(builder.build());
     return sendRequest(post, retType, uri, uriParams);
+  }
+
+  private <V> V sendJSONRequest(
+      HttpRequestBase req,
+      Class<V> retType,
+      String uri,
+      Object... uriParams) throws Exception {
+    req.setHeader(HttpHeaders.ACCEPT, APPLICATION_JSON);
+    req.setHeader(HttpHeaders.CONTENT_TYPE, APPLICATION_JSON);
+    req.setHeader(HttpHeaders.CONTENT_ENCODING, "UTF-8");
+    return sendRequest(req, retType, uri, uriParams);
   }
 
   private <V> V sendRequest(
@@ -121,12 +146,10 @@ class LivyConnection {
       Object... uriParams) throws Exception {
     req.setURI(new URI(server.getScheme(), null, server.getHost(), server.getPort(),
       uriRoot + String.format(uri, uriParams), null, null));
-    req.setHeader(HttpHeaders.ACCEPT, APPLICATION_JSON);
-    req.setHeader(HttpHeaders.CONTENT_TYPE, APPLICATION_JSON);
-    req.setHeader(HttpHeaders.CONTENT_ENCODING, "UTF-8");
+    req.setURI(new URI(server.getScheme(), null, server.getHost(), server.getPort(),
+      uriRoot + String.format(uri, uriParams), null, null));
 
-    CloseableHttpResponse res = client.execute(req);
-    try {
+    try (CloseableHttpResponse res = client.execute(req)) {
       int status = (res.getStatusLine().getStatusCode() / 100) * 100;
       HttpEntity entity = res.getEntity();
       if (status == HttpStatus.SC_OK) {
@@ -140,8 +163,6 @@ class LivyConnection {
         throw new IOException(String.format("%s: %s", res.getStatusLine().getReasonPhrase(),
           error));
       }
-    } finally {
-      res.close();
     }
   }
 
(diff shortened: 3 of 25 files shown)


client-http/src/main/java/com/cloudera/livy/client/http/LivyConnection.java
```diff
@@ -114,6 +117,17 @@ synchronized void close() throws IOException {
     return sendRequest(post, retType, uri, uriParams);
   }
 
+  synchronized <V> V post(
+    File f,
```
vanzin: nit: indent args one extra level so they don't get mixed up with the code.

client-http/src/test/scala/com/cloudera/livy/client/http/HttpClientSpec.scala
```diff
@@ -58,6 +58,8 @@ class HttpClientSpec extends FunSpecLike with BeforeAndAfterAll {
 
   override def beforeAll(): Unit = {
     super.beforeAll()
+    val livyConf = new LivyConf()
```
vanzin: not used?

core/src/main/scala/com/cloudera/livy/sessions/SessionManager.scala
```diff
@@ -41,6 +42,14 @@ class SessionManager[S <: Session, R](livyConf: LivyConf, factory: SessionFactor
   private[this] final val sessionTimeout =
     TimeUnit.MILLISECONDS.toNanos(livyConf.getLong(SessionManager.SESSION_TIMEOUT, 1000 * 60 * 60))
 
+  val livyHome = livyConf.livyHome().getOrElse {
+    if (livyConf.getBoolean("test", false)) {
```
vanzin: I'd rather have an env variable or system property set by the root pom instead of having to manually set this in the test code.

core/src/main/scala/com/cloudera/livy/sessions/SessionManager.scala
```diff
@@ -41,6 +42,14 @@ class SessionManager[S <: Session, R](livyConf: LivyConf, factory: SessionFactor
...
+      "file://" + Files.createTempDirectory("livyTemp").toString
```
vanzin: `Files.createTempDirectory("livyTemp").toURI().toString()`

server/src/main/scala/com/cloudera/livy/server/client/ClientSessionServlet.scala
```diff
@@ -53,17 +60,47 @@ class ClientSessionServlet(sessionManager: SessionManager[ClientSession, CreateC
     }
   }
 
+  jpost[Any]("/:id/upload-jar") { req =>
```
vanzin: nit: the better way to ignore the body here is to do:

    jpost[Unit]("/:id/upload-jar") { _ =>

I probably should make the documentation better at some point... I'm not sure what's really happening when you put `[Any]` in the type parameter.

server/src/main/scala/com/cloudera/livy/server/client/ClientSessionServlet.scala
```diff
@@ -78,4 +115,17 @@ class ClientSessionServlet(sessionManager: SessionManager[ClientSession, CreateC
     new SessionInfo(session.id, session.state.toString)
   }
 
+  private def copyResourceToHDFS(file: FileItem, sessionId: Int): URI = {
```
vanzin: This method is not used?

server/src/test/scala/com/cloudera/livy/server/client/ClientServletSpec.scala
```diff
@@ -21,9 +21,13 @@ package com.cloudera.livy.server.client
 import java.nio.ByteBuffer
 import java.util.HashMap
 
+import org.apache.hadoop.fs.Path
```
vanzin: nit: wrong place

spark/src/main/scala/com/cloudera/livy/spark/SparkManager.scala
```diff
@@ -91,9 +101,25 @@ private class SparkYarnManager(processFactory: SparkProcessBuilderFactory) exten
...
   override def shutdown(): Unit = {
     batchManager.shutdown()
     interactiveManager.shutdown()
     client.close()
   }
-}
\ No newline at end of file
+}
+
+private class SparkClientManager(processFactory: SparkProcessBuilderFactory)
```
vanzin: I'm a little confused about these changes; what do they have to do with file uploads? Why are they needed? Why is `SparkClientManager` inheriting from `SparkYarnManager` since it's not necessarily YARN-related?

server/src/main/scala/com/cloudera/livy/server/client/ClientSessionServlet.scala
harishreedharan: Ah, ok. I was wondering how to test this I guess I will need to set the body to Unit and then somehow set the file attachments.

spark/src/main/scala/com/cloudera/livy/spark/client/ClientSession.scala
```diff
@@ -65,6 +71,14 @@ class ClientSession(val sessionId: Int, createRequest: CreateClientRequest)
     performOperation(job, false)
   }
 
+  def addFile(fileData: Array[Byte], fileName: String): Unit = {
```
vanzin: Why is the argument an `Array[Byte]`? Does that mean that if I try to upload a really large file it will load the whole thing in memory? Or fail? Neither option sounds really good.

spark/src/main/scala/com/cloudera/livy/spark/client/ClientSession.scala
```diff
@@ -80,6 +94,18 @@ class ClientSession(val sessionId: Int, createRequest: CreateClientRequest)
     new JobStatus(id, status.state, status.result, status.error)
   }
 
+  private def copyResourceToHDFS(data: Array[Byte], name: String): URI = {
```
vanzin: So, this method has a few security implications... since we don't have security hooked up yet it's probably ok, but in the long time we need at least a big "TODO: " comment here to remind us.

Basically, with proper security in place, you can't write these files as the Livy user, because then any session could read the files, not just sessions belonging to the user uploading them.

spark/src/main/scala/com/cloudera/livy/spark/client/ClientSession.scala
harishreedharan: I thought this was the only way scalatra supported it, but it looks like we can actually write to a local file and then copy that over to HDFS. I will change this to do that.

spark/src/main/scala/com/cloudera/livy/spark/client/ClientSessionFactory.scala
```diff
@@ -18,16 +18,24 @@
 
 package com.cloudera.livy.spark.client
 
+import org.apache.hadoop.fs.Path
```
vanzin: nit: wrong place

spark/src/main/scala/com/cloudera/livy/spark/SparkManager.scala
harishreedharan: So this is needed to simply be able to create `ClientSession`s. The `SparkManager` type is specified by the configuration and only one of those is loaded by Livy. In this case, we can have 2 different `SparkClientManager`s (one for YARN and one for non-YARN), or add support for the creating `ClientSessionManager`s in both the other `SparkManager`s (which I don't mind either)

spark/src/main/scala/com/cloudera/livy/spark/client/ClientSession.scala
harishreedharan: Will do.

core/src/main/scala/com/cloudera/livy/sessions/SessionManager.scala
harishreedharan: Sounds good, will update this.

harishreedharan:
@vanzin Could you please take a look? I'd like to avoid another rebase.

api/src/main/java/com/cloudera/livy/LivyClient.java
```diff
@@ -58,6 +59,14 @@
   void stop();
 
   /**
+   * Upload a jar to be added to the Spark application classpath
+   * @param f The local file to be uploaded
+   * @return A future that can be used to monitor this operation
+   */
+  Future<?> uploadJar(File jar);
+
```
vanzin: nit: extra blank line

client-http/pom.xml
```diff
@@ -69,6 +69,16 @@
...
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.apache.httpcomponents</groupId>
+      <artifactId>httpmime</artifactId>
+      <version>4.3.4</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.httpcomponents</groupId>
+      <artifactId>httpmime</artifactId>
```
vanzin: duplicate?

client-http/pom.xml
```diff
@@ -69,6 +69,16 @@
       <version>${project.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.apache.httpcomponents</groupId>
+      <artifactId>httpmime</artifactId>
+      <version>4.3.4</version>
```
vanzin: It seems there's a version 4.5.1 which is closer to what we're using above... although I'm really confused about the versioning of this project.

client-http/src/test/scala/com/cloudera/livy/client/http/HttpClientSpec.scala
```diff
@@ -18,11 +18,15 @@
 
 package com.cloudera.livy.client.http
 
+import java.io.{ByteArrayInputStream, File}
 import java.net.{InetAddress, URI}
+import java.nio.file.{Paths, Files}
```
vanzin: hmm, this should have failed the style checker... in fact I don't see the style checker running in your test build, so I wonder what's going on.

harishreedharan: I think you are looking at an outdated diff. The style checker is running - here is one that failed the style check: https://travis-ci.org/cloudera/livy/builds/106647672

Travis results is a bit confusing (it runs `mvn install -DskipTests` and then `mvn clean package` - so the install run's logs are not visible unless the install run fails fail)

harishreedharan: Hmm, so this change is still there - but the style checker seems to miss it (it passes locally too for me). Anyway, fixed it now.

server/src/main/scala/com/cloudera/livy/server/client/ClientSessionServlet.scala
```diff
@@ -29,7 +32,10 @@ import com.cloudera.livy.sessions.SessionManager
 import com.cloudera.livy.spark.client._
 
 class ClientSessionServlet(sessionManager: SessionManager[ClientSession, CreateClientRequest])
-  extends SessionServlet(sessionManager) {
+  extends SessionServlet(sessionManager) with FileUploadSupport {
+
+  configureMultipartHandling(MultipartConfig(maxFileSize = Some(100*1024*1024)))
```
vanzin: Should probably make the max size a configuration option.

server/src/test/scala/com/cloudera/livy/server/client/ClientServletSpec.scala
```diff
@@ -89,6 +108,19 @@ class ClientServletSpec extends BaseSessionServletSpec[ClientSession, CreateClie
...
   }
 
+  private def testResourceUpload(cmd: String, sessionId: Int): Unit = {
+    val f = File.createTempFile("uploadTestFile", cmd)
+
+    Files.write(Paths.get(f.getAbsolutePath), "Test data".getBytes())
+
+    jpost[Unit](s"/$sessionId/upload-$cmd", Map(cmd -> f), 200) { _ =>
+      val resultFile = new File(
+        new URI(sessionManager.livyHome + "/" + sessionId + s"/${f.getName}"))
```
vanzin: use interpolation for the whole string?

spark/src/main/scala/com/cloudera/livy/spark/client/ClientSession.scala
```diff
@@ -80,6 +99,22 @@ class ClientSession(val sessionId: Int, createRequest: CreateClientRequest)
     new JobStatus(id, status.state, status.result, status.error)
   }
 
+
```
vanzin: nit: extra blank line

spark/src/main/scala/com/cloudera/livy/spark/client/ClientSession.scala
```diff
@@ -80,6 +99,22 @@ class ClientSession(val sessionId: Int, createRequest: CreateClientRequest)
...
+  private def copyResourceToHDFS(dataStream: InputStream, name: String): URI = {
+    val filePath = new Path(sessionHome, name)
+    val outFile = fs.create(filePath, true)
+    val buffer = new Array[Byte](512 * 1024)
+    var read = -1
+    try {
+      while ({read = dataStream.read(buffer); read != -1}) {
```
vanzin: I wonder if there's a short NIO-based version to replace this, to avoid the in-memory buffer...

vanzin:
Mostly LGTM, just have minor comments.

spark/src/main/scala/com/cloudera/livy/spark/client/ClientSession.scala
harishreedharan: Couldn't find one - Guava has a [ByteStreams#copy](http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/io/ByteStreams.html#copy(java.io.InputStream,%20java.io.OutputStream)), but that uses a buffer too.

vanzin: Looking at the API it doesn't seem hard to do, but given this is using the `FileSystem` API and it doesn't really expose NIO interfaces, it's probably not going to be any more efficient...

I'm mostly a little worried about the large buffer size.

harishreedharan:
Looks like the last build passed. Any more comments, or shall I merge this?

vanzin:
We can fix issues later; and hopefully we can clean up this manager / factory / factory / another factory thing which I still can't quite follow in the process.

PR #47, opened 2016-01-30
Fixed server and repl to quit if they fail to bind ports.

tc0312:
server.start() will throw an exception when it fails to bind port. It also starts some non-daemon thread even if it throws.
Without calling server.stop(), those non-daemon threads will block JVM shutdown.


harishreedharan:
LGTM

harishreedharan:
Closed via 6ff1188

PR #50, opened 2016-02-02
Add Java and Scala style checkers.

vanzin:
Let's get this out of the way and enforce some style rules during the
build. The rules are mostly borrowed from Spark, with some very minor
tweaks.

Most changes are to break long lines and reorder imports; notable exception
is the renaming of the "scala" repl session package, because some very
wise person involved in defining Scala decided that relative imports were
a good idea.


harishreedharan:
LGTM (+ it looks good to travis, and this does not have any code changes)

PR #53, opened 2016-02-03
Print test output to stdout in Travis runs.

vanzin:
Hopefully can help with debugging flaky tests. Also run the "verify"
target to run the style checker during builds.


vanzin:
Of course now that I want the test to fail it passes on every try. I'll merge this since it will inevitably fail again sometime...

PR #57, opened 2016-02-05
LIVY-34. Sessions API Location Http Header is fixed.

erenavsarogullari:
**JIRA =>** https://issues.cloudera.org/browse/LIVY-34

Sessions API Location Http Header needs to be fixed.

Current Response returns as follows : 
**Location:** /0

Expected Response :
http POST localhost:8998/sessions kind:='"spark"'
HTTP/1.1 201 Created
Content-Length: 69
Content-Type: application/json; charset=UTF-8
Date: Fri, 05 Feb 2016 23:08:25 GMT
**Location:** /sessions/0
Server: Jetty(9.2.10.v20150310)

{
    "id": 0, 
    "kind": "spark", 
    "logs": [], 
    "proxyUser": null, 
    "state": "starting"
}


vanzin:
Hi @erenavsarogullari,

Please check https://github.com/cloudera/livy/wiki/Contributing-to-Livy. Unfortunately we cannot review / merge PRs unless we have you sign a CLA, because Cloudera requires copyright reassignment.

erenavsarogullari:
Hi @vanzin,
Thanks for the quick feedback.
I have just sent the signed CLA. Probably, it will be approved in a couple days.

ksakellis:
@erenavsarogullari Thank you for signing the ICLA. It looks like your branch now has conflicts. Can you please fix the conflicts on this patch?

erenavsarogullari:
Hi @ksakellis, 

Thanks for reply. It should be ok now.

vanzin:
So this PR has a couple of problems.

- you're confounding "sessions", the name of the property in the JSON object that contains the list of active sessions with "sessions", the part of the URI that refers to interactive sessions in the server

- "/sessions" is very specific to interactive sessions (as confusing as that might be). Batches are under "/batches" and clients are under "/clients", so your change is actually incorrect and will not work for batches nor clients.

Instead, it's probably better to use `HttpServletRequest.getPathInfo` to build the full URI. That seems to return the correct information needed to form the full route in my local non-scientific tests.

vanzin:
@erenavsarogullari do you plan on fixing this PR? Otherwise I'd like to close it. thanks!

erenavsarogullari:
Hi @vanzin,
Yes,  will have a look at this weekend and send an update asap. 
Thanks ;)

erenavsarogullari:
Hi @vanzin,
This PR is updated.
Thanks

vanzin:
Ok, merging. Thanks!

PR #59, opened 2016-02-08
LIVY-52. Add SPNEGO authentication, basic authorization.

vanzin:
This change adds support for SPNEGO authentication (a.k.a.
Kerberos) using the hadoop-auth library; it also adds basic
authorization to SessionServlet, so that only owners of
sessions can modify them.

The changes to implement the above are kinda small. To see them,
filter through all the noise by looking at these files directly.

- Main.scala: configure and install the SPNEGO filter.
- SessionServlet.scala: do authorization checks when performing
  actions on sessions.
- SessionServletSpec.scala: test suite for the authorization part.

There were minor changes to the batch and interactive servlets to
use the APIs that do auth checks, instead of the copy & pasted code
they currently were using.

Everything else is noise needed to plumb the owner info through all
the different session factories and types...

### DIFF
--- a/client-http/src/test/scala/com/cloudera/livy/client/http/HttpClientSpec.scala
+++ b/client-http/src/test/scala/com/cloudera/livy/client/http/HttpClientSpec.scala
@@ -262,7 +262,7 @@ private class HttpClientTestBootstrap extends LifeCycle {
 
   override def init(context: ServletContext): Unit = {
     val factory = mock(classOf[SessionFactory[ClientSession, CreateClientRequest]])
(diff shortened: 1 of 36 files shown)


vanzin:
Aside from the unit tests, I tested this on a real cluster and verified SPNEGO auth is working.

Tagar:
Thank you. Does this also address https://issues.cloudera.org/browse/LIVY-44 ? E.g. kerberos tickets renewal?

vanzin:
> Does this also address https://issues.cloudera.org/browse/LIVY-44

No.

PR #60, opened 2016-02-08
LIVY-68. Add SPNEGO authentication, basic authorization.

vanzin:
This change adds support for SPNEGO authentication (a.k.a.
Kerberos) using the hadoop-auth library; it also adds basic
authorization to SessionServlet, so that only owners of
sessions can modify them.

The changes to implement the above are kinda small. To see them,
filter through all the noise by looking at these files directly.

- Main.scala: configure and install the SPNEGO filter.
- SessionServlet.scala: do authorization checks when performing
  actions on sessions.
- SessionServletSpec.scala: test suite for the authorization part.

There were minor changes to the batch and interactive servlets to
use the APIs that do auth checks, instead of the copy & pasted code
they currently were using.

Everything else is noise needed to plumb the owner info through all
the different session factories and types...

### DIFF
--- a/server/pom.xml
+++ b/server/pom.xml
@@ -86,14 +86,12 @@
 
         <dependency>
             <groupId>org.apache.hadoop</groupId>
-            <artifactId>hadoop-common</artifactId>
-            <scope>provided</scope>
+            <artifactId>hadoop-auth</artifactId>
         </dependency>
 
         <dependency>
             <groupId>org.apache.hadoop</groupId>
             <artifactId>hadoop-client</artifactId>
-            <scope>provided</scope>
         </dependency>
 
         <dependency>
--- a/server/src/main/scala/com/cloudera/livy/server/SessionServlet.scala
+++ b/server/src/main/scala/com/cloudera/livy/server/SessionServlet.scala
@@ -18,6 +18,8 @@
 
 package com.cloudera.livy.server
 
+import javax.servlet.http.HttpServletRequest
+
 import scala.concurrent.Future
 import scala.reflect.ClassTag
 
@@ -31,6 +33,9 @@ import com.cloudera.livy.spark.ConfigOptionNotAllowed
 object SessionServlet extends Logging
 
 /**
+ * Base servlet for session management. All helper methods in this class assume that the session
+ * id parameter in the handler's URI is "id".
+ *
  * Type parameters:
  *  S: the session type
  *  R: the type representing the session create parameters.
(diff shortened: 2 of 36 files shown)


server/pom.xml
```diff
@@ -86,14 +86,12 @@
 
         <dependency>
             <groupId>org.apache.hadoop</groupId>
-            <artifactId>hadoop-common</artifactId>
-            <scope>provided</scope>
+            <artifactId>hadoop-auth</artifactId>
```
harishreedharan: Do we want to package the hadoop libraries with Livy?

server/src/main/scala/com/cloudera/livy/server/SessionServlet.scala
```diff
@@ -60,54 +65,45 @@ abstract class SessionServlet[S <: Session, R: ClassTag](sessionManager: Session
...
     )
   }
 
   val getSession = jget("/:id") {
-    val id = params("id").toInt
-
-    sessionManager.get(id) match {
-      case None => NotFound("session not found")
-      case Some(session) => clientSessionView(session)
+    withUnprotectedSession { session =>
```
harishreedharan: Is this required? Do we need to keep this unprotected?

server/src/main/scala/com/cloudera/livy/server/SessionServlet.scala
```diff
@@ -60,54 +65,45 @@ abstract class SessionServlet[S <: Session, R: ClassTag](sessionManager: Session
...
   }
 
   jget("/:id/state") {
-    val id = params("id").toInt
-
-    sessionManager.get(id) match {
-      case None => NotFound("batch not found")
-      case Some(batch) =>
-        Map("id" -> batch.id, "state" -> batch.state.toString)
+    withUnprotectedSession { session =>
```
harishreedharan: Same as above - do we need to keep this unprotected?

server/pom.xml
vanzin: Yes, because Livy actually uses them now.

harishreedharan:
Mostly LGTM, just a couple clarifications

server/src/main/scala/com/cloudera/livy/server/SessionServlet.scala
vanzin: The behavior I decided on was:

- anyone can list sessions and read non-sensitive information about them. Meaning "/" and "/:id" routes don't need to be protected.
- anything that can show sensitive info (like job status) or modify a session needs to be protected.

We could decide that we don't want the first behavior; but then we might need some notion of a "superuser" that is allowed to look at the complete server state.

PR #61, opened 2016-02-09
LIVY-65. Dependency and package cleanup.

vanzin:
A few changes to the build and packaging:

- Drop the "_2.10" prefix from Livy artifacts. The version of Scala does
  not matter for Livy; the server can talk to backends running whatever
  version of Scala (or no Scala at all). So it doesn't make sense to
  expose that.

- Remove Spark classes that leaked through the public API paths; this makes
  sure clients can run without Spark classes in the classpath.

- Clean up the module poms and assembly descriptor a bit.

- Make deployment to a maven repo "opt in", so that we control artifacts that
  should be published.

Note that Scala dependencies are still stuck at 2.10. This means Livy cannot
currently run (or even compile) its repl against a Spark that's using 2.11.
That's a separate future enhancement, though.

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -19,17 +19,18 @@
   <modelVersion>4.0.0</modelVersion>
   <parent>
     <groupId>com.cloudera.livy</groupId>
-    <artifactId>livy-main_2.10</artifactId>
+    <artifactId>livy-main</artifactId>
     <version>0.2.0-SNAPSHOT</version>
   </parent>
 
   <groupId>com.cloudera.livy</groupId>
-  <artifactId>livy-api_2.10</artifactId>
+  <artifactId>livy-api</artifactId>
   <version>0.2.0-SNAPSHOT</version>
   <packaging>jar</packaging>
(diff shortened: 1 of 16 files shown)


harishreedharan:
@vanzin Someone reported a servlet jar issue on the mailing lists. I think it has to do with the YARN dependency. Is it something that we need to worry about? If you think it is, we could add that to this one?

vanzin:
We have some exclusions for servlet-related jars already, maybe we're missing one somewhere. I'll double check.

vanzin:
It doesn't seem like there's any conflict left in the build. There's a possible conflict in case Spark's servlet jar is from a different version; since the repl runs as a Spark app, it will inherit the servlet classes from Spark, and if they're an older version, weird thinks can happen.

But that seems like a separate problem to look at. I'd like the REPL to stop using an embedded Jetty server, but not sure we'll be able to do that for the next release.

vanzin:
I'm gonna merge this.

PR #63, opened 2016-02-09
Fix potential race in local client driver shutdown.

vanzin:
I believe this is the cause of the timeouts that sometimes show up
in travis runs. Looking at logs from previous runs, it seems that
a new SparkContext is being instantiated while the old one is still
shutting down, meaning that both operations can run into each other
(there was an NPE in the logs that seems caused by that).

The reason seems to be that the client side requests a shutdown of
the driver, but doesn't wait until it actually shuts down completely
before tearing down RPC channels; that could cause the client to think
the driver was shut down when it was still in the process of doing it,
among other issues in the code.

The change makes the client wait for the driver to shutdown before
closing the RPC channel, and also changes the shutdown order in the
driver to be a little safer. Also, a small change to silence logs
that aren't really useful.

### DIFF
--- a/client-local/src/main/java/com/cloudera/livy/client/local/LocalClient.java
+++ b/client-local/src/main/java/com/cloudera/livy/client/local/LocalClient.java
@@ -140,23 +140,22 @@ public class LocalClient implements LivyClient {
       isAlive = false;
       try {
         protocol.endSession();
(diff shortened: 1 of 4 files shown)


harishreedharan:
LGTM

PR #64, opened 2016-02-11
LIVY-70. Simplify Livy type and module hierarchy.

vanzin:
Fasten your seatbelts; lots going on here.

The first big change is the removal of the manager / factory / session
hierarchy. Instead, each session servlet knows how to create their own
sessions. There's no need for a "manager" concept anymore, and factories
are all replaced with a single method in each servlet.

This leads to the loss of two existing functions:

- the whole "yarn" module, which wasn't very helpful anyway. All it did
  was monitor the stdout of spark-submit; we should instead switch to
  Spark 1.6 and use the enhanced launcher API for that.

- the Spark config whitelist; that was touted as a security feature, but
  there's really no security if you're running use code as the same
  user as the Livy server. So it didn't really serve any real purpose.

The second big change is the removal of the "spark" module; well, it isn't
really removed, but actually merged with the "server" module which was its
only consumer. This makes development easier since you just need to rebuild
one module when you touch those files. For the same reason, backend-specific
session classes were moved from the "core" module to the "server" module.

There is still more code in "core" that could potentially be cleaned up or
moved, but that's left for the future.

Third but not last, the change removes the concept of absolute or relative
paths, and filesystem root. Clients should always provide paths as they would
for a "spark-submit" invocation; the difference being that local paths
wouldn't really work. There are probably extra enhancements to be made in
this area, but those wouldn't be related to this cleanup.

Finally, some intermediary classes (like "InteractiveWebSession" and
"BaseInteractiveSessionSpec") were merged with their (single) concrete
implementations. That also allowed the removal of the livy-core test
jar.


vanzin:
I'm merging this.

PR #67, opened 2016-02-11
Enable code coverage reports using codecov / jacoco.

vanzin:


harishreedharan:
Since this works, it LGTM

PR #69, opened 2016-02-12
Create a fake SPARK_HOME so tests that fork Spark can run.

vanzin:
This change adds fake "spark-submit" and fake "sparkR" so that tests
can run them as if they were a proper Spark installation. This works
well for Scala / Java, but for python and R it means no pyspark or
SparkR code can actually be tested - just the interaction with the
underlying shell. So a few tests that tried to run Spark code were
removed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ before_install:
 
 before_script:
   - sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys E084DAB9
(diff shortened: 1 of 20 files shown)


vanzin:
Let's see if Travis already gives us R or if I'll need to mess with the config file...

vanzin:
Well, R is not there, now let's figure out how to add it...

vanzin:
Merging this.

vanzin:
Actually, tests failed locally after recent pushes, so will take a second look...

vanzin:
Now merging.

PR #71, opened 2016-02-17
Fix session manager test.

vanzin:
The main cause was that the mock session used by the test was not
overriding the default timeout (1 hour), so the test was failing.

I used the opportunity to clean things up a bit: lastActivity now
always returns something, and activity is recorded as soon as the
session is created. The test also tries a few times to avoid issues
with thread scheduling.

This means that sessions can timeout before they're returned to
clients, but that should be rare (and at the same time, desired,
since sessions should have an upper bound on the time it takes for
them to be created).


vanzin:
I'm gonna merge this since the test is failing for me locally consistently (not sure why it passed on travis).

PR #72, opened 2016-02-18
LIVY-47. Added an override for the scope of Hadoop and Spark dependencies.

tc0312:
In most environment, users already have Hadoop & Spark's jars available on their clusters.
This change added an override so users can exclude them in Livy assembly and provide them to Livy at runtime using CLASSPATH.
To use it, build with: mvn -Dhadoop.scope=provided -Dspark.scope=provided package

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -44,6 +44,7 @@
   <properties>
     <hadoop.version>2.6.0-cdh5.5.0</hadoop.version>
     <spark.version>1.5.0-cdh5.5.0</spark.version>
+    <hadoop.scope>compile</hadoop.scope>
     <commons-codec.version>1.9</commons-codec.version>
     <dispatch.version>0.11.2</dispatch.version>
     <guava.version>14.0.1</guava.version>
@@ -283,6 +284,7 @@
         <groupId>org.apache.hadoop</groupId>
         <artifactId>hadoop-auth</artifactId>
         <version>${hadoop.version}</version>
+        <scope>${hadoop.scope}</scope>
       </dependency>
 
       <dependency>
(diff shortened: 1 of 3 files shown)


vanzin:
Hi @tc0312 ,

If you look at the current final zip file, there are no Spark jars in it. So the changes to the spark dependencies scopes is not necessary.

Regarding the Hadoop dependencies, I don't think this is desired. Livy actually directly depends on Hadoop jars now; so if they're not there, Livy won't work. So I think it's better to leave packaging the way it is.

It might be possible to tweak the dependencies (e.g. Livy currently only uses HDFS but depends on hadoop-client which pulls a lot of extra stuff), if you want to try that.

See commit 52e2cf82 (LIVY-65).

tc0312:
I can remove spark.scope.
I totally understand that this might not fit Cloudera's use case and that's why I made it an override. The default scope is "compile". It doesn't have any side effects unless user specify the override.
In my use case, a single copy of Hadoop jars are provided to different clients using CLASSPATH. Tweaking the dependencies to pull less hadoop jars in mitigates the problem but does not solve it.

tc0312:
I found that without setting spark.scope, following files are included in livy-repl's assembly.
spark-catalyst_2.10-1.6.0.2.4.0.0-149.jar
spark-sql_2.10-1.6.0.2.4.0.0-149.jar
spark-streaming_2.10-1.6.0.2.4.0.0-149.jar
spark-unsafe_2.10-1.6.0.2.4.0.0-149.jar

vanzin:
> In my use case, the source of Hadoop jars for different Hadoop clients are centralized. 

I won't block you if you really think you need this, but bear in mind that we've gone through something similar in Spark and, with time, it turned out to be a huge maintenance nightmare. There's some discussion in SPARK-11157 if you want to read more.

> I found that without setting spark.scope, following files are included in livy-repl's assembly.

What branch are you building? There is no such thing as a "livy-repl assembly" in master.

The only package is the final zip file, and `unzip -l assembly/target/livy-server-0.2.0-SNAPSHOT.zip | grep spark-` returns empty.

tc0312:
We have explored other options. Unfortunately we still need it.

vanzin:
Could you answer my previous question about the repl assembly? I'd like to understand why you're still seeing Spark jars in Livy. That should not happen.

tc0312:
They were under repl-jars directory. However it's no longer a problem in current master.

pom.xml
```diff
@@ -44,6 +44,8 @@
   <properties>
     <hadoop.version>2.6.0-cdh5.5.0</hadoop.version>
     <spark.version>1.5.0-cdh5.5.0</spark.version>
+    <hadoop.scope>compile</hadoop.scope>
+    <spark.scope>compile</spark.scope>
```
vanzin: This is not necessary, could you remove `spark.scope` from this patch?

tc0312: Code doesn't build if I remove these lines. Spark dependencies were not found.

vanzin: You have to remove all the references you've added too, obviously. The code obviously builds without the code you're adding, since all the builds are green.

tc0312: What is the point to check this in if the hadoop & spark scope override is removed?

vanzin: I didn't ask you to remove the hadoop override, just the Spark one, because as our previous conversation shows, it is not needed. There are no Spark jars in the final package.

tc0312: You are right. I removed them.

vanzin:
Ok, LGTM. Merging, thanks!

PR #73, opened 2016-02-19
LIVY-13. Send back plots from the interpreter

purechoc:
send back matplotlib plots imgage from the python interpretor
 - use likes %matplot plt
 - working with Hue master branch
 - plan to support another plot library..

### DIFF
--- a/repl/src/main/resources/fake_shell.py
+++ b/repl/src/main/resources/fake_shell.py
@@ -347,11 +347,14 @@ def magic_matplot(name):
         exc_type, exc_value, tb = sys.exc_info()
         return execute_reply_error(exc_type, exc_value, [])
 
-    fig = value.gcf()
-    imgdata = StringIO.StringIO()
-    fig.savefig(imgdata, format='png')
-    imgdata.seek(0)
-    encode = base64.b64encode(imgdata.buf)
+    try:
+        fig = value.gcf()
+        imgdata = StringIO.StringIO()
+        fig.savefig(imgdata, format='png')
+        imgdata.seek(0)
+        encode = base64.b64encode(imgdata.buf)
+    except:
+        raise ExecutionError(sys.exc_info())
 
     return {
         'image/png': encode,
--- a/repl/src/main/resources/fake_shell.py
+++ b/repl/src/main/resources/fake_shell.py
@@ -354,7 +354,10 @@ def magic_matplot(name):
         imgdata.seek(0)
         encode = base64.b64encode(imgdata.buf)
     except:
-        raise ExecutionError(sys.exc_info())
+        exc_type = Exception
+        return execute_reply_error(exc_type, sys.exc_info, [])
+
+#        raise ExecutionError(sys.exc_info())
 
     return {
         'image/png': encode,
--- a/repl/src/main/resources/fake_shell.py
+++ b/repl/src/main/resources/fake_shell.py
@@ -343,21 +343,15 @@ def magic_json(name):
 def magic_matplot(name):
     try:
         value = global_dict[name]
-    except KeyError:
-        exc_type, exc_value, tb = sys.exc_info()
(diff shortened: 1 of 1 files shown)


vanzin:
Hi @purechoc,

Please check out https://github.com/cloudera/livy/wiki/Contributing-to-Livy. Unfortunately, we can't look at this unless we have a signed CLA from you.

purechoc:
Hi. @vanzin 
I have just sent the signed CLA (cla@cloudera.com)

purechoc:
@vanzin 
thank to attention
I have just sent the signed ICLA to cla@cloudera@.com
and i wait approved.

vanzin:
I'm waiting to hear back from the people who handle CLAs, when that's done we'll take a look. Thanks!

ksakellis:
Hey @purechoc thank you for the ICLA. it looks good.

repl/src/main/resources/fake_shell.py
```diff
@@ -338,6 +340,29 @@ def magic_json(name):
...
+        fig = value.gcf()
+        imgdata = StringIO.StringIO()
+        fig.savefig(imgdata, format='png')
+        imgdata.seek(0)
+        encode = base64.b64encode(imgdata.buf)
+    except:
+        exc_type = Exception
+        return execute_reply_error(exc_type, sys.exc_info, [])
+
+#        raise ExecutionError(sys.exc_info())
```
vanzin: Please don't leave debugging code commented out. Either leave it enabled somehow, or delete it.

repl/src/main/resources/fake_shell.py
```diff
@@ -338,6 +340,29 @@ def magic_json(name):
         'application/json': value,
     }
 
+def magic_matplot(name):
+    try:
+        value = global_dict[name]
+    except KeyError:
```
vanzin: This is not your fault, but this exception handling is just copy & pasted everywhere... it would be much better to just have a single exception handler in `MagicNode`. You don't need to worry about that though.

But in your case, you can have a single `try..except` for the whole method, instead of the current two.

vanzin:
Hey @tc0312, since you guys are more familiar with this part of Livy, would you be willing to take a look at this PR? Thanks!

purechoc:
@vanzin 
thanks to feedback. 
this is updated.

repl/src/main/resources/fake_shell.py
```diff
@@ -338,6 +340,23 @@ def magic_json(name):
...
+        fig.savefig(imgdata, format='png')
+        imgdata.seek(0)
+        encode = base64.b64encode(imgdata.buf)
+    except:
+        exc_type, exc_value, tb = sys.exc_info()
+        return execute_reply_error(exc_type, exc_value, [])
+
+    return {
+        'image/png': encode,
+        'text/plain': "",
```
vanzin: Is this really needed? I'm not really familiar with the magic stuff, but this looks a little odd.

vanzin:
Since it seems there's no more feedback, I have just one last question, but otherwise this is good for merging.

purechoc:
i think this magic stuff help to use pyspark with matplotlib more usefully

vanzin:
That's not my question, though. I'm asking why do you need the `text/plain` part of the response together with the plot (`image/png`).

purechoc:
oh, sorry, i'm miss understood,

because hue notebook code line 155
./desktop/libs/notebook/src/notebook/connectors/spark_shell.py
```
try:
  images = [data['image/png']]
except KeyError:
  images = []

data = [[data['text/plain']]]
meta = [{'name': 'Header', 'type': 'STRING_TYPE', 'comment': ''}]
```
using default key factor 'text/plain'

if there is no key 'text/plain', notebook throw key exception

vanzin:
Sounds good. Merging this.

PR #74, opened 2016-02-19
LIVY-67. Support impersonation in client sessions.

vanzin:
There are a few different things going on here to reach that goal:

- client-local now supports spark submit's --proxy-user command line
  argument, and a new configuration to define the user to be proxied.

- the Livy server now has a list of "super users" who can do anything;
  relevant to this change, super users can impersonate anyone.

- SessionServlet now has a helper method to figure out who should be
  impersonated and check for appropriate permissions.

- The three session servlets now use that method; the semantics are:

  * if there is no user information (i.e. auth is off), users are
    allowed to impersonate anyone
  * when users are authenticated, normal users can only impersonate
    themselves; even if they don't specify a proxy user, the server
    will impersonate the calling user.
  * super users can impersonate anyone; if a specific proxy user is not
    requested, the session will run as the super user.

The rest of the changes are mostly test code, and a small tweak to
JsonServlet to be able to serialize raw String into a JSON object
until LIVY-54 is fixed.


vanzin:
I'm gonna merge this since there's a lot of stuff waiting on it.

PR #76, opened 2016-02-19
Move some classes out of core/.

vanzin:
There are only used in the server module so it's better to have
them there. Also remove "SparkProcess" which did not have any
useful functionality over its parent class.

I also removed the remnants of 'ConfigOptionNotAllowed'; we should
instead document how to deploy Livy securely, and that does not
require blocking users from changing any Spark configuration.


vanzin:
Merging this.

PR #77, opened 2016-02-19
Add support for tar.gz assembly files.

harishreedharan:


assembly/pom.xml
```diff
@@ -81,8 +60,75 @@
...
+              <execution>
+                <id>livy-assembly</id>
+                <phase>package</phase>
+                <goals>
+                  <goal>single</goal>
+                </goals>
+                <configuration>
+                  <finalName>${assembly.name}</finalName>
+                  <formats>
+                    <format>tar.gz</format>
```
vanzin: Instead of all this copy & paste, you could have `tar.gz` be a variable (e.g. `${assembly.format}`) and let the profile override the property (or even leave it for the user to do that on the command line).

That's a much, much smaller change.

assembly/pom.xml
```diff
@@ -33,6 +33,23 @@
...
+  <profiles>
+    <profile>
+      <id>targz</id>
+      <properties>
+        <assembly.format>tar.gz</assembly.format>
+      </properties>
+      <activation><property><name>targz</name></property></activation>
+    </profile>
+    <profile>
+      <id>zip</id>
```
vanzin: Isn't this whole profile the same as declaring `<assembly.format>zip</assembly.format>` in the properties section?

assembly/pom.xml
```diff
@@ -31,8 +31,19 @@
...
+    <assembly.format>zip</assembly.format>
   </properties>
 
+  <profiles>
+    <profile>
+      <id>targz</id>
+      <properties>
+        <assembly.format>tar.gz</assembly.format>
+      </properties>
+      <activation><property><name>targz</name></property></activation>
```
vanzin: This is not needed anymore; I'll remove during merge.

PR #78, opened 2016-02-19
LIVY-59. Remove MetricsCollection from the public API.

vanzin:
This was inherited from Hive, and it's used there; but let's wait
until there's an actual need in Livy for this, especially since
the API is highly coupled with APIs that have changed in Spark,
and applications can still write code themselves to gather metrics
if they need to.


vanzin:
Merging this.

PR #79, opened 2016-02-22
LIVY-39. Migrate Livy to use Spark Launcher Library

GayathriMurali:
Removed Hashmap and modified code for values to be added directly to launcher.Fixed style issues


client-local/src/main/java/com/cloudera/livy/client/local/LocalClient.java
```diff
@@ -84,6 +85,7 @@
   private final Rpc driverRpc;
   private final ClientProtocol protocol;
   private volatile boolean isAlive;
+  final SparkLauncher launcher = new SparkLauncher();
```
vanzin: This doesn't need to be an instance field, since it's only used in one method and then thrown away.

client-local/src/main/java/com/cloudera/livy/client/local/LocalClient.java
```diff
@@ -248,76 +251,60 @@ public void run() {
       // mode, the driver options need to be passed directly on the command line. Otherwise,
       // SparkSubmit will take care of that for us.
       String master = conf.get("spark.master");
+      LOG.info("Master : " + master);
+      launcher.setMaster(master);
```
vanzin: Probably should check that `master` is not null before setting it.

vanzin: I see there's a check in the next line to make sure it's not null, so this code should be added after that check.

client-local/src/main/java/com/cloudera/livy/client/local/LocalClient.java
```diff
@@ -248,76 +251,60 @@ public void run() {
...
         if (master.startsWith("local") ||
               master.startsWith("mesos") ||
               master.endsWith("-client") ||
               master.startsWith("spark")) {
           String mem = conf.get("spark.driver.memory");
-          if (mem != null) {
-            argv.add("-Xms" + mem);
-            argv.add("-Xmx" + mem);
-          }
+          launcher.setConf("spark.driver.memory",mem);
```
vanzin: It would be nice to use the constants defined in the launcher lib (like `SparkLauncher.DRIVER_MEMORY`).

vanzin: Also, if you notice a recent change that I made, in particular this line earlier in the file:

    File sparkConf = writeConfToFile(true);

All the spark configuration is saved to a file and should be sent to Spark by calling `SparkLauncher.setPropertiesFile`. All of this code can then go away, because the launcher code already does all this parsing and translation of options.

client-local/src/main/java/com/cloudera/livy/client/local/LocalClient.java
```diff
@@ -248,76 +251,60 @@ public void run() {
...
-          argv.add(numOfExecutors);
+          launcher.setConf("spark.executor.instances",numOfExecutors);
         }
       }
 
-      argv.add("--properties-file");
-      argv.add(sparkConf.getAbsolutePath());
-      argv.add("--class");
-      argv.add(RemoteDriver.class.getName());
+      launcher.addAppArgs("--properties-file" + sparkConf.getAbsolutePath());
```
vanzin: My fault for telling otherwise earlier, but `sparkConf` contains Spark options and you should use `setPropertiesFile` for it. 

`livyConf` contains Livy options, and you should add app args like this; but you need two separate args, not a single arg concatenated with `+`.

client-local/src/main/java/com/cloudera/livy/client/local/LocalClient.java
```diff
@@ -248,76 +251,60 @@ public void run() {
...
+      launcher.setMainClass(RemoteDriver.class.getName());
+      System.out.println(RemoteDriver.class.getName());
```
vanzin: This log is not really useful.

client-local/src/main/java/com/cloudera/livy/client/local/LocalClient.java
```diff
@@ -339,21 +326,14 @@ public void run() {
...
       }
 
-      argv.add(jar);
-      argv.add("--remote-host");
-      argv.add(serverAddress);
-      argv.add("--remote-port");
-      argv.add(serverPort);
-      argv.add("--config-file");
-      argv.add(livyConf.getAbsolutePath());
+      launcher.addAppArgs("--remote-host" + serverAddress);
```
vanzin: As I mentioned earlier, this is not how you add multiple app args. Use:

    launcher.addAppArgs("--remote-host, serverAddress);

client-local/src/main/java/com/cloudera/livy/client/local/LocalClient.java
```diff
@@ -339,21 +326,14 @@ public void run() {
...
+      launcher.addAppArgs("--remote-port" + serverPort);
+      launcher.addAppArgs("--config-file" + livyConf.getAbsolutePath());
+      launcher.setAppName("ClientID");
```
vanzin: The previous code did not set an app name, so the new one shouldn't either, especially not a hardcoded one that users cannot override.

vanzin:
Do you have access to the logs from the travis builds? You can see the error there. Here's part of the exception:

    java.lang.IllegalArgumentException: 'value' must not be null.^M
            at org.apache.spark.launcher.CommandBuilderUtils.checkNotNull(CommandBuilderUtils.java:235)^M
            at org.apache.spark.launcher.SparkLauncher.setConf(SparkLauncher.java:125)^M
            at com.cloudera.livy.client.local.LocalClient.startDriver(LocalClient.java:263)^M
            at com.cloudera.livy.client.local.LocalClient.<init>(LocalClient.java:98)^M

GayathriMurali:
@vanzin All checks passed

client-local/src/main/java/com/cloudera/livy/client/local/LocalClient.java
vanzin: You haven't addressed this comment.

client-local/src/main/java/com/cloudera/livy/client/local/LocalClient.java
```diff
@@ -248,76 +251,61 @@ public void run() {
       // mode, the driver options need to be passed directly on the command line. Otherwise,
       // SparkSubmit will take care of that for us.
       String master = conf.get("spark.master");
+      LOG.info("Master : " + master);
```
vanzin: This log is not particularly interesting.

client-local/src/main/java/com/cloudera/livy/client/local/LocalClient.java
```diff
@@ -248,76 +251,61 @@ public void run() {
...
-        argv.add(new File(sparkHome, "bin/spark-submit").getAbsolutePath());
-      } else {
-        LOG.info("No spark.home provided, calling SparkSubmit directly.");
-        argv.add(new File(System.getProperty("java.home"), "bin/java").getAbsolutePath());
-
         if (master.startsWith("local") ||
               master.startsWith("mesos") ||
               master.endsWith("-client") ||
               master.startsWith("spark")) {
+          launcher.setMaster(master);
```
vanzin: Why is this inside the condition?

If your master is "yarn" or "yarn-cluster", for example, you will not set a master and things will not work the way the user expects.

client-local/src/main/java/com/cloudera/livy/client/local/LocalClient.java
```diff
@@ -248,76 +251,61 @@ public void run() {
...
           String mem = conf.get("spark.driver.memory");
-          if (mem != null) {
-            argv.add("-Xms" + mem);
-            argv.add("-Xmx" + mem);
+          if(mem != null){
```
vanzin: As I mentioned previously, you don't need to manually set these things. The Spark launcher lib will take care of these as long as they are in the config file you pass in `setPropertiesFile`.

client-local/src/main/java/com/cloudera/livy/client/local/LocalClient.java
```diff
@@ -248,76 +251,61 @@ public void run() {
...
-          argv.add(numOfExecutors);
+          launcher.setConf("spark.executor.instances",numOfExecutors);
         }
       }
 
-      argv.add("--properties-file");
-      argv.add(sparkConf.getAbsolutePath());
-      argv.add("--class");
-      argv.add(RemoteDriver.class.getName());
+      launcher.setPropertiesFile(sparkConf.getAbsolutePath());
```
vanzin: Just re-iterating: this is the *only* thing you need. All the other `setConf` calls can be removed.

client-local/src/main/java/com/cloudera/livy/client/local/LocalClient.java
```diff
@@ -505,7 +485,7 @@ private void handle(ChannelHandlerContext ctx, JobResult msg) {
       JobHandleImpl<?> handle = jobs.remove(msg.id);
       if (handle != null) {
         LOG.info("Received result for {}", msg.id);
-        // TODO: need a better exception for this.
+        // TODO: need a better exception for this
```
vanzin: Change is not necessary and actually was more correct before.

client-local/src/main/java/com/cloudera/livy/client/local/LocalClient.java
GayathriMurali: Are you saying that I can remove all setConf calls? which would also make all the code, where we get  values like `String executorCores = conf.get("spark.executor.cores")` redundant, is that right?

vanzin: Correct. That's all just dead code now.

GayathriMurali: I get it. I will cleanup all that code

client-local/src/main/java/com/cloudera/livy/client/local/LocalClient.java
```diff
@@ -75,6 +76,7 @@
   private static final String SPARK_HOME_KEY = "spark.home";
   private static final String DRIVER_OPTS_KEY = "spark.driver.extraJavaOptions";
   private static final String EXECUTOR_OPTS_KEY = "spark.executor.extraJavaOptions";
+  private static final SparkLauncher launcher = new SparkLauncher();
```
vanzin: Again: this should not be a class field. It should be a local variable in the only method where it is used.

client-local/src/main/java/com/cloudera/livy/client/local/LocalClient.java
```diff
@@ -249,75 +252,12 @@ public void run() {
...
-      argv.add(sparkConf.getAbsolutePath());
-      argv.add("--class");
-      argv.add(RemoteDriver.class.getName());
+      launcher.setMaster(master);
+      launcher.setPropertiesFile(sparkConf.getAbsolutePath());
+      launcher.setMainClass(RemoteDriver.class.getName());
 
       String jar = "spark-internal";
+      launcher.setAppResource(jar);
       String livyJars = conf.get(LIVY_JARS);
```
vanzin: Not your fault, but this whole block (down to L282) should be moved before L247, where the configuration is written out. Otherwise it won't take effect in certain cases...

GayathriMurali:
@vanzin Is there anything I need to do get this PR merged? Please let me know

vanzin:
You need to merge with master since there are conflicts.

GayathriMurali:
@vanzin Cancel client '2507748e-fd5a-453f-a18a-301d0aba4a63'. Error: Child process exited before connecting back - This occurs after I merged your recent changes with the existing code. I got this same error when driver memory was  null. Now all that code is removed and just confFile is added to setProperties.

vanzin:
Exception in thread "main" java.lang.IllegalArgumentException: Invalid command line: --remote-host testing-worker-linux-docker-0b426af2-3376-linux-8 --remote-port 52334 --proxy-user __proxy__
      at com.cloudera.livy.client.local.driver.RemoteDriver.<init>(RemoteDriver.java:118)
      at com.cloudera.livy.client.local.driver.RemoteDriver.main(RemoteDriver.java:250)
      at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
      at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
      at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
      at java.lang.reflect.Method.invoke(Method.java:606)
      at org.apache.spark.deploy.SparkSubmit$.org$apache$spark$deploy$SparkSubmit$$runMain(SparkSubmit.scala:672)
      at org.apache.spark.deploy.SparkSubmit$.doRunMain$1(SparkSubmit.scala:180)
      at org.apache.spark.deploy.SparkSubmit$.submit(SparkSubmit.scala:205)
      at org.apache.spark.deploy.SparkSubmit$.main(SparkSubmit.scala:120)
      at org.apache.spark.deploy.SparkSubmit.main(SparkSubmit.scala)

Seems to me like `--proxy-user` is being added in the wrong spot.

GayathriMurali:
@vanzin I tried adding before setting serverAddress and Port as well. All scenarios same error when tested locally.

vanzin:
That's because `--proxy-user` is a Spark argument, not a RemoteDriver argument...

GayathriMurali:
@vanzin all tests passed

client-local/src/main/java/com/cloudera/livy/client/local/LocalClient.java
```diff
@@ -198,7 +199,7 @@ private Thread startDriver(final RpcServer rpcServer, final String clientId, fin
     Runnable runnable;
     final String serverAddress = rpcServer.getAddress();
     final String serverPort = String.valueOf(rpcServer.getPort());
-
+    final SparkLauncher launcher = new SparkLauncher();
```
vanzin: This is only used inside the `else` in L230, so it's better to declare it there.

client-local/src/main/java/com/cloudera/livy/client/local/LocalClient.java
```diff
@@ -237,10 +238,33 @@ public void run() {
       if (sparkHome == null) {
         sparkHome = System.getProperty(SPARK_HOME_KEY);
       }
+      launcher.setSparkHome(sparkHome);
 
       conf.set(CLIENT_ID, clientId);
       conf.set(CLIENT_SECRET, secret);
 
+       String jar = "spark-internal";
+       launcher.setAppResource(jar);
```
vanzin: You could just inline "spark-internal" here and avoid the previous statement.

client-local/src/main/java/com/cloudera/livy/client/local/LocalClient.java
```diff
@@ -252,115 +276,16 @@ public void run() {
...
-
-      argv.add(jar);
-      argv.add("--remote-host");
-      argv.add(serverAddress);
-      argv.add("--remote-port");
-      argv.add(serverPort);
-
-      LOG.info("Running client driver with argv: {}", Joiner.on(" ").join(argv));
-      final Process child = new ProcessBuilder(argv.toArray(new String[argv.size()])).start();
+          launcher.addSparkArg("--proxy-user", conf.get(PROXY_USER));
```
vanzin: This is indented too far; we use 2 spaces.

client-local/src/main/java/com/cloudera/livy/client/local/LocalClient.java
```diff
@@ -252,115 +276,16 @@ public void run() {
...
+     }
```
vanzin: This line is not indented correctly.

vanzin:
Just minor things remaining, otherwise looks good.

vanzin:
Merging, thanks!

PR #80, opened 2016-02-23
LIVY-66. Allow http client to authenticate using SPNEGO.

vanzin:
The httpclient documentation is kinda off, but this seems to work.
The default value for "RequestConfig.isAuthenticationEnabled" is
actually false (docs say it's "true"), and the SPNEGO auth handler
is installed by default in client instances, so the only thing left
is to register the dummy credentials provider.

Note this requires that the user logs in through some other means
(either using kinit or by performing a JAAS login in their application).

### DIFF
--- a/client-http/src/main/java/com/cloudera/livy/client/http/LivyConnection.java
+++ b/client-http/src/main/java/com/cloudera/livy/client/http/LivyConnection.java
@@ -21,12 +21,16 @@ package com.cloudera.livy.client.http;
 import java.io.File;
 import java.io.IOException;
 import java.net.URI;
+import java.security.Principal;
 import java.util.concurrent.TimeUnit;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.apache.http.HttpEntity;
 import org.apache.http.HttpHeaders;
 import org.apache.http.HttpStatus;
+import org.apache.http.auth.AuthScope;
+import org.apache.http.auth.Credentials;
+import org.apache.http.client.CredentialsProvider;
 import org.apache.http.client.config.RequestConfig;
 import org.apache.http.client.methods.CloseableHttpResponse;
 import org.apache.http.client.methods.HttpDelete;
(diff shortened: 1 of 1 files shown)


client-http/src/main/java/com/cloudera/livy/client/http/LivyConnection.java
```diff
@@ -75,8 +80,32 @@ public int getConnectTimeout() {
...
+
+      @Override
+      public Principal getUserPrincipal() {
+        return null;
+      }
     };
 
+    // This is needed to get Kerberos credentials from the environment, instead of
+    // requiring the application to manually obtain the credentials.
+    System.setProperty("javax.security.auth.useSubjectCredsOnly", "false");
```
harishreedharan: So this would actually pass the credentials on to the HTTPClient?

client-http/src/main/java/com/cloudera/livy/client/http/LivyConnection.java
```diff
@@ -85,6 +114,7 @@ public int getSocketTimeout() {
       .setConnectionReuseStrategy(new DefaultConnectionReuseStrategy())
       .setDefaultRequestConfig(reqConfig)
       .setMaxConnTotal(1)
+      .setDefaultCredentialsProvider(credsProvider)
```
harishreedharan: Is this to get the non-secure case to work?

client-http/src/main/java/com/cloudera/livy/client/http/LivyConnection.java
vanzin: The HTTPClient uses JAAS (Java's auth framework) to get credentials. This tells JAAS that it's ok to read credentials from the environment (e.g. the ticket cache in `/tmp/krb5cc_0`).

client-http/src/main/java/com/cloudera/livy/client/http/LivyConnection.java
vanzin: No, this is for kerberos. It's described in the httpclient guide for SPNEGO (which basically just tells you to try to make the example they have in their repo work).

harishreedharan:
LGTM, merging

PR #84, opened 2016-02-29
LIVY-75. Allow a whitelist of URIs to bypass authentication.

vanzin:
At least until we get to LIVY-46, the repl code needs to POST
results back to the Livy server. Since it doesn't run with
proper kerberos credentials (it's running as a proxy user in
a YARN container), allow the callback URI to bypass SPNEGO.
This is achieved by wrapping the auth filter in another filter
that will not do auth when certain URIs are requested.

While this might be seen as making the repl less secure (which
is true), the repl is currently not secure anyway, since it runs
its own web server and doesn't really perform authorization when
requests arrive.


vanzin:
Closing this since #94 makes it obsolete.

PR #86, opened 2016-03-08
Add explicit dependency to spark-launcher in client-local.

vanzin:
The code now needs that library to run, so it needs to be packaged
with Livy. It's not needed on the driver side of the RSC, so avoid
packaging it under the client-jars directory.

### DIFF
--- a/client-local/pom.xml
+++ b/client-local/pom.xml
@@ -58,6 +58,10 @@
       <groupId>io.netty</groupId>
       <artifactId>netty-all</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.apache.spark</groupId>
+      <artifactId>spark-launcher_${scala.binary.version}</artifactId>
+    </dependency>
 
     <!-- Needed by Spark but excluded in the Livy root pom. -->
     <dependency>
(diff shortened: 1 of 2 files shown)


tc0312:
It's not about the size, but more about importing spark jars from another directory. In our cluster setup, al Spark jars are put into a separate directory. When a Spark patch is installed, jars in that directory will be replaced. We want to keep all Spark jars in a central location to keep patching simple.

vanzin:
This is not a "spark jar" in the sense of providing any spark functionality. It's *just* the spark launcher library. Patching spark doesn't change any functionality in this jar, since the jar basically just runs spark-submit.

If you really don't want this in you distribution, I suggest keeping a patch in your branch. But it really does not make any sense to exclude this from Livy packaging.

tc0312:
I'm not asking for excluding this from Livy packaging. I'm asking for an override so people who want that behavior can do it without custom patching.
Keeping a patch in your branch will eventually lead to forking. We are trying our best to avoid that.

vanzin:
>  I'm asking for an override so people who want that behavior can do it without custom patching.

And that's what I'm saying I want to avoid, because it leads to messy things like we have with Spark. I was against the hadoop override already, but I let it slide because for Hadoop it's a little easier to get those dependencies later (using `hadoop classpath`), even if that's not ideal. But if you start asking about more and more overrides, I'll have to kindly ask you to make those changes in your own branch, because it's very particular to your own distribution of Livy.

vanzin:
Merging this one.

PR #87, opened 2016-03-09
LIVY-88. Replace net.databinder.dispatch with com.ning.async-http-client

zjffdu:


vanzin:
Your change is not passing style checks...

harishreedharan:
I agree. Btw, LIVY-46 code is here: https://github.com/cloudera/livy/tree/repl-rsc

zjffdu:
Update the patch by using "--amend" by accident :(

repl/src/main/scala/com/cloudera/livy/repl/Main.scala
```diff
@@ -135,17 +133,17 @@ class ScalatraBootstrap extends LifeCycle with Logging {
...
         val replUrl = waitForReplUrl()
-
-        var req = url(callbackUrl).setContentType("application/json", "UTF-8")
-        req = req << write(Map("url" -> replUrl))
-
         info(s"Calling $callbackUrl...")
-        val rep = Http(req OK as.String)
-        rep.onFailure {
-          case _ => System.exit(1)
+        val response = new AsyncHttpClient().preparePost(callbackUrl)
```
vanzin: Shouldn't you keep the reference to the new http client and close it in a finally block?

repl/src/main/scala/com/cloudera/livy/repl/Main.scala
```diff
@@ -22,14 +22,14 @@ import java.util.concurrent.TimeUnit
...
+import scala.concurrent.{ExecutionContext, Future}
 import scala.concurrent.duration._
 
-import dispatch._
+import com.ning.http.client.AsyncHttpClient
 import org.json4s.{DefaultFormats, Formats}
 import org.json4s.jackson.Serialization.write
-import org.scalatra.LifeCycle
 import org.scalatra.servlet.ScalatraListener
+import org.scalatra.LifeCycle
```
vanzin: nit: this was correct in the previous location; the style checker doesn't catch that because of a bug (that I've fixed but hasn't been released yet).

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
```diff
@@ -167,19 +167,21 @@ class InteractiveSession(
...
                 _state = SessionState.Dead()
               }
-
               Future.successful(())
-
-            case Left(t: Throwable) =>
-              Future.failed(t)
+            } catch {
+              case e: ConnectException =>
+                synchronized {
```
vanzin: It shouldn't be needed to synchronize here; the synchronization in the rest of this class is kinda spotty already, it would be better to fix all of it in a separate change.

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
```diff
@@ -167,19 +167,21 @@ class InteractiveSession(
...
         case SessionState.Idle() =>
           _state = SessionState.Busy()
-
-          Http(svc.DELETE OK as.String).either() match {
-            case (Right(_) | Left(_: ConnectException)) =>
-              // Make sure to eat any connection errors because the repl shut down before it sent
-              // out an OK.
+          Future {
+            try {
+              new AsyncHttpClient().prepareDelete(url.get.toString).execute()
```
vanzin: Same thing with closing the client instance.

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
```diff
@@ -243,12 +245,15 @@ class InteractiveSession(
       _state = SessionState.Busy()
       recordActivity()
 
-      val req = (svc / "execute").setContentType("application/json", "UTF-8") << write(content)
-
-      val future = Http(req OK as.json4s.Json).map { case resp: JValue =>
-        parseResponse(resp).getOrElse {
+      val future = Future {
+        val response = new AsyncHttpClient().preparePost(url.get.toString + "/execute")
```
vanzin: same re: closing http client instance.

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
```diff
@@ -305,10 +305,10 @@ class InteractiveSession(
   }
 
   private def replErroredOut() = {
-    val req = svc.setContentType("application/json", "UTF-8")
-    val response = Await.result(Http(req OK as.json4s.Json), Duration.Inf)
-
-    response \ "state" match {
+    val response = new AsyncHttpClient().prepareGet(url.get.toString)
```
vanzin: same re: closing http client.

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
```diff
@@ -266,17 +271,12 @@ class InteractiveSession(
...
-    dispatch.url(url.toString)
-  }
-
   @tailrec
   private def waitForStatement(id: Int): JValue = {
-    val req = (svc / "history" / id).setContentType("application/json", "UTF-8")
-    val resp = Await.result(Http(req OK as.json4s.Json), Duration.Inf)
-
-    parseResponse(resp) match {
+    val response = new AsyncHttpClient().prepareGet(url.get.toString + "/history/" + id)
```
vanzin: same re: closing http client.

server/src/main/scala/com/cloudera/livy/utils/Json.scala
```diff
@@ -0,0 +1,30 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.cloudera.livy.utils
+
+import com.ning.http.client.Response
+import org.json4s._
+import org.json4s.jackson.JsonMethods._
+
```
vanzin: nit: too many empty lines

server/src/main/scala/com/cloudera/livy/utils/Json.scala
```diff
@@ -0,0 +1,30 @@
...
+
+object Json extends (Response => JValue) {
+  def apply(r: Response): JValue =
+    (((r: Response) => r.getResponseBody) andThen (s => parse(StringInput(s), true)))(r)
```
vanzin: This is a little hard to read: isn't it the same as:

    parse(StringInput(r.getResponseBody(), true))

Or some other simpler code like that? Also, a whole object just for this one method seems a little overkill; it seems it's just used in one class so it could be a private method? (I might be wrong though.)

zjffdu:
Put usingResource (for closing connection) and toJson in Utils of livy-core. Although livy-repl only use usingResource for now, it is possible to use toJson in future because livy-repl is also web server based.

zjffdu:
ping @vanzin

repl/src/main/scala/com/cloudera/livy/repl/Main.scala
```diff
@@ -135,17 +133,19 @@ class ScalatraBootstrap extends LifeCycle with Logging {
...
-
-        var req = url(callbackUrl).setContentType("application/json", "UTF-8")
-        req = req << write(Map("url" -> replUrl))
-
         info(s"Calling $callbackUrl...")
-        val rep = Http(req OK as.String)
-        rep.onFailure {
-          case _ => System.exit(1)
+        val response = Utils.usingResource(new AsyncHttpClient()) {
+          request => request.preparePost(callbackUrl)
```
vanzin: nit: put `request =>` in the previous line

    val response = Utils.usingResource(new AsyncHttpClient()) { request =>
      // code
    }

Also, `client` is probably more accurate than `request`.

repl/src/main/scala/com/cloudera/livy/repl/Main.scala
```diff
@@ -135,17 +133,19 @@ class ScalatraBootstrap extends LifeCycle with Logging {
...
+            .setHeader("Content-Type", "application/json;charset=UTF-8")
+            .setBody(write(Map("url" -> replUrl)))
+            .execute().get()
+        }
+        response.getStatusCode match {
```
vanzin: Is it safe to try to read the response's body after the client has been closed? (Not really familiar with the library's semantics.)

repl/src/main/scala/com/cloudera/livy/repl/Main.scala
```diff
@@ -135,17 +133,19 @@ class ScalatraBootstrap extends LifeCycle with Logging {
...
+          case 200 => Future.successful(())
```
vanzin: nit: `HttpServletResponse.SC_OK`.

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
```diff
@@ -266,17 +277,14 @@ class InteractiveSession(
...
-    dispatch.url(url.toString)
-  }
-
   @tailrec
   private def waitForStatement(id: Int): JValue = {
-    val req = (svc / "history" / id).setContentType("application/json", "UTF-8")
-    val resp = Await.result(Http(req OK as.json4s.Json), Duration.Inf)
-
-    parseResponse(resp) match {
+    val response = Utils.usingResource(new AsyncHttpClient()){ request =>
```
vanzin: nit: space before `{`

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
```diff
@@ -18,7 +18,7 @@
 
 package com.cloudera.livy.server.interactive
 
-import java.io.File
+import java.io.{Closeable, File}
```
vanzin: nit: `Closeable` not used?

vanzin:
Mostly style nits left, just one real question.

repl/src/main/scala/com/cloudera/livy/repl/Main.scala
zjffdu: Yes, it is safe to do that. Because Utils.usingResource is a blocking method, it has already fetch the response before it close the connection

vanzin: I don't know. Creating a response object is different from reading the response's body; imagine if this were a huge file being downloaded, would `usingResource` download the whole file before returning, or would it leave it for the code dealing with the response object to decide what to do with the response once it analyses the headers?

But, anyway, not a huge deal now, especially since this code will go away soon.

vanzin:
Merging, thanks

repl/src/main/scala/com/cloudera/livy/repl/Main.scala
zjffdu: Regarding the doc here https://github.com/AsyncHttpClient/async-http-client
All the content has been read to memory for this kind of usage. But, yes, if we use usingResource to return something like InputStream then that would be a problem.

PR #88, opened 2016-03-09
LIVY-89. PySparkInterpreter doesn't work

zjffdu:


repl/src/main/resources/fake_shell.py
```diff
@@ -368,6 +369,11 @@ def main():
     sys.stderr = fake_stderr
 
     try:
+
```
vanzin: nit: remove this empty line

repl/src/main/resources/fake_shell.py
```diff
@@ -368,6 +369,11 @@ def main():
...
+        if os.environ["LIVY_TEST"] == "false":
```
vanzin: This will cause an error if the variable is not set.

    >>> import os
    >>> os.environ['LIVY_TEST']
    Traceback (most recent call last):
      File "<stdin>", line 1, in <module>
      File "/usr/lib/python2.7/UserDict.py", line 23, in __getitem__
        raise KeyError(key)
     KeyError: 'LIVY_TEST'

Instead:

    if os.environ.get('LIVY_TEST') == "true": ...

vanzin:
Oops. We really need to invest some time in adding proper integration tests...

repl/src/main/resources/fake_shell.py
vanzin: Or rather, just store it in a variable:

    is_testing = os.environ.get('LIVY_TEST') == "true"

And use it later when cleaning up also.

repl/src/main/resources/fake_shell.py
```diff
@@ -368,6 +369,10 @@ def main():
     sys.stderr = fake_stderr
 
     try:
+        if os.environ.get("LIVY_TEST") == "false":
```
vanzin: Actually, I take my comment back. This won't work at all, because in production `os.environ.get("LIVY_TEST") == "false"` is `False` so you won't run the code inside the condition.

    >>> import os
    >>> os.environ.get("LIVY_TEST") == "false"
    False

vanzin:
> Do I miss anything here ?

When you run PythonInterpreter outside of the maven build, for example when actually running the Livy server, "LIVY_TEST" will not be set in the environment. So you code won't work. That's the situation I pointed out in my python shell examples above.

(BTW if you just edit your comment I don't get an e-mail and I don't see that there's anything new to look at.)

vanzin:
Ah, wait. You're explicitly setting that env variable. So it makes sense now.

Instead, could you use the existing test variable `livy.test`? That's automatically set during tests so you don't need to make any changes to PythonInterpreter at all.

zjffdu:
Local system test passed

vanzin:
@zjffdu you need to rebase on top of master. I can't merge this patch as is.

zjffdu:
PR rebased

vanzin:
Merging, thanks!

PR #89, opened 2016-03-10
Fix hadoop dependencies scope in client-http.

vanzin:
d8842102dd upgraded hadoop dependencies to be "compile" by default,
instead of inheriting the scope from the transitive dependencies.
So the client-http artifact ballooned up to 70MB.

So override the scope back to "test", bringing the jar back down to
a more manageable 3MB (which is also much faster to build).

### DIFF
--- a/client-http/pom.xml
+++ b/client-http/pom.xml
@@ -31,6 +31,7 @@
 
   <properties>
     <skipDeploy>false</skipDeploy>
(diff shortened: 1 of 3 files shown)


tc0312:
We are going to fork. Please revert d884210.

vanzin:
Great, I really think that's the best path going forward if you need to customize the packaging. (If you look at the updated diff, other places were also broken.)

PR #90, opened 2016-03-10
LIVY-58. Allow connections between Livy and RSC to be re-established.

vanzin:
This change modifies the direction of connections between Livy and the RSC,
so that it's Livy connecting to the RSC instead of the other way around.
This makes the RSC behave sort of like the current REPL code, where the
context will connect back to Livy to tell its listen address so that Livy
can connect to it.

For that to work, the RemoteDriver now runs an RpcServer, and does not
kill itself when client connections go away. It also allows multiple
concurrent client connections to exist, although it should be rare for
that situation to exist in the real world.

The code to launch Spark was refactored into a separate class, since
when connecting to an existing RSC it needs to be bypassed altogether.

### DIFF
--- a/client-local/src/main/java/com/cloudera/livy/client/local/ContextLauncher.java
+++ b/client-local/src/main/java/com/cloudera/livy/client/local/ContextLauncher.java
@@ -59,7 +59,6 @@ class ContextLauncher implements ContextInfo {
 
   private static final Logger LOG = LoggerFactory.getLogger(ContextLauncher.class);
   private static final AtomicInteger CHILD_IDS = new AtomicInteger();
-  private static final long DEFAULT_SHUTDOWN_TIMEOUT = 10000; // In milliseconds
 
   private static final String SPARK_JARS_KEY = "spark.jars";
   private static final String SPARK_HOME_ENV = "SPARK_HOME";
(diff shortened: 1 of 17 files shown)


client-local/src/main/java/com/cloudera/livy/client/local/ContextLauncher.java
```diff
@@ -0,0 +1,446 @@
...
+            "Child has exited before address information was received.");
+          now = System.nanoTime();
+        }
+      }
+      Preconditions.checkState(handler.driverAddress != null,
+        "Timed out waiting for driver connection information.");
+      driverAddress = handler.driverAddress;
+    } catch (Exception e) {
+      factory.getServer().unregisterClient(clientId);
+      throw Throwables.propagate(e);
```
harishreedharan: This can just be `Throwables.propogate(e)`. The `throw` is required only if the calling method expects a return value.

client-local/src/main/java/com/cloudera/livy/client/local/ContextLauncher.java
```diff
@@ -0,0 +1,446 @@
...
+    return secret;
+  }
+
+  @Override
+  public void dispose(boolean forceKill) {
+    try {
+      if (forceKill) {
+        child.kill();
+      }
+      child.detach();
```
harishreedharan: Looks like `kill` will detach anyway. Is this call safe to make on an already killed child? I have not yet reached this method in the review so if it is fine to call it again, then ignore this comment.

client-local/src/main/java/com/cloudera/livy/client/local/ContextLauncher.java
vanzin: I prefer using that idiom always. It avoids other issues such as the compiler complaining about uninitialized local variables, and it doesn't cost anything.

client-local/src/main/java/com/cloudera/livy/client/local/ContextLauncher.java
vanzin: Calling `detach()` again should be safe, but I can add an "else" here.

harishreedharan:
Now both livy and the RSC have servers running, but the one on the livy side is only used by the drivers to send their own addresses back - is my understanding correct?

vanzin:
That is correct. The Livy-side server is only started on-demand, and we could potentially shut it down sooner also.

PR #91, opened 2016-03-11
Specify java version in maven-compiler-plugin

zjffdu:
The purpose is that it can help intellij recognize the java version and set the language level correctly in intellij. Otherwise intellij always use java 5 which is very annoying.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -696,6 +696,10 @@
             <phase>none</phase>
           </execution>
         </executions>
+        <configuration>
+          <source>${java.version}</source>
+          <target>${java.version}</target>
+        </configuration>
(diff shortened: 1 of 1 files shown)


vanzin:
To avoid people compiling for 1.8 inadvertently, could you declare a `java.version` in the properties section with default value 1.7? (The scala plugin already references `java.version`, so this would force it to be 1.7 by default, which should have been the case from the start.)

Also it's funny that intellij complains when the plugin is actually disabled in the build.

vanzin:
Actually, nevermind about the property, it already exists (d'oh).

Merging, thanks!

PR #92, opened 2016-03-12
[LIVY-46] Use RSC in REPL

harishreedharan:
This patch changes the REPL to use the RSC to communicate with Livy instead of the current HTTP based method. This includes a number of architectural changes in the driver side:
- [x] A new driver that runs the REPL has been introduced which communicates with Livy via the RSC protocol.
- [ ] The `InteractiveSession` sends the code from Livy to the new driver via the RSC
- [x] Integrate LIVY-58 changes.


harishreedharan:
This is an initial preview PR. More changes are required to get this completed, most importantly merging changes that went in via LIVY-58.

harishreedharan:
@tc0312 This might be worth taking a look at (of course, this does not yet do LIVY-58 style connections).

api/src/main/java/com/cloudera/livy/JobContext.java
```diff
@@ -59,6 +59,8 @@
   /** Stops the SparkStreaming context. */
   void stopStreamingCtx();
 
+  void stop();
```
vanzin: I haven't looked at the rest, but we shouldn't expose this. Clients should use `LivyClient.stop`.

vanzin:
So, I see you're trying to run the REPL as its own driver; that's the source of most of the refactoring you're having to do here.

When I started thinking about this problem (and I didn't write any code, so take this with a grain of salt), my idea was to run the REPL as an RSC application. Mainly, you'd do something like this:

```
val rsc = new LivyClientBuilder(blah).build()

rsc.run(new StartReplJob())
rsc.submit(new ReplRequest(code))

class StartReplJob extends Job<Unit> {

  override def run(jc: JobContext): Unit = {
    SomeStaticPlace.REPL = new Repl();
  }

}

class ReplRequest(code: String) extends Job<...> {

  override def run(jc: JobContext): ... = {
     SomeStaticPlace.REPL.run(code)
  }

}

```

Do you think that would work? It seems to me that would be possible and potentially could avoid changing the RSC at all as part of this change, making it a lot simpler.

harishreedharan:
That might work. The reason I didn't go that route is because keeping the current driver for the REPL job means that there is a spark context running inside that driver. Creating a REPL automatically creates a new Spark context in the REPL itself (basically how to ensure the resources for the cluster are really used by the REPL's context itself, not the one created by the RSC driver). 

Maybe there is something that I am missing that can allow us to avoid this.

vanzin:
> Creating a REPL automatically creates a new Spark context in the REPL itself 

The REPL would have to be changed to use the context provided by the RSC, instead of instantiating its own. Is that what you're referring to?

harishreedharan:
Right. So if you look at [SparkInterpreter.scala](https://github.com/cloudera/livy/blob/master/repl/src/main/scala/com/cloudera/livy/repl/scalaRepl/SparkInterpreter.scala) it uses the `SparkIMain` class from Spark. Changing this would be painful as it needs to go upstream (even then I don't know if that would be trivial). Do you know of an alternative?

vanzin:
I see. `SparkIMain.createSparkContext()` is public so it could be overridden, although it might be tricky to get the existing Spark context to serve generated classes (see big comment before `spark.repl.class.outputDir` is set in that method).

But you still could use some of the same ideas, for example submitting code using `LivyClient.submit()` instead of modifying the internal RSC protocol to handle a new message type, right?

harishreedharan:
Handing this over to @vanzin

vanzin:
This is now replaced with #94.

PR #93, opened 2016-03-17
LIVY-85. support pyspark with python3

purechoc:
support python3 using pyspark3

added kind type 'pyspark3'
`kind = [pyspark, pyspark3, spark, sparkr]`

Hue Notebook using like this
```
  [[interpreters]]
    [[[pyspark]]]
      name=PySpark
      interface=livy
    [[[pyspark3]]]
      name=PySpark3
      interface=livy
```

if user want to using python3, all cluster node need to be installed python3.


purechoc:
modified fake_shell.py code can run both python2 and python3.

repl/src/main/resources/fake_shell.py
```diff
@@ -261,26 +268,32 @@ def magic_table_convert_map(m):
...
     float: lambda x: ('DOUBLE_TYPE', x),
     str: lambda x: ('STRING_TYPE', str(x)),
-    unicode: lambda x: ('STRING_TYPE', x.encode('utf-8')),
     datetime.date: lambda x: ('DATE_TYPE', str(x)),
     datetime.datetime: lambda x: ('TIMESTAMP_TYPE', str(x)),
     decimal.Decimal: lambda x: ('DECIMAL_TYPE', str(x)),
     tuple: magic_table_convert_seq,
     list: magic_table_convert_seq,
-    dict: magic_table_convert_map,
+    dict: magic_table_convert_map,
```
frol: This seems to be a meaningless change. Would you mind to strip ending spaces (there are some in other places in your PR)?

purechoc:
strip ending spaces

tc0312:
@purechoc Would you mind updating your PR? If you are too busy, let' us know. We can help.

purechoc:
@tc0312 sure, I will update this PR soon. (may be next week)

purechoc:
@tc0312  update completed.

tc0312:
LGTM

repl/src/main/resources/fake_shell.py
```diff
@@ -14,18 +14,24 @@
...
 import json
 import logging
 import sys
 import traceback
+import base64
 import os
 import re
-import StringIO
+
+if sys.version >= '3' :
```
vanzin: nit: no space before `:`

repl/src/main/resources/fake_shell.py
```diff
@@ -90,13 +96,13 @@ def execute(self):
...
-                exec code in global_dict
+                exec(code, global_dict)
 
             for node in to_run_single:
                 mod = ast.Interactive([node])
                 code = compile(mod, '<stdin>', 'single')
-                exec code in global_dict
-        except:
+                exec(code, global_dict)
+        except :
```
vanzin: nit: no space before `:`

repl/src/main/scala/com/cloudera/livy/repl/PythonInterpreter.scala
```diff
@@ -37,8 +37,11 @@ import com.cloudera.livy.client.common.ClientConf
 
 // scalastyle:off println
 object PythonInterpreter extends Logging {
-  def apply(conf: SparkConf): Interpreter = {
-    val pythonExec = sys.env.getOrElse("PYSPARK_DRIVER_PYTHON", "python")
+  def apply(conf: SparkConf, kind: String = "pyspark"): Interpreter = {
```
vanzin: Why use a string here? Why not just use the `Kind` object?

repl/src/main/scala/com/cloudera/livy/repl/PythonInterpreter.scala
```diff
@@ -37,8 +37,11 @@ import com.cloudera.livy.client.common.ClientConf
...
+    val pythonExec = {
+        if (kind == "pyspark3") sys.env.getOrElse("PYSPARK3_DRIVER_PYTHON", "python3")
```
vanzin: So not only you're using a string but you're using it as a boolean... it'd be better to just use a boolean from the get go, or do proper matching (and throw an error if the wrong value is provided).

server/src/test/scala/com/cloudera/livy/server/interactive/CreateInteractiveRequestSpec.scala
```diff
@@ -21,7 +21,7 @@ package com.cloudera.livy.server.interactive
 import com.fasterxml.jackson.databind.ObjectMapper
 import org.scalatest.FunSpec
 
-import com.cloudera.livy.sessions.{PySpark, SessionKindModule}
+import com.cloudera.livy.sessions.{PySpark, PySpark3, SessionKindModule}
```
vanzin: The changes to this file are unnecessary.

vanzin:
My main concern here is the lack of tests. There's no code that's checking whether the python 3 support actually works to any extent.

I'm not sure whether travis-ci VMs have python3 by default; if they don't you could add tests that "auto-disable" themselves in case python3 is not found (see `rtest` in `BaseIntegrationTestSuite.scala` for how to do something like that).

repl/src/main/scala/com/cloudera/livy/repl/PythonInterpreter.scala
purechoc: because using choose pythonExec  condition and "kind" string
` new PythonInterpreter(process, gatewayServer, kind)`
`private class PythonInterpreter(process: Process, gatewayServer: GatewayServer, pyKind: String)`
`override def kind: String = pyKind`

vanzin: You can use `Kind` and return `kind.toString` in that method. That's better than having magic strings like this.

repl/src/main/resources/fake_shell.py
purechoc: fixed

repl/src/main/resources/fake_shell.py
purechoc: fixed

repl/src/main/resources/fake_shell.py
```diff
@@ -186,8 +193,8 @@ def execute_request(content):
             result = node.execute()
     except UnknownMagic:
         exc_type, exc_value, tb = sys.exc_info()
-        return execute_reply_error(exc_type, exc_value, [])
-    except ExecutionError, e:
+        return execute_reply_error(exc_type, exc_value, None)
+    except (ExecutionError) as e:
```
purechoc: fixed

repl/src/main/scala/com/cloudera/livy/repl/PythonInterpreter.scala
purechoc: it's fixed.
using kind object instead of string

purechoc:
@vanzin 
thank to advise
need I add test case for pyspark3?
something like add Python3InterpreterSpec, Python3SessionSpec?

vanzin:
> need I add test case for pyspark3?

Yes that would be good. They could be the existing tests, executed on python3 (aside from python2), if you refactor things to make that possible.

purechoc:
pyspark test case add completed
and add 'skipPySpark3Tests' property like 'skipRTests'

repl/src/test/scala/com/cloudera/livy/repl/Python3InterpreterSpec.scala
```diff
@@ -208,15 +208,5 @@ class Python3InterpreterSpec extends BaseInterpreterSpec {
     ))
   }
 
-  // Scalastyle is treating unicode escape as non ascii characters. Turn off the check.
-  // scalastyle:off non.ascii.character.disallowed
-  it should "print unicode correctly" in withInterpreter { intp =>
-    intp.execute("print(u\"\u263A\")") should equal(Interpreter.ExecuteSuccess(
```
tc0312: Mind if I ask why is this test case removed?

purechoc: because, that case not working python3.
```
    intp.execute("""print("\xE2\x98\xBA")""") should equal(Interpreter.ExecuteSuccess(
      TEXT_PLAIN -> "\u263A"
    ))
```
return \xE2\x98\xBA not \u263A
i try fix this. but need more time.
could you help if you do not mind

vanzin: Instead of removing the test, the code should be fixed, no?

purechoc: that time, i'm not sure that case need to be python3.
but. you right. i will fixed code next week.

purechoc: @tc0312 @vanzin 
Does above testcase need to Python3? 
Does ` print("\xE2\x98\xBA")` result need to be ""\u263A""?
a  = "'\\xE2\\x98\\xBA'"
python2 
```
print(isinstance(a, bytes))
True
```

python3
```
print(isinstance(a, bytes))
False
```

vanzin: It needs to be there for python 2, but I'm not familiar enough with python 3 to say whether the same problem exists there. @tc0312 ?

repl/src/main/resources/fake_shell.py
```diff
@@ -395,27 +412,29 @@ def clearOutputs():
...
     sys.stderr = UnicodeDecodingStringIO()
 
-
 def main():
     sys_stdin = sys.stdin
     sys_stdout = sys.stdout
     sys_stderr = sys.stderr
 
-    sys.stdin = cStringIO.StringIO()
+    if sys.version >= '3' :
```
vanzin: nit: no space before `:`

repl/src/main/resources/fake_shell.py
```diff
@@ -395,27 +412,29 @@ def clearOutputs():
...
+        sys.stdin = io.StringIO()
+    else :
```
vanzin: nit: no space before `:`

repl/src/main/resources/fake_shell.py
```diff
@@ -395,27 +412,29 @@ def clearOutputs():
...
 
     try:
         if os.environ.get("LIVY_TEST") != "true":
             # Load spark into the context
-            exec 'from pyspark.shell import sc' in global_dict
-
-        print >> sys_stderr, sys.stdout.getvalue()
-        print >> sys_stderr, sys.stderr.getvalue()
+            exec('from pyspark.shell import sc', global_dict)
+        print(sys.stdout.getvalue(), file=sys_stderr)
```
vanzin: nit: could you leave the empty line between the blocks?

repl/src/main/resources/fake_shell.py
```diff
@@ -351,15 +367,17 @@ def magic_json(name):
...
         value = global_dict[name]
-
         fig = value.gcf()
-        imgdata = StringIO.StringIO()
+        imgdata = io.BytesIO()
         fig.savefig(imgdata, format='png')
         imgdata.seek(0)
-        encode = base64.b64encode(imgdata.buf)
+        encode = base64.b64encode(imgdata.getvalue())
+        if sys.version >= '3' :
```
vanzin: nit: no space before `:`

repl/src/main/scala/com/cloudera/livy/repl/PythonInterpreter.scala
```diff
@@ -34,11 +34,15 @@ import py4j.GatewayServer
 
 import com.cloudera.livy.Logging
 import com.cloudera.livy.client.common.ClientConf
+import com.cloudera.livy.sessions._
 
 // scalastyle:off println
 object PythonInterpreter extends Logging {
-  def apply(conf: SparkConf): Interpreter = {
-    val pythonExec = sys.env.getOrElse("PYSPARK_DRIVER_PYTHON", "python")
+  def apply(conf: SparkConf, kind: Kind = PySpark()): Interpreter = {
```
vanzin: Instead of a default value, I'd rather make this explicit when creating a `PythonInterpreter`.

repl/src/test/scala/com/cloudera/livy/repl/Python3InterpreterSpec.scala
```diff
@@ -0,0 +1,212 @@
...
+package com.cloudera.livy.repl
+
+import org.apache.spark.SparkConf
+import org.json4s.{DefaultFormats, JValue}
+import org.json4s.JsonDSL._
+import org.scalatest._
+
+import com.cloudera.livy.sessions._
+
+class Python3InterpreterSpec extends BaseInterpreterSpec {
```
vanzin: This is a good start, but it doesn't share any code with the existing tests. It would be better to have an abstract base class for python interpreter tests, and different real implementations for python 2 and python 3; then adding tests for both is much easier - just change one class, and it's also easy to add version-specific tests (which should be very rare).

The 3 classes could all live in the same source file, since I don't expect too much code in the version-specific classes.

repl/src/test/scala/com/cloudera/livy/repl/Python3SessionSpec.scala
```diff
@@ -0,0 +1,179 @@
...
+
+package com.cloudera.livy.repl
+
+import org.apache.spark.SparkConf
+import org.json4s.Extraction
+import org.scalatest._
+
+import com.cloudera.livy.sessions._
+
+class Python3SessionSpec extends BaseSessionSpec {
```
vanzin: Same comments about sharing test code applies here.

repl/src/main/resources/fake_shell.py
purechoc: fixed it

repl/src/main/resources/fake_shell.py
purechoc: fixed it

repl/src/main/resources/fake_shell.py
purechoc: fixed it

repl/src/main/resources/fake_shell.py
purechoc: fixed it

repl/src/main/scala/com/cloudera/livy/repl/PythonInterpreter.scala
purechoc: fixed it

repl/src/test/scala/com/cloudera/livy/repl/Python3InterpreterSpec.scala
purechoc: thank to comment.
i'll try update this code.

purechoc:
make abstract base class for python interpreter tests, 
python 2 and python 3 implements that. 
now python 2 and 3 shared common test case.

repl/src/test/scala/com/cloudera/livy/repl/PythonSessionSpec.scala
```diff
@@ -167,5 +168,34 @@ class PythonSessionSpec extends BaseSessionSpec {
...
+class Python3SessionSpec extends PythonSessionSpec {
+
+  override protected def withFixture(test: NoArgTest): Outcome = {
+    assume(!sys.props.getOrElse("skipPySpark3Tests", "false").toBoolean, "Skipping PySpark3 tests.")
+    test()
+  }
+
+  override def createInterpreter(): Interpreter = PythonInterpreter(new SparkConf(), PySpark3())
 
+  it should "execute `1 / 2` == 0.5" in withSession { session =>
```
vanzin: Is this just to check it's really running python 3? Wouldn't it be easier to just check `sys.version`?

repl/src/test/scala/com/cloudera/livy/repl/PythonInterpreterSpec.scala
```diff
@@ -215,3 +221,22 @@ class PythonInterpreterSpec extends BaseInterpreterSpec {
...
+  implicit val formats = DefaultFormats
+
+  override protected def withFixture(test: NoArgTest): Outcome = {
+    assume(!sys.props.getOrElse("skipPySpark3Tests", "false").toBoolean, "Skipping PySpark3 tests.")
+    test()
+  }
+
+  override def createInterpreter(): Interpreter = PythonInterpreter(new SparkConf(), PySpark3())
+
+  it should "execute `1 / 2` == 0.5" in withInterpreter { interpreter =>
```
vanzin: Same question re: what's this really testing.

vanzin:
Looks ok aside from the test questions. Could you also rebase this now that there's a conflict?

repl/src/test/scala/com/cloudera/livy/repl/Python3InterpreterSpec.scala
tc0312: Sorry for the late reply.
Feel free to remove case "\xE2\x98\xBA". But please keep other test cases.

Also, I think we can use io.StringIO() directly for stdout and stderr in Python3:
Mind if u try this and see does it work?

```python
if sys.version >= '3':
  sys.stdin = io.StringIO()
  sys.stdout = io.StringIO()
  sys.stderr = io.StringIO()
else:
  sys.stdin = cStringIO.StringIO()
  sys.stdout = UnicodeDecodingStringIO()
  sys.stderr = UnicodeDecodingStringIO()
```

repl/src/test/scala/com/cloudera/livy/repl/PythonInterpreterSpec.scala
purechoc: tested python3 result.
execute `1 / 2` == 0.5
python 2 result : 0
python 3 result : 0.5

vanzin: I know that. But it's not interesting to test that in python 3, "1/2 = 0.5". It is interesting to test that you're really running python 3, and it's much clearer if instead you explicitly check that, instead of testing for some weird language feature.

purechoc: ok. i understand. i fixed this test case simply to check sys.version

purechoc:
@vanzin 
update completed.

tc0312:
Sorry for the late reply.
Feel free to remove case "\xE2\x98\xBA". But please keep other test cases.

Also, I think we can use io.StringIO() directly for stdout and stderr in Python3:
Mind if u try this and see does it work?

```python
if sys.version >= '3':
  sys.stdin = io.StringIO()
  sys.stdout = io.StringIO()
  sys.stderr = io.StringIO()
else:
  sys.stdin = cStringIO.StringIO()
  sys.stdout = UnicodeDecodingStringIO()
  sys.stderr = UnicodeDecodingStringIO()
```

purechoc:
@tc0312 
i was test about that, but it's not work.
(not error, just different output)
python 3 does cognized "\xE2\x98\xBA" is string. not bytes.
``` 
if sys.version >= '3':
  sys.stdin = io.StringIO()
  sys.stdout = io.StringIO()
  sys.stderr = io.StringIO()
else:
  sys.stdin = cStringIO.StringIO()
  sys.stdout = UnicodeDecodingStringIO()
  sys.stderr = UnicodeDecodingStringIO()
```

vanzin:
Merging.

PR #94, opened 2016-03-22
LIVY-46. Use the RSC to drive REPL sessions.

vanzin:
This change modifies interactive sessions so they use the RSC (livy-client-local) to
run the REPL driver. This removes the embedded Jetty server in the REPL process,
so it solves other issues too; namely, LIVY-95, LIVY-84, and maybe others that I have
forgotten about.

The change works by having the RSC instantiate a REPL-specific driver when a certain
config option is set. This is currently a little hacky and not very extensible, but we can work
on those parts later if there's a need. The `LocalClient` class exposes some REPL-specific
methods so that its main client (the "server" module) doesn't need REPL jars in its classpath.
These methods just send RPCs to the REPL driver and return their results.

One notable loss of functionality is that the interactive session servlet does not expose REPL
logs anymore.  But because those logs were the output of "spark-submit", they're not very useful.
In yarn-cluster mode, there's only app status information there, nothing related about what's
really happening in Spark. LIVY-55 is the proper way of exposing logs. Another loss is the removal of the `bin/livy-repl` script; if that functionality is desired, a new module that brings back the HTTP frontend to the repl will have to be created, to avoid the issues in LIVY-84.


vanzin:
Note there's more work to do here, especially cleanups. I want to leave this open so others can take a look and also to run tests on travis, since they're now passing locally.

vanzin:
I'm gonna merge this since it fixes a whole lot of current issues. There's still cleanup / code reorg that can be done, but let's do that separately.

PR #96, opened 2016-03-30
LIVY-12. R + Yarn support

purechoc:
using SparkR in yarn cluster mode

**Changes**
- remove fake_R and SPARKR_DRIVER_R

**How to work**
- ContextLauncher launch SparkRInterpreter  with sparkRArchive file ($SPARK_HOME/R/lib/sparkr.zip)
- SparkRInterpreter Launch a SparkR backend server for the R process
- exec R (all cluster node must installed R and R in $PATH)
- load SparkR library


client-local/src/main/java/com/cloudera/livy/client/local/ContextLauncher.java
```diff
@@ -247,6 +247,10 @@ public void run() {
       }
       launcher.addAppArgs("--remote-host", serverAddress);
       launcher.addAppArgs("--remote-port",  serverPort);
+      if ("sparkr".equals(conf.get("session.kind"))) {
+        launcher.addSparkArg("--archives", conf.get("sparkr.package"));
```
vanzin: I'd rather see a `LocalConf.Entry` being created for this setting instead of hardcoding strings.

But more importantly, you shouldn't be using `--archives`; instead, you should be merging this with the `spark.yarn.dist.archives` config option which the user might be setting. This is done above for `SPARK_JARS_KEY` - you could refactor the code so that you can reuse it here.

conf/livy-defaults.conf.template
```diff
@@ -38,3 +38,6 @@
 # or an hdfs:// directory. Be careful setting this option, as it may allow users to access
 # any file the Livy process can access, such as Livy's SSL certificate.
 ## livy.files.dir = /var/run/livy/uploaded-files
+
+# Location to sparkr package file(sparkr.zip) with symbolic link
```
vanzin: Doesn't the archive always need to be extracted to `sparkr` on the YARN side? You should't tell the user to do that, much else allow the user to override that. The code in Livy reading this configuration should make sure the fragment (i.e. anything after the '#') is ignored.

repl/src/main/scala/com/cloudera/livy/repl/sparkr/SparkRInterpreter.scala
```diff
@@ -62,19 +67,37 @@ object SparkRInterpreter {
...
+        sparkRBackendClass.getMethod("run").invoke(ctor)
+      }
+    }
 
-    val builder = new ProcessBuilder(Seq(executable.getAbsolutePath).asJava)
+    sparkRBackendThread.start()
+    // Wait for RBackend initialization to finish
+    initialized.tryAcquire(backendTimeout, TimeUnit.SECONDS)
 
+    val builder = new ProcessBuilder(Seq("R", "--slave @").asJava)
```
vanzin: This assumes `R` is in `PATH`. Shouldn't you also allow `SPARKR_DRIVER_R` to be set somehow?

repl/src/main/scala/com/cloudera/livy/repl/sparkr/SparkRInterpreter.scala
```diff
@@ -62,19 +67,37 @@ object SparkRInterpreter {
...
+    val backendTimeout = sys.env.getOrElse("SPARKR_BACKEND_TIMEOUT", "120").toInt
+    val mirror = universe.runtimeMirror(getClass.getClassLoader)
+    val sparkRBackendClass = mirror.classLoader.loadClass("org.apache.spark.api.r.RBackend")
+    val ctor = sparkRBackendClass.getDeclaredConstructor().newInstance()
+
+    var sparkRBackendPort = 0
+    val initialized = new Semaphore(0)
+    // Launch a SparkR backend server for the R process to connect to
+    val sparkRBackendThread = new Thread("SparkR backend") {
+      override def run() {
```
vanzin: `run(): Unit = {`

repl/src/main/scala/com/cloudera/livy/repl/sparkr/SparkRInterpreter.scala
```diff
@@ -62,19 +67,37 @@ object SparkRInterpreter {
...
+        sparkRBackendPort = sparkRBackendClass.getMethod("init").invoke(ctor).asInstanceOf[Int]
+        initialized.release()
+        sparkRBackendClass.getMethod("run").invoke(ctor)
+      }
+    }
 
-    val builder = new ProcessBuilder(Seq(executable.getAbsolutePath).asJava)
+    sparkRBackendThread.start()
```
vanzin: Should this be a daemon thread?

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
```diff
@@ -79,7 +79,10 @@ class InteractiveSession(
         val pySparkFiles = if (!LivyConf.TEST_MODE) findPySparkArchives() else Nil
         builder.setConf(SparkYarnIsPython, "true")
         builder.setConf(SparkSubmitPyFiles, (pySparkFiles ++ request.pyFiles).mkString(","))
-
+      case SparkR() =>
+        val packageFile = livyConf.get("livy.repl.sparkr.package")
+        require(packageFile != null, "Can not find SparkR pakcage path.")
```
vanzin: "package". Also, why not look it up under "SPARK_HOME" if it's not explicitly set?

Generally, these configs exist so that the user can cache these files in HDFS, for example, making distribution quicker. But you should be able to find them without the user having to explicitly tell them where they are.

repl/src/main/scala/com/cloudera/livy/repl/sparkr/SparkRInterpreter.scala
```diff
@@ -62,19 +67,37 @@ object SparkRInterpreter {
...
 
+    val builder = new ProcessBuilder(Seq("R", "--slave @").asJava)
     val env = builder.environment()
+
+    env.put("EXISTING_SPARKR_BACKEND_PORT", sparkRBackendPort.toString)
+    env.put("SPARKR_PACKAGE_DIR", "./sparkr")
+    env.put("R_PROFILE_USER",
+        Seq("./sparkr", "SparkR", "profile", "general.R").mkString(File.separator))
     env.put("SPARK_HOME", sys.env.getOrElse("SPARK_HOME", "."))
-    env.put("SPARKR_DRIVER_R", createFakeShell().toString)
```
vanzin: Can't you just delete the `createFakeShell` method now, along with `fake_R.sh` and `dev/spark/bin/sparkR`?

repl/src/main/scala/com/cloudera/livy/repl/sparkr/SparkRInterpreter.scala
```diff
@@ -120,13 +143,23 @@ class SparkRInterpreter(process: Process)
 
   override def kind: String = "sparkR"
 
+  private[this] var isStart = false
+
   final override protected def waitUntilReady(): Unit = {
+    if (! LivyConf.TEST_MODE) {
```
vanzin: nit: no space after `!`.

repl/src/main/scala/com/cloudera/livy/repl/sparkr/SparkRInterpreter.scala
```diff
@@ -120,13 +143,23 @@ class SparkRInterpreter(process: Process)
 
   override def kind: String = "sparkR"
 
+  private[this] var isStart = false
```
vanzin: Use a `CountDownLatch`.

vanzin:
Also could you write a more descriptive change summary? As in, what was changed to allow SparkR to run in cluster mode?

purechoc:
thank feedback
i fixed some code and push again.
 - remove "--archives" and using "spark.yarn.dist.archives" and SPARK_ARCHIVES_KEY
 - remove SPARKR_DRIVER_R, fake_R.sh, dev/spark/bin/sparkR (beacause SPARKR_DRIVER_R does not need anymore, if I'm wrong, tell me again, i will fixed that)
 - add sparkRBackendThread.setDaemon(true)
 - find sparkr archive file using $SPARK_HOME

vanzin:
Hi @purechoc , I don't see any of my comments addressed in your last push. Also, the summary is not really readable - looks like you're not using correct markdown.

purechoc:
Hi. @vanzin. 
it's my mistake. i will fix this soon.

purechoc:
i was missing command 'add' before 'commit'

repl/src/main/scala/com/cloudera/livy/repl/sparkr/SparkRInterpreter.scala
```diff
@@ -62,51 +67,40 @@ object SparkRInterpreter {
     ).r.unanchored
 
   def apply(): SparkRInterpreter = {
-    val executable = sparkRExecutable
-      .getOrElse(throw new IllegalStateException("Cannot find sparkR executable."))
+    val backendTimeout = sys.env.getOrElse("SPARKR_BACKEND_TIMEOUT", "120").toInt
+    val mirror = universe.runtimeMirror(getClass.getClassLoader)
+    val sparkRBackendClass = mirror.classLoader.loadClass("org.apache.spark.api.r.RBackend")
+    val ctor = sparkRBackendClass.getDeclaredConstructor().newInstance()
```
vanzin: `ctor` is a misleading name. This is an instance of the class, not a reference to the class's constructor.

repl/src/main/scala/com/cloudera/livy/repl/sparkr/SparkRInterpreter.scala
```diff
@@ -62,51 +67,40 @@ object SparkRInterpreter {
...
+        initialized.release()
+        sparkRBackendClass.getMethod("run").invoke(ctor)
+      }
+    }
 
-    val builder = new ProcessBuilder(Seq(executable.getAbsolutePath).asJava)
+    sparkRBackendThread.setDaemon(true)
+    sparkRBackendThread.start()
+    // Wait for RBackend initialization to finish
+    initialized.tryAcquire(backendTimeout, TimeUnit.SECONDS)
```
vanzin: If an exception is thrown here, do you need to shut down (i.e. interrupt) the backend thread?

repl/src/main/scala/com/cloudera/livy/repl/sparkr/SparkRInterpreter.scala
vanzin: You haven't addressed this question. What if the user has `R` installed on a custom location that is not in the path? You need some way to support that.

repl/src/main/scala/com/cloudera/livy/repl/sparkr/SparkRInterpreter.scala
```diff
@@ -117,16 +111,22 @@ class SparkRInterpreter(process: Process)
   implicit val formats = DefaultFormats
 
   private[this] var executionCount = 0
-
   override def kind: String = "sparkR"
+  private[this] val isStart = new CountDownLatch(1);
 
   final override protected def waitUntilReady(): Unit = {
+    if (!LivyConf.TEST_MODE) {
+        sendRequest("library(SparkR)")
```
vanzin: nit: indented too far

repl/src/main/scala/com/cloudera/livy/repl/sparkr/SparkRInterpreter.scala
```diff
@@ -117,16 +111,22 @@ class SparkRInterpreter(process: Process)
   implicit val formats = DefaultFormats
 
   private[this] var executionCount = 0
-
   override def kind: String = "sparkR"
+  private[this] val isStart = new CountDownLatch(1);
```
vanzin: nit: `isStarted`

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
```diff
@@ -202,6 +204,21 @@ class InteractiveSession(
     }
   }
 
+  private def findSparkRArchives(): Seq[String] = {
+    sys.env.get("SPARKR_ARCHIVES_PATH")
+      .map(_.split(",").toSeq)
+      .getOrElse {
+        sys.env.get("SPARK_HOME").map { case sparkHome =>
+          val rLibPath = Seq(sparkHome, "R", "lib").mkString(File.separator)
+          val rArchivesFile = new File(rLibPath, "sparkr.zip")
```
vanzin: So, there's a minor problem here, in that sparkr.zip does not exist before Spark 1.6. Spark 1.5 has those files exploded into a directory, which won't work here.

I think that's fine, but we need to document it for users who might really want to run Spark 1.5.

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
```diff
@@ -202,6 +204,21 @@ class InteractiveSession(
...
+          require(rArchivesFile.exists(),
+            "sparkr.zip not found; cannot run sparkr application in YARN mode.")
```
vanzin: The message is misleading; this method is called regardless of how you're running Spark, so if the package is not found, you'll fail with this message even if the user has configured Spark to run in standalone mode.

vanzin:
Hi @purechoc, you need to rebase on top of master now.

Does this still work when not in YARN mode? (I guess the answer is "yes" since unit tests are passing?) Can you try to run a sparkr session with `spark.master=local-cluster[1,1,1024]` and seeing whether everything works as expected?

purechoc:
Hi @vanzin . thanks to kindly advises.

this is working both YARN and local mode. and everything works as expected.

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
purechoc: yes. i agree. 
It needs to be added in a future README file (Prerequisites section)

repl/src/main/scala/com/cloudera/livy/repl/sparkr/SparkRInterpreter.scala
purechoc: now, user can using like this when R is installed custom location
export DRIVER_R=<custom location R>

repl/src/main/scala/com/cloudera/livy/repl/sparkr/SparkRInterpreter.scala
purechoc: oops. it is fixed

repl/src/main/scala/com/cloudera/livy/repl/sparkr/SparkRInterpreter.scala
purechoc: added beckend thread kill code, when exception is thrown

repl/src/main/scala/com/cloudera/livy/repl/sparkr/SparkRInterpreter.scala
purechoc: oops. it is fixed

client-local/src/main/java/com/cloudera/livy/client/local/LocalConf.java
```diff
@@ -48,6 +48,7 @@
     CLIENT_REPL_MODE("repl", false),
 
     LIVY_JARS("jars", null),
+    SPARKR_PACKAGE("spark.package", null),
```
vanzin: oops, missed this before, but should be `sparkr.package`. I'll fix it during merge.

vanzin:
Merging to master, thanks.

PR #97, opened 2016-03-31
LIVY-99. Support _HOST pattern for principal

zjffdu:


server/src/main/scala/com/cloudera/livy/server/Main.scala
```diff
@@ -83,7 +85,8 @@ object Main extends Logging {
 
     livyConf.get(AUTH_TYPE) match {
       case authType @ KerberosAuthenticationHandler.TYPE =>
-        val principal = livyConf.get(KERBEROS_PRINCIPAL)
+        val principal = SecurityUtil.getServerPrincipal(livyConf.get(KERBEROS_PRINCIPAL),
+          InetAddress.getLocalHost.getCanonicalHostName)
```
vanzin: This doesn't really work. This is what I get on my machine:

```
scala> java.net.InetAddress.getLocalHost.getCanonicalHostName
res0: String = localhost
```

You need to use the `SERVER_HOST` configuration here, and have a fallback for when it's set to the wildcard address. In fact, because of how SPNEGO works, `SERVER_HOST` probably cannot be the wildcard address when security is enabled, since clients have to request a ticket that matches that address.

So, basically, I think this should be how this works:
- first check whether the principal contains `SecurityUtils.HOSTNAME_PATTERN`; if it doesn't, don't event try the options below, just use it.
- check `SERVER_HOST`, if it's the wildcard address, print a warning, and don't use it
- try `InetAddress.getLocalHost`; if `isLoopbackAddress()` is true, print a warning. Or maybe even error out.

zjffdu:
SecurityUtil.getServerPrincipal will check the HOSTNAME_PATTERN.
I just use SERVER_HOST, it will be set as canonical hostname if it is wildcard address. Doesn't spnego support loopback address ?

vanzin:
> Doesn't spnego support loopback address ?

It might (it's just a string anyway), but that isn't really useful except for testing.

server/src/main/scala/com/cloudera/livy/server/WebServer.scala
```diff
@@ -96,11 +99,11 @@ class WebServer(livyConf: LivyConf, var host: String, var port: Int) extends Log
     val connector = server.getConnectors()(0).asInstanceOf[NetworkConnector]
 
     if (host == "0.0.0.0") {
-      host = InetAddress.getLocalHost.getHostAddress
+      host = InetAddress.getLocalHost.getCanonicalHostName
     }
     port = connector.getLocalPort
 
-    info("Starting server on %s" format port)
+    info("Starting server on %s" format (protocol + "://" + host + ":" + port))
```
vanzin: Could you use interpolation instead of this hybrid of concatenation and `String.format`?

server/src/main/scala/com/cloudera/livy/server/WebServer.scala
```diff
@@ -42,11 +42,14 @@ class WebServer(livyConf: LivyConf, var host: String, var port: Int) extends Log
   server.setStopTimeout(1000)
   server.setStopAtShutdown(true)
 
+  var protocol = "http"
```
vanzin: `private`?

vanzin: You could also make this a `val` (which is preferable) by changing the code that initializes `connector`:

```
val (connector, protocol) = ... {
  case None => (new ServerConnector(server), "http")
  ...
}
```

vanzin:
LGTM, just style suggestions.

zjffdu:
Looks like the stdout of python process and livy server are mixed together somehow.
```
[31m- should start in the starting or idle state *** FAILED ***[0m
[31m  com.fasterxml.jackson.core.JsonParseException: Unrecognized token 'READY': was expecting ('true', 'false' or 'null')[0m
[31m at [Source: READY; line: 1, column: 11][0m
[31m  at com.fasterxml.jackson.core.JsonParser._constructError(JsonParser.java:1419)[0m
[31m  at com.fasterxml.jackson.core.base.ParserMinimalBase._reportError(ParserMinimalBase.java:508)[0m
[31m  at com.fasterxml.jackson.core.json.ReaderBasedJsonParser._reportInvalidToken(ReaderBasedJsonParser.java:2300)[0m
[31m  at com.fasterxml.jackson.core.json.ReaderBasedJsonParser._handleOddValue(ReaderBasedJsonParser.java:1459)[0m
[31m  at com.fasterxml.jackson.core.json.ReaderBasedJsonParser.nextToken(ReaderBasedJsonParser.java:683)[0m
[31m  at com.fasterxml.jackson.databind.ObjectMapper._initForReading(ObjectMapper.java:3105)[0m
[31m  at com.fasterxml.jackson.databind.ObjectMapper._readMapAndClose(ObjectMapper.java:3051)[0m
[31m  at com.fasterxml.jackson.databind.ObjectMapper.readValue(ObjectMapper.java:2161)[0m
[31m  at org.json4s.jackson.JsonMethods$class.parse(JsonMethods.scala:19)[0m
[31m  at org.json4s.jackson.JsonMethods$.parse(JsonMethods.scala:44)[0m
[31m  ...[0m
```

vanzin:
LGTM, merging. I think that might be the test that fails sporadically, once it happens again I might take a closer look.

PR #106, opened 2016-04-13
The README file doesn't reflect the actual endpoint implemented to GE…

berngp:
…T the Batch Logs as well as the response to GET all Batches.

The current documentation has the following mistakes:
1. Points to `GET /batches/{batchId}/logs` when in fact is `GET /batches/{batchId}/log`.
2. The `GET /batches` refers to a response that contains `batches` when in fact returns `sessions`.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -485,7 +485,7 @@ Response Body
 +---------+---------------+------+
 | name    | description   | type |
 +=========+===============+======+
-| batches | `batch`_ list | list |
+|sessions | `batch`_ list | list |
 +---------+---------------+------+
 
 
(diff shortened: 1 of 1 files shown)


vanzin:
I'm gonna merge this since it's trivial enough to not really fall into "derivative work" territory... but if you plan to contribute more code in the future, please check out https://github.com/cloudera/livy/wiki/Contributing-to-Livy.

ksakellis:
@berngp We are preparing to donate this project to the ASF and just to be extra careful, can you please fill out and send us the CCLA or ICLA (or both) as described here: https://github.com/cloudera/livy/wiki/Contributing-to-Livy

PR #108, opened 2016-04-14
LIVY-112. Rename "client-local" to "client-rsc".

vanzin:
"Local" is becoming more and more confusing as more work is done
on the RSC. Let's just call it by its usual name. I chose to keep
the "client" prefix since it still provides a client API, although
that is only expected to be used by unit tests (aside from Livy
internals).


vanzin:
I'll close this and just rename the artifact to "livy-rsc".

PR #113, opened 2016-04-20
should check keytab rather than principal

zjffdu:
trivial PR


zjffdu:
close it, will fix it in https://github.com/cloudera/livy/pull/97

PR #114, opened 2016-04-20
LIVY-123. Initial framework for integration tests.

vanzin:
This change adds an initial framework for integration tests, based on the code
from https://github.com/hdinsight/livy/tree/int-test/.

It allows both real clusters and "mini" clusters (managed by the framework) to be
used. The default is to run a small mini cluster. The code also allows defining your
own cluster config file without having to modify files in the repo for that; that could be
used to run different invocations of the tests as part of the build later on, for example.

There's probably code that needs to be cleaned up or fixed up, but this provides
the basic functionality, at least for mini clusters. I haven't (or hope I haven't) modified
the real cluster code so it should be working as well as it was in the original branch.

Two blatant shortcomings of the current mini cluster is the inability to run pyspark
or SparkR tests. Because it uses a fake Spark installation, it doesn't have access
to the needed python / R libraries. That will be taken care of in separate changes.


vanzin:
Hey @tc0312, this is mostly the framework you guys created but without the stuff that doesn't work on master, and with a few extra bells & whistles. I just want to have something in master so we can start to add more tests. If you could take a look it would be great.

tc0312:
Would you mind to rebase your work on latest int-test?
It has some additional change in pom.xml and file layout to enable running integration test without maven.

vanzin:
I'll take a look.

vanzin:
@tc0312 could you point out which changes you made? You squashed all commits, and it's not obvious what changed from the previous version of the code you had.

vanzin:
@tc0312 did you do anything more than move the framework into `main` instead of `test`?

tc0312:
I also modified the pom.xml to generate a test assembly. Is it possible to rebase your work on my commit?

vanzin:
Rebasing would probably cause a lot of conflicts at this point. Let me look at the pom again.

tc0312:
If it's too much hassle, leave it for now, I can add it back when I submit the PR for test clean up.

vanzin:
That sounds good, thanks. I was trying it locally with the code in your branch merged into mine, but the build fails when I run "mvn -pl :livy-integration-test"; it works fine if I build everything from the beginning. So there's probably some kinks to iron out with that code.

vanzin:
Ok, I think I cleaned up all the unused stuff I could find. I'll push this soon so that the important part (actually getting more tests in) can be unblocked.

PR #115, opened 2016-04-22
LIVY-125. Refactor existing jobs into a new library.

vanzin:
This will help with writing integration tests, by avoiding
having to create new versions of these simple jobs. It also
serves as a nice cleanup of the existing code.

I also removed the async job from the rsc tests, since the
support code for dealing with async Spark jobs has been
removed from Livy.


vanzin:
Merging to master.

PR #117, opened 2016-04-25
LIVY-114. Merge interactive and client sessions.

vanzin:
The code to handle both session types was very, very similar. So instead
of having two completely separate endpoints, merge them, and allow sessions
to be used both in interactive mode and to run pre-compiled jobs. Note
that pre-compiled jobs are still restricted to Scala sessions.

Most of this change is code motion; I also cleaned up the repl package
hierarchy since there was no need to a bunch of different packages containing
a single class each.

The unit tests for interactive and pre-compiled jobs are still kept separate
in the server module, since the former uses mock sessions while the latter
requires a proper running Spark to run.


vanzin:
Merging to unblock other work.

PR #118, opened 2016-04-27
LIVY-127. Add integration tests for the public job API.

vanzin:
A few changes in other areas were made:

- Always overwrite the URI used to create the RSC in InteractiveSession,
  otherwise things fail.
- Removed the "timeout" from the session creation structure, because the
  existing interactive session API doesn't have it. We'll add it back
  somehow (probably as a configuration stuck in the "conf" map).
- Removed the logging from LivyClientBuilder and just let exceptions from
  the factories propagate; this removes the logging dependency, which also
  avoids having to install the jul-to-slf4j bridge to avoid ugly log messages
  in test output. As a bonus, it makes the exceptions from the builder less
  cryptic.
- Renamed a few classes and methods to make things cleaner and also to avoid
  a naming clash.
- Removed some dead code and dependencies.

### DIFF
--- a/api/src/main/java/com/cloudera/livy/LivyClientBuilder.java
+++ b/api/src/main/java/com/cloudera/livy/LivyClientBuilder.java
@@ -26,8 +26,6 @@ import java.net.URL;
 import java.util.Map;
 import java.util.Properties;
 import java.util.ServiceLoader;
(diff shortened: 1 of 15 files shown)


vanzin:
Merging to unblock next PR.

PR #132, opened 2016-05-12
LIVY-94. SQLContext/HiveContext is not available in Interpreter

zjffdu:
It is straightforward to add SQLContext. But for HiveContext is a little complicated, because it require hive-site.xml and datanucleus jars. There's 2 ways to specify that
1. Ask the client to specify that in rest api, but client should have no idea where the hive-site.xml and datanucleus jars are located in the livy server machine. Although they put these files on hdfs and specify them using hdfs path. But this way still have one issue in yarn-cluster mode due to SPARK-14845
2. Add a option like enableHiveContext in rest api and livy server pick up the hive-site.xml and datanucleus jars automatically. But this way has limitation that it always use the same hive-site.xml, prevent user to user another custom hive-site.xml if he want to connect to another hive metastore.
In this PR, I just use option 1.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -661,7 +661,7 @@
             </systemProperties>
             <redirectTestOutputToFile>${test.redirectToFile}</redirectTestOutputToFile>
             <useFile>${test.redirectToFile}</useFile>
-            <argLine>${argLine} -Xmx1g -XX:MaxPermSize=256m</argLine>
+            <argLine>${argLine} -Xmx2g -XX:MaxPermSize=512m</argLine>
             <failIfNoTests>false</failIfNoTests>
           </configuration>
         </plugin>
@@ -688,7 +688,7 @@
             <reportsDirectory>${project.build.directory}/surefire-reports</reportsDirectory>
             <junitxml>.</junitxml>
             <filereports>WDF TestSuite.txt</filereports>
-            <argLine>${argLine} -Xmx1g -XX:MaxPermSize=256m</argLine>
+            <argLine>${argLine} -Xmx2g -XX:MaxPermSize=512m</argLine>
           </configuration>
           <executions>
             <execution>
--- a/repl/src/main/scala/com/cloudera/livy/repl/SparkInterpreter.scala
+++ b/repl/src/main/scala/com/cloudera/livy/repl/SparkInterpreter.scala
@@ -349,7 +349,7 @@ class SparkInterpreter(conf: SparkConf) extends Interpreter with Logging {
                         // Remove Interpreter frames
                         .take(interpreterFrameIdx)
                         // Replace weird internal class name
-                        .map(_.replace("$iwC$$iwC", "<user code>"))
+                        .map(_.replaceAll("(\\$iwC\\$)*\\$iwC", "<user code>"))
                       // TODO Proper translate line number in stack trace for $iwC$$iwC.
                     }
                     (ename.trim, traceback)
--- a/repl/src/main/scala/com/cloudera/livy/repl/SparkInterpreter.scala
+++ b/repl/src/main/scala/com/cloudera/livy/repl/SparkInterpreter.scala
@@ -121,20 +121,23 @@ class SparkInterpreter(conf: SparkConf) extends Interpreter with Logging {
       }
 
       sparkContext = SparkContext.getOrCreate(conf)
-      if (conf.getBoolean("spark.repl.enableHiveContext", false)) {
+      if (conf.getBoolean("livy.repl.enableHiveContext", false)) {
         try {
           val loader = Option(Thread.currentThread().getContextClassLoader)
             .getOrElse(getClass.getClassLoader)
+          if (loader.getResource("hive-site.xml") == null) {
+            warn("livy.repl.enableHiveContext is true but no hive-site.xml found on classpath.")
+          }
           sqlContext = new HiveContext(sparkContext)
-          info("Created sql context (with Hive support)..")
+          info("Created sql context (with Hive support).")
         } catch {
           case _: java.lang.NoClassDefFoundError =>
             sqlContext = new SQLContext(sparkContext)
-            info("Created sql context..")
+            info("Created sql context.")
         }
       } else {
         sqlContext = new SQLContext(sparkContext)
-        info("Created sql context..")
+        info("Created sql context.")
       }
       sparkIMain.beQuietDuring {
         sparkIMain.bind("sc", "org.apache.spark.SparkContext", sparkContext, List("""@transient"""))
--- a/repl/src/main/scala/com/cloudera/livy/repl/SparkInterpreter.scala
+++ b/repl/src/main/scala/com/cloudera/livy/repl/SparkInterpreter.scala
@@ -29,6 +29,8 @@ import scala.util.{Failure, Success, Try}
 import org.apache.spark.{SparkConf, SparkContext}
 import org.apache.spark.rdd.RDD
 import org.apache.spark.repl.SparkIMain
+import org.apache.spark.sql.hive.HiveContext
+import org.apache.spark.sql.SQLContext
 import org.json4s.{DefaultFormats, Extraction}
 import org.json4s.JsonAST._
 import org.json4s.JsonDSL._
@@ -53,6 +55,7 @@ class SparkInterpreter(conf: SparkConf) extends Interpreter with Logging {
   private val outputStream = new ByteArrayOutputStream()
   private var sparkIMain: SparkIMain = _
   private var sparkContext: SparkContext = _
+  private var sqlContext: SQLContext = _
 
   def kind: String = "spark"
 
@@ -118,12 +121,33 @@ class SparkInterpreter(conf: SparkConf) extends Interpreter with Logging {
       }
 
       sparkContext = SparkContext.getOrCreate(conf)
-
+      if (conf.getBoolean("spark.repl.enableHiveContext", false)) {
+        try {
+          val loader = Option(Thread.currentThread().getContextClassLoader)
+            .getOrElse(getClass.getClassLoader)
+          sqlContext = new HiveContext(sparkContext)
+          info("Created sql context (with Hive support)..")
+        } catch {
+          case _: java.lang.NoClassDefFoundError =>
+            sqlContext = new SQLContext(sparkContext)
+            info("Created sql context..")
+        }
+      } else {
+        sqlContext = new SQLContext(sparkContext)
+        info("Created sql context..")
+      }
       sparkIMain.beQuietDuring {
         sparkIMain.bind("sc", "org.apache.spark.SparkContext", sparkContext, List("""@transient"""))
       }
+      sparkIMain.beQuietDuring {
+        sparkIMain.bind("sqlContext", sqlContext.getClass.getCanonicalName,
+          sqlContext, List("""@transient"""))
+      }
+      execute("import org.apache.spark.SparkContext._")
+      execute("import sqlContext.implicits._")
+      execute("import sqlContext.sql")
+      execute("import org.apache.spark.sql.functions._")
     }
-
     sparkContext
   }
 
@@ -144,6 +168,7 @@ class SparkInterpreter(conf: SparkConf) extends Interpreter with Logging {
       sparkIMain.close()
       sparkIMain = null
     }
+    sqlContext = null
   }
 
   private def executeMagic(magic: String, rest: String): Interpreter.ExecuteResponse = {
--- a/repl/src/main/scala/com/cloudera/livy/repl/SparkInterpreter.scala
+++ b/repl/src/main/scala/com/cloudera/livy/repl/SparkInterpreter.scala
@@ -121,7 +121,7 @@ class SparkInterpreter(conf: SparkConf) extends Interpreter with Logging {
       }
 
       sparkContext = SparkContext.getOrCreate(conf)
-      if (conf.getBoolean("livy.repl.enableHiveContext", false)) {
+      if (conf.getBoolean("spark.repl.enableHiveContext", false)) {
         try {
           val loader = Option(Thread.currentThread().getContextClassLoader)
             .getOrElse(getClass.getClassLoader)
--- a/server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
+++ b/server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
@@ -41,7 +41,7 @@ class LivyServer extends Logging {
   private var server: WebServer = _
   private var _serverUrl: Option[String] = None
   // make livyConf accessible for testing
-  var livyConf: LivyConf = _
+  private[livy] var livyConf: LivyConf = _
 
   def start(): Unit = {
     livyConf = new LivyConf().loadFromFile("livy.conf")
--- a/server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
+++ b/server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
@@ -40,9 +40,11 @@ class LivyServer extends Logging {
 
   private var server: WebServer = _
   private var _serverUrl: Option[String] = None
+  // make livyConf accessible for testing
+  var livyConf: LivyConf = _
 
   def start(): Unit = {
-    val livyConf = new LivyConf().loadFromFile("livy.conf")
+    livyConf = new LivyConf().loadFromFile("livy.conf")
     val host = livyConf.get(SERVER_HOST)
     val port = livyConf.getInt(SERVER_PORT)
     val multipartConfig = MultipartConfig(
--- a/server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
+++ b/server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
@@ -86,23 +86,19 @@ class InteractiveSession(
 
     def mergeHiveSiteAndHiveDeps(): Unit = {
       val sparkFiles = conf.get("spark.files").map(_.split(",")).getOrElse(Array.empty[String])
-      if (livyConf.getBoolean(LivyConf.ENABLE_HIVE_CONTEXT)) {
-        hiveSiteFile(sparkFiles, livyConf) match {
-          case (_, true) =>
-            info("Enable HiveContext because hive-site.xml is found in user request.")
-            mergeConfList(datanucleusJars(livyConf), LivyConf.SPARK_JARS)
-          case (Some(file), false) =>
-            info("Enable HiveContext because hive-site.xml is found under classpath, "
-              + file.getAbsolutePath)
-            mergeConfList(List(file.getAbsolutePath), LivyConf.SPARK_FILES)
-            mergeConfList(datanucleusJars(livyConf), LivyConf.SPARK_JARS)
-          case (None, false) =>
-            warn("Enable HiveContext but no hive-site.xml found under" +
-              " classpath and user request.")
-        }
+      hiveSiteFile(sparkFiles, livyConf) match {
+        case (_, true) =>
+          debug("Enable HiveContext because hive-site.xml is found in user request.")
+          mergeConfList(datanucleusJars(livyConf), LivyConf.SPARK_JARS)
+        case (Some(file), false) =>
+          debug("Enable HiveContext because hive-site.xml is found under classpath, "
+            + file.getAbsolutePath)
+          mergeConfList(List(file.getAbsolutePath), LivyConf.SPARK_FILES)
+          mergeConfList(datanucleusJars(livyConf), LivyConf.SPARK_JARS)
+        case (None, false) =>
+          warn("Enable HiveContext but no hive-site.xml found under" +
+            " classpath or user request.")
       }
-      builderProperties.put("spark.repl.enableHiveContext",
-        livyConf.getBoolean(LivyConf.ENABLE_HIVE_CONTEXT).toString)
     }
 
     kind match {
@@ -120,7 +116,12 @@ class InteractiveSession(
     builderProperties.put(RSCConf.Entry.SESSION_KIND.key, kind.toString)
 
     mergeConfList(livyJars(livyConf), LivyConf.SPARK_JARS)
-    mergeHiveSiteAndHiveDeps()
+    val enableHiveContext = livyConf.getBoolean(LivyConf.ENABLE_HIVE_CONTEXT)
+    builderProperties.put("spark.repl.enableHiveContext",
+      livyConf.getBoolean(LivyConf.ENABLE_HIVE_CONTEXT).toString)
+    if (enableHiveContext) {
+      mergeHiveSiteAndHiveDeps()
+    }
 
     val userOpts: Map[String, Option[String]] = Map(
       "spark.driver.cores" -> request.driverCores.map(_.toString),
--- a/server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
+++ b/server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
@@ -326,7 +326,6 @@ class InteractiveSession(
    * Look for hive-site.xml (for now just ignore spark.files defined in spark-defaults.conf)
    * 1. First look for hive-site.xml in user request
    * 2. Then look for that under classpath
-   * 3. Finally look for that under SPARK_HOME/conf
    * @param livyConf
    * @return  (hive-site.xml path, whether it is provided by user)
    */
@@ -337,16 +336,10 @@ class InteractiveSession(
     } else {
       val hiveSiteURL = getClass.getResource("/hive-site.xml")
       if (hiveSiteURL != null && hiveSiteURL.getProtocol == "file") {
-        return (Some(new File(hiveSiteURL.toURI)), false)
+        (Some(new File(hiveSiteURL.toURI)), false)
+      } else {
+        (None, false)
       }
-      // only check SPARK_HOME/conf when SPARK_CONF_DIR is not defined
-      if (!sys.env.contains("SPARK_CONF_DIR")) {
-        val hiveSiteFile = new File(livyConf.sparkHome().get + "/conf/hive-site.xml")
-        if (hiveSiteFile.isFile) {
-          return (Some(hiveSiteFile), false)
-        }
-      }
-      (None, false)
     }
   }
 
--- a/server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
+++ b/server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
@@ -27,7 +27,6 @@ import java.util.concurrent.atomic.AtomicLong
 import scala.annotation.tailrec
 import scala.collection.JavaConverters._
 import scala.collection.mutable
-import scala.collection.mutable.ArrayBuffer
 import scala.concurrent.{Future, _}
 import scala.util.{Failure, Success, Try}
 
@@ -45,7 +44,6 @@ import com.cloudera.livy.sessions._
 object InteractiveSession {
   val LivyReplJars = "livy.repl.jars"
   val SparkYarnIsPython = "spark.yarn.isPython"
-  val EnableHiveContext = "spark.repl.enableHiveContext"
 }
 
 class InteractiveSession(
@@ -87,22 +85,24 @@ class InteractiveSession(
     }
 
     def mergeHiveSiteAndHiveDeps(): Unit = {
-      hiveSiteFile(request.files, livyConf) match {
-        case (_, true) =>
-          info("Enalble HiveContext because hive-site.xml is found in user request.")
-          mergeConfList(datanucleusJars(livyConf), LivyConf.SPARK_JARS)
-          builderProperties.put(EnableHiveContext, "true")
-        case (Some(file), false) =>
-          info("Enalble HiveContext because hive-site.xml is found under classpath, "
-            + file.getAbsolutePath)
-          mergeConfList(List(file.getAbsolutePath), LivyConf.SPARK_FILES)
-          mergeConfList(datanucleusJars(livyConf), LivyConf.SPARK_JARS)
-          builderProperties.put(EnableHiveContext, "true")
-        case (None, false) =>
-          info("Disable HiveContext because no hive-site.xml found under" +
-            " classpath and user request.")
-          builderProperties.put(EnableHiveContext, "false")
+      val sparkFiles = conf.get("spark.files").map(_.split(",")).getOrElse(Array.empty[String])
+      if (livyConf.getBoolean(LivyConf.ENABLE_HIVE_CONTEXT)) {
+        hiveSiteFile(sparkFiles, livyConf) match {
+          case (_, true) =>
+            info("Enable HiveContext because hive-site.xml is found in user request.")
+            mergeConfList(datanucleusJars(livyConf), LivyConf.SPARK_JARS)
+          case (Some(file), false) =>
+            info("Enable HiveContext because hive-site.xml is found under classpath, "
+              + file.getAbsolutePath)
+            mergeConfList(List(file.getAbsolutePath), LivyConf.SPARK_FILES)
+            mergeConfList(datanucleusJars(livyConf), LivyConf.SPARK_JARS)
+          case (None, false) =>
+            warn("Enable HiveContext but no hive-site.xml found under" +
+              " classpath and user request.")
+        }
       }
+      builderProperties.put(LivyConf.ENABLE_HIVE_CONTEXT.key,
+        livyConf.getBoolean(LivyConf.ENABLE_HIVE_CONTEXT).toString)
     }
 
     kind match {
@@ -301,7 +301,6 @@ class InteractiveSession(
       // datanucleus jars has already been in classpath in integration test
       Seq.empty
     } else {
-      val jars = ArrayBuffer.empty[String]
       val sparkHome = livyConf.sparkHome().get
       val libdir =
         if (new File(sparkHome, "RELEASE").isFile) {
@@ -311,11 +310,12 @@ class InteractiveSession(
         }
 
       if (libdir.isDirectory()) {
-        jars ++= libdir.listFiles().filter(_.getName.startsWith("datanucleus-"))
+        libdir.listFiles().filter(_.getName.startsWith("datanucleus-"))
           .map(_.getAbsolutePath)
+      } else {
+        warn("datanucleus jars can not be found")
+        Seq.empty[String]
       }
-      warn("datanucleus jars can not be found")
-      jars.toSeq
     }
   }
 
@@ -327,18 +327,21 @@ class InteractiveSession(
    * @param livyConf
    * @return  (hive-site.xml path, whether it is provided by user)
    */
-  private def hiveSiteFile(requestFiles: List[String],
+  private def hiveSiteFile(sparkFiles: Array[String],
                            livyConf: LivyConf): (Option[File], Boolean) = {
-    if (requestFiles.exists(_.split("/").last == "hive-site.xml")) {
+    if (sparkFiles.exists(_.split("/").last == "hive-site.xml")) {
       (None, true)
     } else {
       val hiveSiteURL = getClass.getResource("/hive-site.xml")
       if (hiveSiteURL != null && hiveSiteURL.getProtocol == "file") {
         return (Some(new File(hiveSiteURL.toURI)), false)
       }
-      val hiveSiteFile = new File(livyConf.sparkHome().get + "/conf/hive-site.xml")
-      if (hiveSiteFile.isFile) {
-        return (Some(hiveSiteFile), false)
+      // only check SPARK_HOME/conf when SPARK_CONF_DIR is not defined
+      if (!sys.env.contains("SPARK_CONF_DIR")) {
+        val hiveSiteFile = new File(livyConf.sparkHome().get + "/conf/hive-site.xml")
+        if (hiveSiteFile.isFile) {
+          return (Some(hiveSiteFile), false)
+        }
       }
       (None, false)
     }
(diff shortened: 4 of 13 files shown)


vanzin:
@zjffdu this PR has conflicts and can't be tested in its current state

pom.xml
```diff
@@ -652,7 +652,7 @@
             </systemProperties>
             <redirectTestOutputToFile>${test.redirectToFile}</redirectTestOutputToFile>
             <useFile>${test.redirectToFile}</useFile>
-            <argLine>${argLine} -Xmx1g -XX:MaxPermSize=256m</argLine>
+            <argLine>${argLine} -Xmx4g -XX:MaxPermSize=1024m</argLine>
```
vanzin: Please don't change these; the travis machines are not very large and this might cause memory pressure. If the extra memory is really needed we should investigate why and, if it's really needed, change the travis config also to use larger machines if possible.

repl/src/main/scala/com/cloudera/livy/repl/SparkInterpreter.scala
```diff
@@ -84,11 +87,33 @@ class SparkInterpreter(conf: SparkConf) extends Interpreter {
...
+        .getOrElse(getClass.getClassLoader)
+      val sqlContextClassName = "org.apache.spark.sql.hive.HiveContext"
+      sqlContext = loader.loadClass(sqlContextClassName).getConstructor(classOf[SparkContext])
+        .newInstance(sparkContext).asInstanceOf[SQLContext]
+      info("Created sql context (with Hive support)..")
+    } catch {
+      case _: java.lang.ClassNotFoundException | _: java.lang.NoClassDefFoundError =>
+        sqlContext = new SQLContext(sparkContext)
+        info("Created sql context..")
+      case _ if sys.env.getOrElse("LIVY_TEST", "false") == "true" =>
```
vanzin: There are constants for this (`LivyConf.TEST_MODE` and `ClientConf.TEST_MODE`).

vanzin: Also why would you get an exception here during tests? The hive code is there, it should work.

server/src/main/scala/com/cloudera/livy/server/interactive/CreateInteractiveRequest.scala
```diff
@@ -36,5 +36,5 @@ class CreateInteractiveRequest {
   var queue: Option[String] = None
   var name: Option[String] = None
   var conf: Map[String, String] = Map()
-
+  var addHiveDeps: Boolean = false
```
vanzin: This is touching a public API... while adding fields is not bad, I'm not sure this is the right way to do it.

The right way would be to read the Hive conf and see whether the datanucleus jars are needed (which is basically what this option translates to). Forcing users to do this is not very user-friendly.

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
```diff
@@ -266,6 +274,36 @@ class InteractiveSession(
...
+    val sparkHome = livyConf.sparkHome().get
+    val libdir =
+      if (new File(sparkHome, "RELEASE").isFile) {
+        new File(sparkHome, "lib")
+      } else {
+        new File(sparkHome, "lib_managed/jars")
+      }
+
+    if (libdir.isDirectory()) {
+      jars ++= libdir.listFiles().filter(_.getName.startsWith("datanucleus-"))
```
vanzin: Instead of using a mutable buffer, just return regular seqs.

```
if (isdir) {
  dir.listFiles()......map { }.toSeq
} else {
  Nil
}

**Comment by vanzin** [comment]
One issue here is that users might already have code that instantiates a variable called `sqlContext` and this might be conflicting with that. Can you make sure that case still works? A test for that would be nice. Otherwise you'll need a flag so that users can explicit ask for a SQLContext when creating sessions.

Barring that issue, I'd rather see Livy be smarter about the Hive configuration. It should be able to read the Hive configuration (which can either be in the classpath, or using the code you're adding to find `hive-site.xml`) and figure out whether it needs to add the datanucleus jars by itself, without the user having to request those.

File: server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala [hunk 5]
```diff
@@ -90,16 +91,23 @@ class InteractiveSession(
     }
     builder.setConf(RSCConf.Entry.SESSION_KIND.key, kind.toString)
 
-    val allJars = livyJars(livyConf) ++ request.jars
+    val allJars = ArrayBuffer((livyJars(livyConf) ++ request.jars): _*)
```
vanzin: This code will change a lot when I push an outstanding PR, so you'll have conflicts to resove; while doing that I'd encourage you to try to avoid mutable data structures, they tend to make the code less readable.

pom.xml
zjffdu: This is due to initialize HiveContext cost much memory.

zjffdu:
@vanzin  Overall, I would like to make livy interactive mode to be consistent with spark-shell, pyspark-shell and sparkR-shell. So SparkContext/SQLContext(HiveContext) would be created automatically.  What I think now is that livy first check whether hive-site.xml is specified in rest api, if yes, then pick up the datanucleus jars, if not then check whether there's hive-site.xml in $SPARK_CONF_DIR, if yes, pick up datanucleus jars.  And livy server will pass one flag whether hive-site.xml is found to interpreter, so that interpreter know whether it is necessary to create HiveContext.  What do you think ?

vanzin:
The approach to detect hive-site.xml sounds fine; I'd just include the classpath also in the list of places to look. And optimally, you should peek into the file to see whether the datanucleus jars are needed (i.e. if `hive.metastore.uris` is set, then they're not needed). Also be aware that some distributions (e.g. CDH) do not include the datanucleus jars in `SPARK_HOME`, so Livy shouldn't fail if they're not found.

In fact, I'm not sure you should really look at the list of files sent by the user; distributing `hive-site.xml` that way only works in yarn-cluster mode, and it sounds weird for users to be using their own Hive configuration. But that's a minor thing, in the end.

I assume that you comment means you'd still be creating a `SQLContext` if the Hive config is not present. My only concern is what I mentioned before, that this may have side-effects for user code that already creates one; you should add a test that manually instantiates a `SQLContext` after your automatic instantiation happens, and see that it works fine.

zjffdu:
Didn't find why the CI fails.

```
[INFO] 
[INFO] livy-main .......................................... SUCCESS [ 21.907 s]
[INFO] livy-api ........................................... SUCCESS [01:28 min]
[INFO] livy-client-common ................................. SUCCESS [  3.999 s]
[INFO] livy-test-lib ...................................... SUCCESS [ 12.440 s]
[INFO] livy-rsc ........................................... SUCCESS [ 15.653 s]
[INFO] livy-core .......................................... SUCCESS [ 14.934 s]
[INFO] livy-repl .......................................... SUCCESS [ 57.651 s]
[INFO] livy-server ........................................ SUCCESS [01:06 min]
[INFO] livy-assembly ...................................... SUCCESS [  5.612 s]
[INFO] livy-client-http ................................... SUCCESS [ 21.292 s]
[INFO] livy-integration-test .............................. SUCCESS [ 41.065 s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 05:49 min
[INFO] Finished at: 2016-05-24T03:58:38+00:00
[INFO] Final Memory: 48M/471M
[INFO] ------------------------------------------------------------------------

```

vanzin:
The error from the log is:

```
^[[31m*** RUN ABORTED *** (4 minutes, 48 seconds)^[[0m^M
^[[31m  java.lang.OutOfMemoryError: GC overhead limit exceeded^[[0m^M
```

Plus a couple of timeouts in the same suite, probably because of gc activity.

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
```diff
@@ -80,8 +80,22 @@ class InteractiveSession(
 
     def mergeConfList(list: Seq[String], key: String): Unit = {
       if (list.nonEmpty) {
-        val newList = (list ++ conf.get(key)).mkString(",")
-        builder.setConf(key, newList)
+        val oldList = builder.getConf(key)
```
vanzin: You don't need to expose `getConf` in the public API for this. There's a `conf` map above in this method with all the user configuration. Just modify it before creating the client.

repl/src/main/scala/com/cloudera/livy/repl/SparkInterpreter.scala
```diff
@@ -48,6 +50,7 @@ class SparkInterpreter(conf: SparkConf) extends Interpreter {
   private val outputStream = new ByteArrayOutputStream()
   private var sparkIMain: SparkIMain = _
   private var sparkContext: SparkContext = _
+  private var sqlContext: SQLContext = _
```
vanzin: It's probably a good idea to set this variable to `null` in the `close` method (since `SQLContext` doesn't have a stop or close or anything).

vanzin:
I ran the livy-repl tests with your branch and took a heap dump when it hit the OOM, and I see a lot of live `SparkIMain` instances holding on to a whole lot of memory. Something is keeping them around, let me see how hard it is to fix that.

vanzin:
@zjffdu the main reason for the test failures is that your code is instantiating a HiveContext even when no `hive-site.xml` file exists; since multiple of those are instantiated in the same VM during unit tests, you end up with multiple instances of derby and other things in memory, and that ends up consuming a lot of memory. And it doesn't seem like there's an easy way to shut down those instances.

I thought you were going to change things so that HiveContext would only be instantiated if a `hive-site.xml` file was found?

The HiveContext-specific parts of the change should be a separate unit test; or perhaps even an integration test since each test there runs a separate Spark session in a different VM (so memory issues like these are less prone to happen).

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
zjffdu: The reason I expose LivyClientBuilder.getConf is that mergeConfList will have issue when being called multiple times for the same key.  e.g. I want to call mergeConfList multiple times for spark.jars, For the first call, it would update builder conf correctly, but for the second call, it would lose the first update because it didn't get the old values from LivyClientBuilder.

vanzin: That sounds like a problem that can be solved without changing the `LivyClientBuilder`. I just don't want to expose public methods (and `LivyClientBuilder` is a public class in Livy) that are not really necessary.

zjffdu:
I do see multiple instance of derby in the log, that's why I will fallback to create SQLContext in test. But looks like it is not enough for that. 
```
} catch {
      case _: java.lang.ClassNotFoundException | _: java.lang.NoClassDefFoundError =>
        sqlContext = new SQLContext(sparkContext)
        info("Created sql context..")
      case _ if sys.env.getOrElse("LIVY_TEST", "false") == "true" =>
        sqlContext = new SQLContext(sparkContext)
        info("Created sql context..")
    }
```

zjffdu:
It's weird that the test can pass after I change the integration log level to INFO

zjffdu:
Not build is trigged ?

vanzin:
@zjffdu builds don't trigger when you have merge conflicts.

zjffdu:
The test pass in my local box, but always fails on travis ci.  @vanzin  Is it possible to see the livy log and yarn log of the integration test in travis ci ?

vanzin:
Unfortunately travis doesn't really give you a way to retrieve anything other than the terminal output of the build. Anyway, the error you're hitting is LIVY-154, so most probably unrelated to your code.

vanzin:
Hi @zjffdu,

I can consistently hit the problem on my laptop; that and travis being unhappy make me wary of merging this change right now. I'll take a closer look at LIVY-154 and try to fix it, since I don't see any other issues while the tests are running - gc activity seems pretty low so it's not like the processes are starving for resources.

zjffdu:
Patch rebased, and increase timeout of 2 UT to avoid intermittent test failure.

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
```diff
@@ -70,36 +71,48 @@ class InteractiveSession(
...
+            builderProperties.put(key, newList)
+        }
+      }
+    }
+
+    def mergeHiveSiteAndHiveDeps(): Unit = {
+      hiveSiteFile(livyConf) match {
+        case Some(file) =>
+          mergeConfList(List(file.getAbsolutePath), LivyConf.SPARK_FILES)
+          mergeConfList(datanucleusJars(livyConf), LivyConf.SPARK_JARS)
```
vanzin: Does this work when distributed with `--jars`? I thought datanucleus needed to be in the system classpath. We always use a metastore server here so haven't done any testing of those jars.

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
```diff
@@ -268,6 +288,51 @@ class InteractiveSession(
     }
   }
 
+  private def datanucleusJars(livyConf: LivyConf): ArrayBuffer[String] = {
```
vanzin: Please use immutable lists.

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
```diff
@@ -268,6 +288,51 @@ class InteractiveSession(
...
+          new File(sparkHome, "lib")
+        } else {
+          new File(sparkHome, "lib_managed/jars")
+        }
+
+      if (libdir.isDirectory()) {
+        jars ++= libdir.listFiles().filter(_.getName.startsWith("datanucleus-"))
+          .map(_.getAbsolutePath)
+      }
+      require(jars.nonEmpty, "datanucleus jars can not be found")
```
vanzin: This will break CDH since it doesn't ship datanucleus jars in Spark's lib directory. This shouldn't be a requirement anyway since most people use proper metastore servers (or at least most people I know).

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
```diff
@@ -268,6 +288,51 @@ class InteractiveSession(
...
+   * 2. If not found under classpath, then look for that under SPARK_CONF_DIR and SPARK_HOME/conf
+   * @param livyConf
+   * @return
+   */
+  private def hiveSiteFile(livyConf: LivyConf): Option[File] = {
+    val hiveSiteURL = getClass.getResource("/hive-site.xml")
+    if (hiveSiteURL != null && hiveSiteURL.getProtocol == "file") {
+      return Some(new File(hiveSiteURL.toURI))
+    }
+    val hiveSiteFile = sys.env.get("SPARK_CONF_DIR")
```
vanzin: These checks are redundant; `bin/livy-server` already adds all these directories to the classpath, so the check above will already find the file.

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
```diff
@@ -268,6 +288,51 @@ class InteractiveSession(
...
+      .map(conf_dir => new File(conf_dir + "/hive-site.xml"))
+      .getOrElse(new File(livyConf.sparkHome().get + "/conf/hive-site.xml"))
+    if (hiveSiteFile.isFile) {
+      return Some(hiveSiteFile)
+    }
+    warn("Can't find hive-site.xml under SPARK_CONF_DIR, SPARK_HOME/conf and livy's CLASSPATH")
```
vanzin: Shouldn't be a warning, but see above.

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
```diff
@@ -70,36 +71,48 @@ class InteractiveSession(
...
+            builderProperties.put(key, list.mkString(","))
+          case Some(oldList) =>
+            val newList = (oldList :: list.toList).mkString(",")
+            builderProperties.put(key, newList)
+        }
+      }
+    }
+
+    def mergeHiveSiteAndHiveDeps(): Unit = {
+      hiveSiteFile(livyConf) match {
```
vanzin: This checks the Livy configuration, but not the user's session config; in yarn-cluster mode, `--files` can be used to distribute a custom hive-site.xml. Since it's a little tricky to figure out the master configuration here, and it really only works in yarn-cluster mode, I think it's ok to ignore that for now.

vanzin:
Hey @zjffdu, there's one issue that this change might cause when using the programmatic API for jobs. You can use the `JobContext` API to create instances of `SQLContext` or `HiveContext`, and with these pre-existing instances you might be wasting memory and even running into problems if you use that API.

It would be better to hook the `JobContext` to use these instances when they've already been created. But feel free to file a bug to fix it separately instead of changing this PR.

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
zjffdu: HiveContext will fail to start when it is using metastore server without datanucleus jars. I suspect this might be spark issue, because from the logs it seems spark will still initialize embedded derby even when I configure to use metastore server. I will look at it later. 
```
16/06/11 08:38:40 INFO MetaStoreDirectSql: Using direct SQL, underlying DB is DERBY
```

vanzin: Hmmm... might be that we changed that behavior in CDH. But we don't have those jars (not in that location anyway), so this assert would hit every time on CDH, so I'd rather not have it.

zjffdu: I just checked this issue only happens in spark-1.6.1, but not in spark-2.0, might be resolved in some ticket.

zjffdu: To make it work with both CDH and other distribution, I may need to detect whether it is CDH. Does CDH change spark's version to include 'cdh' ?

zjffdu:
@vanzin I just post another commit.
* For datanucleus jars missing, I will only print warn message instead of failing the app.
* Regarding the issue with JavaContext, I think it is fine for now because JavaContext is not exposed in InteractiveSession (only used internally) and this PR only create SQLContext/HiveContext in InteractiveSession.

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
```diff
@@ -268,6 +296,54 @@ class InteractiveSession(
     }
   }
 
+  private def datanucleusJars(livyConf: LivyConf): Seq[String] = {
+    if (sys.env.getOrElse("LIVY_INTEGRATION_TEST", "false").toBoolean) {
+      // datanucleus jars has already been in classpath in integration test
+      Seq.empty
+    } else {
+      val jars = ArrayBuffer.empty[String]
```
vanzin: This doesn't need to be a mutable list.

```
val jars = if (libdir.isDirectory()) {
  libdir.listFiles().filter(_.getName.startsWith("datanucleus-")).map(_.getAbsolutePath)
} else {
  Nil
}
```

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
```diff
@@ -70,36 +71,56 @@ class InteractiveSession(
...
+            builderProperties.put(key, list.mkString(","))
+          case Some(oldList) =>
+            val newList = (oldList :: list.toList).mkString(",")
+            builderProperties.put(key, newList)
+        }
+      }
+    }
+
+    def mergeHiveSiteAndHiveDeps(): Unit = {
+      hiveSiteFile(request.files, livyConf) match {
```
vanzin: `request.files` is not enough; user might have set `spark.files` in the configuration instead. L71 will merge the two lists, so you can just directly look at the configuration to find everything.

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
```diff
@@ -70,36 +71,56 @@ class InteractiveSession(
...
+        case (_, true) =>
+          info("Enalble HiveContext because hive-site.xml is found in user request.")
```
vanzin: Enable

repl/src/main/scala/com/cloudera/livy/repl/SparkInterpreter.scala
```diff
@@ -84,12 +88,33 @@ class SparkInterpreter(conf: SparkConf) extends Interpreter {
...
+      if (conf.getBoolean("spark.repl.enableHiveContext", false)) {
+        try {
+          val loader = Option(Thread.currentThread().getContextClassLoader)
+            .getOrElse(getClass.getClassLoader)
+          sqlContext = new HiveContext(sparkContext)
+          info("Created sql context (with Hive support)..")
+        } catch {
+          case _: java.lang.NoClassDefFoundError =>
+            sqlContext = new SQLContext(sparkContext)
+            info("Created sql context..")
```
vanzin: Stray period.

repl/src/main/scala/com/cloudera/livy/repl/SparkInterpreter.scala
```diff
@@ -84,12 +88,33 @@ class SparkInterpreter(conf: SparkConf) extends Interpreter {
...
+        }
+      } else {
+        sqlContext = new SQLContext(sparkContext)
+        info("Created sql context..")
```
vanzin: Stray period.

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
```diff
@@ -268,6 +296,54 @@ class InteractiveSession(
...
+  private def hiveSiteFile(requestFiles: List[String],
+                           livyConf: LivyConf): (Option[File], Boolean) = {
+    if (requestFiles.exists(_.split("/").last == "hive-site.xml")) {
+      (None, true)
+    } else {
+      val hiveSiteURL = getClass.getResource("/hive-site.xml")
+      if (hiveSiteURL != null && hiveSiteURL.getProtocol == "file") {
+        return (Some(new File(hiveSiteURL.toURI)), false)
+      }
+      val hiveSiteFile = new File(livyConf.sparkHome().get + "/conf/hive-site.xml")
```
vanzin: This is not needed and actually wrong. If the user has set `SPARK_CONF_DIR`, he explicitly doesn't want Livy to look at `$SPARK_HOME/conf`. Since `SPARK_CONF_DIR` is in the classpath, the previous check suffices.

vanzin:
A few minor things that need looking after. There's also some weirdness about how Spark behaves with "hive-site.xml" in general; for example, distributing it with `--files` only really works with yarn-cluster mode, IIRC. It won't do anything in client mode, or with other cluster managers.

My personal preference is that this should be left for the admin to figure out. e.g., if he wants Livy sessions to have Hive access, place hive-site.xml in `HADOOP_CONF_DIR`, and Spark will take care of uploading it to the cluster and make everything work. If you feel that other use cases are important is fine, but that restriction would make your code a lot simpler.

repl/src/main/scala/com/cloudera/livy/repl/SparkInterpreter.scala
```diff
@@ -84,12 +88,33 @@ class SparkInterpreter(conf: SparkConf) extends Interpreter {
       setContextClassLoaderMethod.invoke(sparkIMain)
 
       sparkContext = SparkContext.getOrCreate(conf)
-
+      if (conf.getBoolean("spark.repl.enableHiveContext", false)) {
```
vanzin: You can replace this with a check that `hive-site.xml` is in the application's classpath, which is basically the only way this will work anyway. (See my earlier comment about distributing `hive-site.xml`.)

zjffdu:
> place hive-site.xml in HADOOP_CONF_DIR, and Spark will take care of uploading it to the cluster and make everything work.

I don't think this is a good idea, because it may affect other applications as lots of hadoop downstream project will put HADOOP_CONF_DIR on classpath, while they may don't expect hive-site there and want to provide hive-site in other places. In the perspective of administration, I think I can add one flag option of whether enable HiveContext automatically. If this flag is enabled. then hive-site.xml is automatically detected in livy server, otherwise only client can enable hivecontext by providing hive-site.xml, What do you think ?

vanzin:
> lots of hadoop downstream project will put HADOOP_CONF_DIR on classpath, while they may don't expect hive-site there

There's no reason why you need to have the same HADOOP_CONF_DIR for everybody. It's trivial to create a specific one for Livy. Any decent admin tool will allow you to even automate that process.

zjffdu:
We could do that, but it would bring additional maintenance overhead IMHO.

tc0312:
Do people put hive-site.xml in HADOOP_CONF_DIR? I thought most of them put it in HIVE_CONF_DIR.

> There's no reason why you need to have the same HADOOP_CONF_DIR for everybody. It's trivial to create a specific one for Livy. Any decent admin tool will allow you to even automate that process.

Users have to clone the HADOOP_CONF_DIR and then modify hive-site.xml. It's not hassle free.

vanzin:
> Users have to clone the HADOOP_CONF_DIR and then modify hive-site.xml. It's not hassle free.

It's pretty trivial unless they're doing it manually. Manual anything is, by definition, not hassle free.

> Do people put hive-site.xml in HADOOP_CONF_DIR?

They can put it wherever they want. They don't usually put it in SPARK_CONF_DIR either (at least not when using CDH, because we automatically set that up for them). Copying the file into SPARK_CONF_DIR is about as complicated as copying it into HADOOP_CONF_DIR.

vanzin:
BTW another trick is just to set YARN_CONF_DIR (or HADOOP_CONF_DIR) to point at HIVE_CONF_DIR. The latter generally contains both HDFS and YARN configs, since Hive needs those.

vanzin:
Let me write a longer version of why I think supporting hive-site.xml through spark.files is the wrong thing.

- the Spark documentation tells you to drop that file in SPARK_CONF_DIR. But that does not work in yarn-cluster mode; you may call it a bug in Spark and you'd be correct, but that's how the code that's out there behaves.

- uploading hive-site.xml as part of `--files` (or spark.files) is something that works *only* in yarn-cluster mode. For any other mode you'd have to write explicit code in Livy to make that file available on the Spark classpath somehow.

- Livy doesn't know about local files and won't even let you reference them in "spark.files" and friends. So this hive-site.xml would have to be uploaded somewhere first before it can be referenced in a Livy session. Add that to the previous point, and now Livy would have to download that file from somewhere before running Spark, adding even more complexity.

- On the other hand, dropping hive-site.xml in the Hadoop conf dir, or pointing HADOOP_CONF_DIR at the Hive configuration directory, has none of those problems. It works both in client and cluster mode, regardless of cluster manager. (For standalone and mesos you'd have to do the configuration on all nodes if you want cluster mode, but that's the same if you were placing hive-site.xml in SPARK_CONF_DIR.)

So my main point is that all the code being added to `InteractiveSession.scala` could just be removed, and it wouldn't make anyone's life more or less painful. Configuration is as easy or hard as it was before, depending on how you want to look at it, and it avoids a lot of code in Livy that, in my view, is not serving a real use case.

zjffdu:
> the Spark documentation tells you to drop that file in SPARK_CONF_DIR. But that does not work in yarn-cluster mode; you may call it a bug in Spark and you'd be correct, but that's how the code that's out there behaves.

This do looks like weird which make the yarn-client and yarn-cluster mode behave differently. We might need to fix it although it would bring in incompatibility. 

> uploading hive-site.xml as part of --files (or spark.files) is something that works only in yarn-cluster mode. For any other mode you'd have to write explicit code in Livy to make that file available on the Spark classpath somehow.
> Livy doesn't know about local files and won't even let you reference them in "spark.files" and friends. So this hive-site.xml would have to be uploaded somewhere first before it can be referenced in a Livy session. Add that to the previous point, and now Livy would have to download that file from somewhere before running Spark, adding even more complexity.

AFAIK, --files support hdfs file, so user can upload it to hdfs first and specify it in --files. So user don't need to upload any files to livy server. 

@vanzin dropping hive-site.xml in Hadoop conf dir would enforce this hive-site to be used globally, user would lose the ability to use custom hive-site.xml, e.g. user may want to connect to another hive metastore.

zjffdu:
retest this please

vanzin:
> So user don't need to upload any files to livy server.

I didn't say "the Livy server", I said "somewhere". Meaning, there's an extra manual step the user has to perform for that to work.

> dropping hive-site.xml in Hadoop conf dir would enforce this hive-site to be used globally

No. As I've said numerous times, any machine has this command called `cp` you can use to make a copy of something.

repl/src/main/scala/com/cloudera/livy/repl/SparkInterpreter.scala
```diff
@@ -84,12 +88,36 @@ class SparkInterpreter(conf: SparkConf) extends Interpreter {
       setContextClassLoaderMethod.invoke(sparkIMain)
 
       sparkContext = SparkContext.getOrCreate(conf)
-
+      if (conf.getBoolean("livy.repl.enableHiveContext", false)) {
```
vanzin: How do you set this?

SparkConf will only automatically load configs that start with `spark.`, and I don't see any code that's explicitly setting this config through `SparkConf.set`.

vanzin:
@tc0312 there's some weird interaction between this code and a recent change of yours. Here's the error from travis:

```
- should report an error if exception is thrown *** FAILED *** (4 seconds, 565 milliseconds)
  "	at <user code>$<user code>$<user code>$<user code>.func1(<console>:20)" did not fully match the regular expression \tat <user code>\.func1\(<console>:\d+\) (SparkSessionSpec.scala:140)
```

Unfortunately travis doesn't give access to the build logs, so that's all we have.

zjffdu:
@tc0312 
Here's output in master
`	at <user code>.func1(<console>:8)-`
and this is after applying this PR
`at <user code>$<user code>$<user code>$<user code>.func1(<console>:20)`

zjffdu:
@vanzin  I still feel some concern of copying `HADOOP_CONF_DIR` and put `hive-site.xml` there. Because it brings additional administration overhead for managing multiple  `HADOOP_CONF_DIR`. Besides that even we put hive-site under `HADOOP_CONF_DIR`, we still need to detect datanucleus jars even when user use remote hive metastore (this is a spark bug, fixed in spark 2.0). Since we encourage user to use yarn-cluster mode in livy, at least we should make HiveContext works in `yarn-cluster` mode. For the short term solution, I think we can detect the hive-site.xml and datanucleus jars in livy server as in my PR. For the long term solution, we can fix it in spark side, spark can should hive-site.xml and  datanucleus jars by itself in yarn-cluster mode and distribute them to driver & executors, then the behavior of yarn-client and yarn-cluster is consistent.  What do you think ?

vanzin:
> we still need to detect datanucleus jars even when user use remote hive metastore

Yeah, I keep forgetting about that because it's not a problem in CDH. :-/

As far as documentation goes, I still think it's better to recommend the HADOOP_CONF_DIR approach. That can easily be automated by admins: even if they don't have some management tool that does it for them, they just have to make a copy of the original HADOOP_CONF_DIR, and copy hive-site.xml there, in the startup code of Livy.

The other approaches are just too cumbersome; spark.files requires someone to upload the config to HDFS and every request to list the file. SPARK_CONF_DIR does not work in yarn-cluster mode.

We can make tweaks later if needed; right now I'm just worried that tests are not passing.

zjffdu:
@vanzin I have fixed test (this is due to LIVY-164).  And I just keep the changes in InteractiveSession if you feel OK, I think we still have chance to change that before the next release.

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
```diff
@@ -268,6 +296,59 @@ class InteractiveSession(
...
+                           livyConf: LivyConf): (Option[File], Boolean) = {
+    if (sparkFiles.exists(_.split("/").last == "hive-site.xml")) {
+      (None, true)
+    } else {
+      val hiveSiteURL = getClass.getResource("/hive-site.xml")
+      if (hiveSiteURL != null && hiveSiteURL.getProtocol == "file") {
+        return (Some(new File(hiveSiteURL.toURI)), false)
+      }
+      // only check SPARK_HOME/conf when SPARK_CONF_DIR is not defined
+      if (!sys.env.contains("SPARK_CONF_DIR")) {
```
vanzin: I think this is actually a bug in `bin/livy-server`. If `SPARK_CONF_DIR` is not defined it should be adding `$SPARK_HOME/conf` to the classpath in its place, otherwise its behavior is inconsistent.

Another possibility is to not add either to the classpath, although I'm not sure what other code that would affect.

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
```diff
@@ -70,36 +69,58 @@ class InteractiveSession(
...
+        }
       }
     }
 
+    def mergeHiveSiteAndHiveDeps(): Unit = {
+      val sparkFiles = conf.get("spark.files").map(_.split(",")).getOrElse(Array.empty[String])
+      if (livyConf.getBoolean(LivyConf.ENABLE_HIVE_CONTEXT)) {
+        hiveSiteFile(sparkFiles, livyConf) match {
+          case (_, true) =>
+            info("Enable HiveContext because hive-site.xml is found in user request.")
```
vanzin: These should be `debug`; no need to print this out for every session that is created.

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
```diff
@@ -70,36 +69,58 @@ class InteractiveSession(
...
+            mergeConfList(datanucleusJars(livyConf), LivyConf.SPARK_JARS)
+          case (Some(file), false) =>
+            info("Enable HiveContext because hive-site.xml is found under classpath, "
+              + file.getAbsolutePath)
+            mergeConfList(List(file.getAbsolutePath), LivyConf.SPARK_FILES)
+            mergeConfList(datanucleusJars(livyConf), LivyConf.SPARK_JARS)
+          case (None, false) =>
+            warn("Enable HiveContext but no hive-site.xml found under" +
```
vanzin: I don't think this warrants a warning. If you're gonna allow hdfs-site in the user request, it's perfectly normal for an admin to not provide their own default, and require users to provide their own if they want to use HiveContext. The code in the interpreter should detect that and not start a HiveContext.

So, like the others, `debug`.

server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
```diff
@@ -48,9 +48,11 @@ class LivyServer extends Logging {
 
   private var server: WebServer = _
   private var _serverUrl: Option[String] = None
+  // make livyConf accessible for testing
+  var livyConf: LivyConf = _
```
vanzin: This probably could be done more cleanly, but at the very least, `private[livy]`.

repl/src/main/scala/com/cloudera/livy/repl/SparkInterpreter.scala
```diff
@@ -118,12 +121,36 @@ class SparkInterpreter(conf: SparkConf) extends Interpreter with Logging {
       }
 
       sparkContext = SparkContext.getOrCreate(conf)
-
+      if (conf.getBoolean("spark.repl.enableHiveContext", false)) {
+        try {
+          val loader = Option(Thread.currentThread().getContextClassLoader)
+            .getOrElse(getClass.getClassLoader)
+          if (loader.getResource("hive-site.xml") == null) {
+            warn("livy.repl.enableHiveContext is true but no hive-site.xml found on classpath.")
```
vanzin: Similarly, not worth of a warning given how the functionality is set up.

vanzin:
@zjffdu pointed out a couple of minor issues (the SPARK_CONF_DIR one being the more serious one), but otherwise looks ok; there's some possible cleanup (e.g. I find `mergeHiveSiteAndHiveDeps` confusing) but that can be done later.

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
zjffdu: Agree, should add `$SPARK_HOME/conf` to the classpath to make it consistent with spark.  Created LIVY-182 for this bug.

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
zjffdu: But I will check hive-site.xml both in classpath and user request. I think it deserves a warning when both places can not find hive-site.xml

repl/src/main/scala/com/cloudera/livy/repl/SparkInterpreter.scala
zjffdu: This warning is for the case that user want HiveContext but hive-site.xml is not available somehow, such as misconfiguration. I think at least in the interpreter side, it should print some log to indicate whether he use SQLContext or HiveContext.

vanzin:
@zjffdu could you fix the merge conflicts? Then I can merge the PR.

vanzin:
Merging.

zjffdu:
@vanzin  Thanks for patient review. I forgot to update PR summary and just update that. Would you help to update that commit. Thanks

PR #138, opened 2016-05-17
LIVY-98. Run kinit before launching spark session in secure mode

zjffdu:
The solution here is always execute kinit command before launching spark session for both interactive & batch. The weird thing of this PR is that I would always translate the related conf from LivyConf to RSCConf. The easy way I think is just make RSCConf extends LivyConf, but sure whether it is proper in semantic.


vanzin:
Hi @zjffdu,

Thanks for working on this but I don't believe this is the right approach. It has a few problems that I can see:

- too much code duplication; you could reduce that by having helper code to run kinit, though, instead of having the code copy & pasted in both places.

- too many login attempts; every session will try to log in to the kdc. If you have a burst of sessions being launched, that will end up causing failures.

- Livy itself needs a kerberos login. To be able to upload files to HDFS when launching user sessions, Livy itself needs a TGT so that it can talk to HDFS. Later, when we add support for reading information from YARN too, Livy will need the TGT for that too. So to properly solve this problem you need an approach that also handles the login for the Livy server itself.

I did some investigation about this in the past but didn't have the time to sit down and write the code. I think the proper approach here would be a combination of `LivyServer` running some code periodically to make sure the kerberos login is valid; that means running `kinit` if the TGT is expired, and `kinit -R` while the TGT is renewable.

That periodic task also should do two other things:
- call `UserGroupInformation.reloginFromTicketCache` when a new ticket cache is generated.
- make sure the `kinit` command generates a ticket cache and that the ticket cache is picked up by child processes.

That way there's nothing that needs to be done when `spark-submit` is executed, since it will pick up the ticket cache that Livy is keeping up-to-date.

zjffdu:
Thanks @vanzin , I miss the part that livy itself also needs a kerberos login for uploading files to hdfs. And running kinit thread is also I what I think. Will do that.

zjffdu:
@vanzin  I just add a thread to run kinit periodically. I don't know the api to check whether TGT is expired and renewable.  But seems my solution works as long as the the ticket refresh internal is less than ticket_lifetime in /etc/krb5.conf

zjffdu:
Close it PR, and open #145

PR #144, opened 2016-06-03
LIVY-154. Don't reuse connections in client-http.

vanzin:
Either the current client code is doing something wrong when setting
up the http client, or there's a bug in the http client library, but
the result is that a connection ends up being closed while still in
use, causing errors when the client is talking to Livy.

While investigating that I found a couple of other issues that I also
fixed here:

- the client-http packaging was relocating too much stuff, which broke
  logging for the relocated classes, making it hard to debug things.

- the server wasn't configuring multipart request handling for servlets;
  fixing that means that we can't use asynchronous tasks, since Jetty
  (at least the version we use) does not allow that. Async tasks weren't
  really needed anyway, since they're either pretty quick or, otherwise,
  would block a thread in another shared thread pool anyway, so there
  wasn't really any gain.

### DIFF
--- a/client-http/pom.xml
+++ b/client-http/pom.xml
@@ -146,8 +146,12 @@
                   <shadedPattern>com.cloudera.livy.shaded.jackson</shadedPattern>
                 </relocation>
                 <relocation>
-                  <pattern>org.apache</pattern>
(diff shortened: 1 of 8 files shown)


vanzin:
I tested this on top of #132 and tests now pass for me.

PR #145, opened 2016-06-06
LIVY-98. Run kinit thread periodically in secure mode

zjffdu:
The solution here is always execute kinit command periodically as long as the refresh interval is less than the ticket lifetime which is specified in /etc/krb5.conf

### DIFF
--- a/rsc/src/main/java/com/cloudera/livy/rsc/ContextLauncher.java
+++ b/rsc/src/main/java/com/cloudera/livy/rsc/ContextLauncher.java
@@ -28,13 +28,7 @@ import java.io.OutputStreamWriter;
 import java.io.Reader;
 import java.io.Writer;
 import java.nio.file.Files;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.EnumSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Properties;
-import java.util.UUID;
+import java.util.*;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
@@ -129,9 +123,13 @@ class ContextLauncher {
     factory.getServer().unregisterClient(clientId);
     try {
       if (forceKill) {
-        child.kill();
+        if (child != null) {
+          child.kill();
+        }
       } else {
-        child.detach();
+        if (child != null) {
+          child.detach();
+        }
       }
     } finally {
       factory.unref();
@@ -201,7 +199,12 @@ class ContextLauncher {
       };
       return new ChildProcess(conf, promise, child, confFile);
     } else {
-      final SparkLauncher launcher = new SparkLauncher();
+      Map<String, String> env = new HashMap<String, String>();
+      // it is no harm to set env KRB5CCNAME in non-secured cluster.
+      if (conf.get(RSCConf.Entry.KINIT_CACHE_FILE) != null) {
+        env.put("KRB5CCNAME", conf.get(RSCConf.Entry.KINIT_CACHE_FILE));
+      }
+      final SparkLauncher launcher = new SparkLauncher(env);
 
       // Spark 1.x does not support specifying deploy mode in conf and needs special handling.
       String deployMode = conf.get(SPARK_DEPLOY_MODE);
--- a/rsc/src/main/java/com/cloudera/livy/rsc/ContextLauncher.java
+++ b/rsc/src/main/java/com/cloudera/livy/rsc/ContextLauncher.java
@@ -28,7 +28,13 @@ import java.io.OutputStreamWriter;
 import java.io.Reader;
 import java.io.Writer;
 import java.nio.file.Files;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.EnumSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Properties;
+import java.util.UUID;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
@@ -123,13 +129,9 @@ class ContextLauncher {
     factory.getServer().unregisterClient(clientId);
     try {
       if (forceKill) {
-        if (child != null) {
-          child.kill();
-        }
+        child.kill();
       } else {
-        if (child != null) {
-          child.detach();
-        }
+        child.detach();
       }
     } finally {
       factory.unref();
@@ -199,12 +201,7 @@ class ContextLauncher {
       };
       return new ChildProcess(conf, promise, child, confFile);
     } else {
-      Map<String, String> env = new HashMap<String, String>();
-      // it is no harm to set env KRB5CCNAME in non-secured cluster.
-      if (conf.get(RSCConf.Entry.KINIT_CACHE_FILE) != null) {
-        env.put("KRB5CCNAME", conf.get(RSCConf.Entry.KINIT_CACHE_FILE));
-      }
-      final SparkLauncher launcher = new SparkLauncher(env);
+      final SparkLauncher launcher = new SparkLauncher();
 
       // Spark 1.x does not support specifying deploy mode in conf and needs special handling.
       String deployMode = conf.get(SPARK_DEPLOY_MODE);
--- a/server/src/main/scala/com/cloudera/livy/LivyConf.scala
+++ b/server/src/main/scala/com/cloudera/livy/LivyConf.scala
@@ -66,7 +66,7 @@ object LivyConf {
   val LAUNCH_KERBEROS_KEYTAB =
     LivyConf.Entry("livy.server.launch.kerberos.keytab", null)
   val LAUNCH_KERBEROS_REFRESH_INTERVAL =
-    LivyConf.Entry("livy.server.launch.kerberos.refresh_interval_seconds", 3600)
+    LivyConf.Entry("livy.server.launch.kerberos.refresh_interval", "1h")
   val KINIT_FAIL_THRESHOLD =
     LivyConf.Entry("livy.server.launch.kerberos.kinit_fail_threshold", 5)
 
--- a/server/src/main/scala/com/cloudera/livy/LivyConf.scala
+++ b/server/src/main/scala/com/cloudera/livy/LivyConf.scala
@@ -20,6 +20,7 @@ package com.cloudera.livy
 
 import java.io.File
 import java.lang.{Boolean => JBoolean, Long => JLong}
+import java.nio.file.Files
 
 import org.apache.hadoop.conf.Configuration
 
@@ -57,9 +58,22 @@ object LivyConf {
   val SUPERUSERS = Entry("livy.superusers", null)
 
   val AUTH_TYPE = Entry("livy.server.auth.type", null)
-  val KERBEROS_PRINCIPAL = Entry("livy.server.auth.kerberos.principal", null)
-  val KERBEROS_KEYTAB = Entry("livy.server.auth.kerberos.keytab", null)
-  val KERBEROS_NAME_RULES = Entry("livy.server.auth.kerberos.name_rules", "DEFAULT")
+  val AUTH_KERBEROS_PRINCIPAL = Entry("livy.server.auth.kerberos.principal", null)
+  val AUTH_KERBEROS_KEYTAB = Entry("livy.server.auth.kerberos.keytab", null)
+  val AUTH_KERBEROS_NAME_RULES = Entry("livy.server.auth.kerberos.name_rules", "DEFAULT")
+
+  val LAUNCH_KERBEROS_PRINCIPAL =
+    LivyConf.Entry("livy.server.launch.kerberos.principal", null)
+  val LAUNCH_KERBEROS_KEYTAB =
+    LivyConf.Entry("livy.server.launch.kerberos.keytab", null)
+  val LAUNCH_KERBEROS_REFRESH_INTERVAL =
+    LivyConf.Entry("livy.server.launch.kerberos.refresh_interval_seconds", 3600)
+  val KINIT_FAIL_THRESHOLD =
+    LivyConf.Entry("livy.server.launch.kerberos.kinit_fail_threshold", 5)
+  val KINIT_CACHE_ROOT_DIR =
+    LivyConf.Entry("livy.server.launch.kerberos.kinit_cache_root_dir",
+      Files.createTempDirectory("kinit_cache").toFile.getAbsolutePath)
+
 
   val SPARK_MASTER = "spark.master"
   val SPARK_DEPLOY_MODE = "spark.submit.deployMode"
--- a/server/src/main/scala/com/cloudera/livy/LivyConf.scala
+++ b/server/src/main/scala/com/cloudera/livy/LivyConf.scala
@@ -20,7 +20,6 @@ package com.cloudera.livy
 
 import java.io.File
 import java.lang.{Boolean => JBoolean, Long => JLong}
-import java.nio.file.Files
 
 import org.apache.hadoop.conf.Configuration
 
@@ -70,10 +69,6 @@ object LivyConf {
     LivyConf.Entry("livy.server.launch.kerberos.refresh_interval_seconds", 3600)
   val KINIT_FAIL_THRESHOLD =
     LivyConf.Entry("livy.server.launch.kerberos.kinit_fail_threshold", 5)
-  val KINIT_CACHE_ROOT_DIR =
-    LivyConf.Entry("livy.server.launch.kerberos.kinit_cache_root_dir",
-      Files.createTempDirectory("kinit_cache").toFile.getAbsolutePath)
-
 
   val SPARK_MASTER = "spark.master"
   val SPARK_DEPLOY_MODE = "spark.submit.deployMode"
--- a/server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
+++ b/server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
@@ -21,12 +21,9 @@ package com.cloudera.livy.server
 import java.io.{File, IOException}
 import java.util.concurrent._
 import java.util.EnumSet
-import java.util.UUID
 import javax.servlet._
 
-import org.apache.commons.io.IOUtils
-import org.apache.commons.lang.StringUtils
-import org.apache.hadoop.security.{SecurityUtil, UserGroupInformation}
+import org.apache.hadoop.security.SecurityUtil
 import org.apache.hadoop.security.authentication.server._
 import org.eclipse.jetty.servlet.FilterHolder
 import org.scalatra.metrics.MetricsBootstrap
@@ -34,7 +31,6 @@ import org.scalatra.metrics.MetricsSupportExtensions._
 import org.scalatra.servlet.{MultipartConfig, ServletApiImplicits}
 
 import com.cloudera.livy._
-import com.cloudera.livy.rsc.RSCConf
 import com.cloudera.livy.server.batch.BatchSessionServlet
 import com.cloudera.livy.server.interactive.InteractiveSessionServlet
 import com.cloudera.livy.util.LineBufferedProcess
@@ -138,9 +134,9 @@ class LivyServer extends Logging {
         if (!runKinit(launch_keytab, launch_principal)) {
           error("Failed to run kinit, stopping the server.")
           sys.exit(1)
-        } else {
-          startKinitDaemon(launch_keytab, launch_principal)
         }
+        startKinitThread(launch_keytab, launch_principal)
+
       case null =>
         // Nothing to do.
 
@@ -174,7 +170,6 @@ class LivyServer extends Logging {
       case 0 =>
         debug("Ran kinit command successfully.")
         kinitFailCount = 0
-        UserGroupInformation.getCurrentUser.reloginFromTicketCache()
         true
       case _ =>
         warn("Fail to run kinit command.")
@@ -183,21 +178,17 @@ class LivyServer extends Logging {
     }
   }
 
-  def startKinitDaemon(keytab: String, principal: String): Unit = {
+  def startKinitThread(keytab: String, principal: String): Unit = {
+    val refreshInterval = livyConf.getTimeAsMs(LAUNCH_KERBEROS_REFRESH_INTERVAL)
+    val kinitFailThreshold = livyConf.getInt(KINIT_FAIL_THRESHOLD)
     executor.schedule(
       new Runnable() {
-        def run(): Unit = {
+        override def run(): Unit = {
           if (runKinit(keytab, principal)) {
             // schedule another kinit run with a fixed delay.
-            executor.schedule(
-              new Runnable() {
-                def run(): Unit = {
-                  runKinit(keytab, principal)
-                }
-              }, livyConf.getInt(LAUNCH_KERBEROS_REFRESH_INTERVAL), TimeUnit.SECONDS)
+            executor.schedule(this, refreshInterval, TimeUnit.MILLISECONDS)
           } else {
             // schedule another retry at once or fail the livy server if too many times kinit fail
-            val kinitFailThreshold = livyConf.getInt(KINIT_FAIL_THRESHOLD)
             if (kinitFailCount >= kinitFailThreshold) {
               error(s"Exit LivyServer after ${kinitFailThreshold} times failures running kinit.")
               if (server.server.getState == "STARTED") {
@@ -206,13 +197,11 @@ class LivyServer extends Logging {
                 sys.exit(1)
               }
             } else {
-              executor.submit(new Runnable() {
-                def run(): Unit = runKinit(keytab, principal)
-              })
+              executor.submit(this)
             }
           }
         }
-      }, livyConf.getInt(LAUNCH_KERBEROS_REFRESH_INTERVAL), TimeUnit.SECONDS)
+      }, refreshInterval, TimeUnit.MILLISECONDS)
   }
 
   def join(): Unit = server.join()
--- a/server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
+++ b/server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
@@ -49,7 +49,6 @@ class LivyServer extends Logging {
   private[livy] var livyConf: LivyConf = _
 
   private var kinitFailCount: Int = 0
-  private var kinitCacheDir: String = _
   private var executor: ScheduledExecutorService = _
 
   def start(): Unit = {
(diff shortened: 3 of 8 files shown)


server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
```diff
@@ -85,27 +97,31 @@ class LivyServer extends Logging {
...
+          livyConf.get(AUTH_KERBEROS_NAME_RULES))
         server.context.addFilter(holder, "/*", EnumSet.allOf(classOf[DispatcherType]))
         info(s"SPNEGO auth enabled (principal = $principal)")
         if (!livyConf.getBoolean(LivyConf.IMPERSONATION_ENABLED)) {
           info(s"Enabling impersonation since auth type is $authType.")
           livyConf.set(LivyConf.IMPERSONATION_ENABLED, true)
         }
 
+        // run kinit periodically
+        val executor = new ScheduledThreadPoolExecutor(1)
```
vanzin: It's safer to add an explicit call to `runKinit` here instead of scheduling at task with delay `0`.

server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
```diff
@@ -85,27 +97,31 @@ class LivyServer extends Logging {
...
+        executor.scheduleAtFixedRate(runKinit(livyConf),
+          0, livyConf.getInt(LAUNCH_KERBEROS_REFRESH_INTERVAL), TimeUnit.SECONDS)
```
vanzin: nit: add empty line after this one

server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
```diff
@@ -126,6 +142,29 @@ class LivyServer extends Logging {
     sys.props("livy.server.serverUrl") = _serverUrl.get
   }
 
+  def runKinit(livyConf: LivyConf): Runnable = new Runnable {
+    override def run(): Unit = {
+      UserGroupInformation.getCurrentUser.reloginFromTicketCache()
```
vanzin: This should be done after kinit is run successfully, right?

server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
```diff
@@ -35,16 +39,24 @@ import com.cloudera.livy.server.batch.BatchSessionServlet
 import com.cloudera.livy.server.interactive.InteractiveSessionServlet
 import com.cloudera.livy.util.LineBufferedProcess
 
+
```
vanzin: nit: remove

server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
```diff
@@ -126,6 +142,29 @@ class LivyServer extends Logging {
...
+  def runKinit(livyConf: LivyConf): Runnable = new Runnable {
+    override def run(): Unit = {
+      UserGroupInformation.getCurrentUser.reloginFromTicketCache()
+      val keytab = livyConf.get(LAUNCH_KERBEROS_KEYTAB)
+      val principal = livyConf.get(LAUNCH_KERBEROS_PRINCIPAL)
+      require(principal != null,
+        s"Kerberos requires ${LAUNCH_KERBEROS_KEYTAB.key} to be provided.")
+      require(keytab != null,
+        s"Kerberos requires ${LAUNCH_KERBEROS_PRINCIPAL.key} to be provided.")
+      val commands = ArrayBuffer.empty[String]
```
vanzin: Don't use mutable lists when not strictly necessary; for example:

    val args = Seq("kinit", "-kt", ...)

server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
```diff
@@ -126,6 +142,29 @@ class LivyServer extends Logging {
...
+      commands += "kinit"
+      commands += "-kt"
+      commands += keytab
+      commands += principal
+      val proc = new ProcessBuilder(commands: _*).start()
+      proc.waitFor() match {
+        case 0 => info("Run kinit command")
+        case _ => warn("Fail to run kinit command," +
```
vanzin: This looks like a very serious error that should be fatal; maybe allow a certain number of failures, but shut down the server after that. The first login should definitely be required (server shouldn't even start if kinit fails).

server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
```diff
@@ -126,6 +142,29 @@ class LivyServer extends Logging {
...
+      val principal = livyConf.get(LAUNCH_KERBEROS_PRINCIPAL)
+      require(principal != null,
+        s"Kerberos requires ${LAUNCH_KERBEROS_KEYTAB.key} to be provided.")
+      require(keytab != null,
+        s"Kerberos requires ${LAUNCH_KERBEROS_PRINCIPAL.key} to be provided.")
+      val commands = ArrayBuffer.empty[String]
+      commands += "kinit"
+      commands += "-kt"
+      commands += keytab
+      commands += principal
```
vanzin: Shouldn't you also pass `-l` so that you at least request a ticket with a lifetime long enough for the re-login period set up in the configuration? (I think it's `-l`, don't have the man page handy.)

zjffdu: I am not sure whether we should introduce this option, as I feel it would make it more complicated to understand. '-l' can specify lifetime but it is constrained by max_life of KDC (if it is larger than max_life, then it won't take effect). For now, as long as user make the kinit refresh interval larger than ticket_lifetime, then the ticket will never expired

server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
```diff
@@ -19,12 +19,16 @@
 package com.cloudera.livy.server
 
 import java.io.{File, IOException}
-import java.net.InetAddress
 import java.util.EnumSet
+import java.util.concurrent.{ScheduledThreadPoolExecutor, TimeUnit}
 import javax.servlet._
 
+import scala.collection.mutable.ArrayBuffer
```
vanzin: Not needed anymore.

server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
vanzin: This should be a class field and properly stopped in the `stop()` method.

server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
```diff
@@ -135,6 +161,32 @@ class LivyServer extends Logging {
...
+    val principal = livyConf.get(LAUNCH_KERBEROS_PRINCIPAL)
+    require(principal != null,
+      s"Kerberos requires ${LAUNCH_KERBEROS_KEYTAB.key} to be provided.")
+    require(keytab != null,
+      s"Kerberos requires ${LAUNCH_KERBEROS_PRINCIPAL.key} to be provided.")
+    val commands = Seq("kinit", "-kt", keytab, principal)
+    val proc = new ProcessBuilder(commands: _*).start()
+    proc.waitFor() match {
+      case 0 =>
+        info("Run kinit command successfully")
```
vanzin: s/Run/Ran

I also think this should be less than info... it doesn't really provide a lot of interesting info and will just add noise to the logs.

server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
```diff
@@ -135,6 +161,32 @@ class LivyServer extends Logging {
...
 
+  def runKinit(livyConf: LivyConf): Unit = {
+    val keytab = livyConf.get(LAUNCH_KERBEROS_KEYTAB)
+    val principal = livyConf.get(LAUNCH_KERBEROS_PRINCIPAL)
+    require(principal != null,
+      s"Kerberos requires ${LAUNCH_KERBEROS_KEYTAB.key} to be provided.")
+    require(keytab != null,
+      s"Kerberos requires ${LAUNCH_KERBEROS_PRINCIPAL.key} to be provided.")
+    val commands = Seq("kinit", "-kt", keytab, principal)
+    val proc = new ProcessBuilder(commands: _*).start()
```
vanzin: I thought about this yesterday and I'm slightly worried about this (running a child process without reading its output). kinit is not really noisy and its output shouldn't really fill the buffers, but you never know... it would be safer to just redirect the output (`ProcessBuilder.redirectOutput(ProcessBuilder.Redirect.INHERIT)`). 

You could do the same for the error stream but then you'd lose the ability to read the error and put it in the exception message. Which shouldn't be awful since the error will still be written to Livy's logs.

But at least stdout should be redirected.

server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
```diff
@@ -135,6 +161,32 @@ class LivyServer extends Logging {
...
+        info("Run kinit command successfully")
+        kinitFailCount = 0
+        UserGroupInformation.getCurrentUser.reloginFromTicketCache()
+      case _ =>
+        warn("Fail to run kinit command," +
+          StringUtils.join(IOUtils.readLines(proc.getErrorStream), "\n"))
+        kinitFailCount += 1
+        val kinitFailThreshold = livyConf.getInt(KINIT_FAIL_THRESHOLD)
+        if (kinitFailCount >= kinitFailThreshold) {
+          error("Exit LivyServer after ${kinitFailThreshold} times of running kinit.")
```
vanzin: "after x failures running kinit"

server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
```diff
@@ -94,27 +109,38 @@ class LivyServer extends Logging {
...
         server.context.addFilter(holder, "/*", EnumSet.allOf(classOf[DispatcherType]))
         info(s"SPNEGO auth enabled (principal = $principal)")
         if (!livyConf.getBoolean(LivyConf.IMPERSONATION_ENABLED)) {
           info(s"Enabling impersonation since auth type is $authType.")
           livyConf.set(LivyConf.IMPERSONATION_ENABLED, true)
         }
 
+        // run kinit periodically
+        val executor = new ScheduledThreadPoolExecutor(1)
+        runKinit(livyConf)
```
vanzin: AFAICT this call isn't failing when kinit fails; the server shouldn't come up if this fails. Maybe return a boolean from `runKinit`?

vanzin:
Just a few small things left to look at. A minor thing is that perhaps we should write the Livy-generated ticket cache to a custom location instead of relying on the default '/tmp'; that would require also propagating an environment variable to child spark-submit processes. That would allow, for example, multiple Livy servers on the same host running as different kerberos principals under the local user. Not a big deal, but also shouldn't be hard to do.

vanzin:
Also could you change the PR summary to explain the change instead of just referring to the other PR?

zjffdu:
@vanzin I didn't find the api for writing the ticket cache to a custom location, do you have know that? Thanks

vanzin:
From kinit's man page:

```
       -c cache_name
              use cache_name as the Kerberos 5 credentials (ticket) cache name and location; if this option is not  used,  the  default
              cache name and location are used.

              The default credentials cache may vary between systems.  If the KRB5CCNAME environment variable is set, its value is used
              to name the default ticket cache.  If a principal name is specified and the type of the default  credentials  cache  sup-
              ports  a  collection  (such as the DIR type), an existing cache containing credentials for the principal is selected or a
              new one is created and becomes the new primary cache.   Otherwise,  any  existing  contents  of  the  default  cache  are
              destroyed by kinit.
```

server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
```diff
@@ -35,19 +37,32 @@ import com.cloudera.livy.server.batch.BatchSessionServlet
...
+  private val LAUNCH_KERBEROS_PRINCIPAL =
+    LivyConf.Entry("livy.server.launch.kerberos.principal", null)
+  private val LAUNCH_KERBEROS_KEYTAB =
+    LivyConf.Entry("livy.server.launch.kerberos.keytab", null)
+  private val LAUNCH_KERBEROS_REFRESH_INTERVAL =
+    LivyConf.Entry("livy.server.launch.kerberos.refresh_interval_seconds", 3600)
+  private val KINIT_FAIL_THRESHOLD =
+      LivyConf.Entry("livy.server.launch.kerberos.kinit_fail_threshold", 5)
+  private val KINIT_CACHE_NAME =
+    LivyConf.Entry("livy.server.launch.kerberos.kinit_cache_name", "/tmp")
```
vanzin: This cannot be a directory. If it's a directory, `kinit` succeeds but doesn't write a ticket cache file, which is not what you want here.

Also, it would be better if instead of this, Livy created a temp directory to store temporary data. The location of the temp directory could be configured, and by default it can be just `java.io.tmpdir`. Then create a unique directory there and write temp data into it. Then you don't need a config entry for this.

Another thing that you're missing is exposing this to the spark-submit processes started by Livy. As the man page says, you need to set `KRB5CCNAME` in the process's environment.

As I mentioned in my original comment, it would be nice to do this, but you don't need to do it right now if you don't want to. But the code you added at the moment just won't work.

server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
```diff
@@ -135,6 +161,40 @@ class LivyServer extends Logging {
...
+        if (kinitFailCount >= kinitFailThreshold) {
+          error(s"Exit LivyServer after ${kinitFailThreshold} times failures running kinit.")
+          if (server.server.getState == "STARTED") {
+            stop()
+          } else {
+            sys.exit(1)
+          }
+        } else {
+          // retry after 5 seconds
+          Thread.sleep(5*1000)
```
vanzin: This is a little weird. I'd just schedule a one-off task to run `runKinit` using the existing executor. Which required addressing my previous feedback of making `executor` a class field.

server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
vanzin: This still needs to be addressed. If the first login fails, the server shouldn't come up, because more likely than not there's an error in the configuration.

server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
```diff
@@ -19,40 +19,56 @@
...
+  private val LAUNCH_KERBEROS_PRINCIPAL =
+    LivyConf.Entry("livy.server.launch.kerberos.principal", null)
+  private val LAUNCH_KERBEROS_KEYTAB =
+    LivyConf.Entry("livy.server.launch.kerberos.keytab", null)
+  private val LAUNCH_KERBEROS_REFRESH_INTERVAL =
+    LivyConf.Entry("livy.server.launch.kerberos.refresh_interval_seconds", 3600)
+  private val KINIT_FAIL_THRESHOLD =
+      LivyConf.Entry("livy.server.launch.kerberos.kinit_fail_threshold", 5)
+  private val KINIT_CACHE_ROOT_DIR =
+    LivyConf.Entry("livy.server.launch.kerberos.kinit_cache_root_dir", System.getenv("java.io.tmp"))
```
vanzin: There's no env variable with that name. You might be thinking of the system property named "java.io.tmpdir". Instead of this, I'd rather use `java.nio.file.Files.createTempDirectory` if this is not explicitly set. See `RSCDriver.java` for an example.

vanzin: In fact, `java.nio.file.Files.createTempDirectory` might be a good idea even if this is set, replacing the random UUID you're using later.

vanzin:
@zjffdu there's feedback I sent a long time ago that you still haven't addressed. For example:

- the `executor` variable needs to be a class instance, and shut down when the server stops. It should also use daemon threads.
- the server *must fail and exit* if the first login attempt fails.

rsc/src/main/java/com/cloudera/livy/rsc/ContextLauncher.java
```diff
@@ -198,12 +196,16 @@ public void run() {
       };
       return new ChildProcess(conf, promise, child, confFile);
     } else {
+      Map<String, String> env = new HashMap<String, String>();
+      // it is no harm to set env KRB5CCNAME in non-secured cluster.
+      if (System.getProperty(RSCConf.Entry.KINIT_CACHE_FILE.key()) != null) {
```
vanzin: It's pretty weird that you're looking at a system property and not in `conf`.

vanzin:
You also need to do the cache file env variable for batches (the code you're changing only affects interactive sessions).

zjffdu:
@vanzin  Sorry to mention that it is not ready to review, I just want to utilize the travis to build it.  Thanks for the quick review, I will update the PR soon.

ksakellis:
@zjffdu are you still working on this patch?

zjffdu:
Sorry, I will update the PR soon.

zjffdu:
ping @vanzin

server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
```diff
@@ -88,27 +97,46 @@ class LivyServer extends Logging {
...
+            override def newThread(r: Runnable): Thread = {
+              val thread = new Thread(r)
+              thread.setName("kinit-thread")
+              thread.setDaemon(true)
+              thread
+            }
+          }
+        )
+        if (!runKinit()) {
+          error("fail to run kinit, stop the server")
```
vanzin: "Failed to run kinit, stopping the server."

server/src/main/scala/com/cloudera/livy/LivyConf.scala
```diff
@@ -57,9 +58,22 @@ object LivyConf {
...
+    LivyConf.Entry("livy.server.launch.kerberos.principal", null)
+  val LAUNCH_KERBEROS_KEYTAB =
+    LivyConf.Entry("livy.server.launch.kerberos.keytab", null)
+  val LAUNCH_KERBEROS_REFRESH_INTERVAL =
+    LivyConf.Entry("livy.server.launch.kerberos.refresh_interval_seconds", 3600)
+  val KINIT_FAIL_THRESHOLD =
+    LivyConf.Entry("livy.server.launch.kerberos.kinit_fail_threshold", 5)
+  val KINIT_CACHE_ROOT_DIR =
+    LivyConf.Entry("livy.server.launch.kerberos.kinit_cache_root_dir",
+      Files.createTempDirectory("kinit_cache").toFile.getAbsolutePath)
```
vanzin: This will always create a temp directory, regardless of whether the user has set this or not. Better to do that only if the user has not set this. Also, "root" in the config name sounds a little confusing, it kinda implies there might be other subdirectories involved; I'd just remove it.

server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
```diff
@@ -135,6 +163,64 @@ class LivyServer extends Logging {
...
+      s"Kerberos requires ${LAUNCH_KERBEROS_PRINCIPAL.key} to be provided.")
+    require(keytab != null,
+      s"Kerberos requires ${LAUNCH_KERBEROS_KEYTAB.key} to be provided.")
+    val cacheFileName = kinitCacheDir + "/krb5cc_0"
+    val commands = Seq("kinit", "-kt", keytab, "-c", cacheFileName, principal)
+    val proc = new ProcessBuilder(commands: _*).inheritIO().start()
+    proc.waitFor() match {
+      case 0 =>
+        debug("Ran kinit command successfully.")
+        livyConf.set(RSCConf.Entry.KINIT_CACHE_FILE, cacheFileName)
```
vanzin: Do you need to set this every time `runKinit` is called? Seems like you could do this initialization once and be done with it.

server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
```diff
@@ -135,6 +163,64 @@ class LivyServer extends Logging {
...
+        kinitFailCount = 0
+        UserGroupInformation.getCurrentUser.reloginFromTicketCache()
```
vanzin: How does `reloginFromTicketCache` know to read the ticket cache being written to the custom location? The only way I can see to override that in `UserGroupInformation.java` is to use the environment variable, and you can't change the environment at this point...

server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
```diff
@@ -135,6 +163,64 @@ class LivyServer extends Logging {
...
+        true
+      case _ =>
+        warn("Fail to run kinit command," +
+          StringUtils.join(IOUtils.readLines(proc.getErrorStream), "\n"))
```
vanzin: Does `readLines` do anything since you're using `inheritIO` when running the child process?

server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
```diff
@@ -135,6 +163,64 @@ class LivyServer extends Logging {
...
+    }
+  }
+
+  def startKinitDaemon(): Unit = {
+    executor.scheduleWithFixedDelay(
+      new Runnable() {
+        def run(): Unit = {
+          if (runKinit()) {
+            // schedule another kinit run with a fixed delay.
+            executor.scheduleWithFixedDelay(
```
vanzin: Why are you scheduling a periodic task from within a periodic task? Won't that cause a whole bunch of tasks to execute as time passes?

server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
```diff
@@ -135,6 +163,64 @@ class LivyServer extends Logging {
...
+      case _ =>
+        warn("Fail to run kinit command," +
+          StringUtils.join(IOUtils.readLines(proc.getErrorStream), "\n"))
+        kinitFailCount += 1
+        false
+    }
+  }
+
+  def startKinitDaemon(): Unit = {
+    executor.scheduleWithFixedDelay(
```
vanzin: I think you want slightly different logic here. Basically, you want to schedule the task once (`executor.schedule` instead of `executor.scheduleWithFixedDelay`), and then:

- if kinit succeeds, schedule the same task again for the future
- if kinit fails, handle the error and either schedule the task "soon" or call `runKinit()` in a loop

So you'd never schedule recurring tasks.

vanzin:
@zjffdu the question about `reloginFromTicketCache` vs. the custom cache location is the one that really worries me at the moment. As I said before, that's a nice to have - if you don't want to handle the custom location in this patch, no problem. But if you do, that needs to be solved, because I don't think this code is really working in the current state.

server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
zjffdu: Yeah, my bad, I was intending to call `executor.schedule`

zjffdu:
@vanzin I push another commit to remove the cache related stuff, please help review, thanks.

server/src/main/scala/com/cloudera/livy/LivyConf.scala
```diff
@@ -57,9 +57,18 @@ object LivyConf {
...
+  val AUTH_KERBEROS_PRINCIPAL = Entry("livy.server.auth.kerberos.principal", null)
+  val AUTH_KERBEROS_KEYTAB = Entry("livy.server.auth.kerberos.keytab", null)
+  val AUTH_KERBEROS_NAME_RULES = Entry("livy.server.auth.kerberos.name_rules", "DEFAULT")
+
+  val LAUNCH_KERBEROS_PRINCIPAL =
+    LivyConf.Entry("livy.server.launch.kerberos.principal", null)
+  val LAUNCH_KERBEROS_KEYTAB =
+    LivyConf.Entry("livy.server.launch.kerberos.keytab", null)
+  val LAUNCH_KERBEROS_REFRESH_INTERVAL =
+    LivyConf.Entry("livy.server.launch.kerberos.refresh_interval_seconds", 3600)
```
vanzin: minor: can you use "1h" as the default, and read the config using `getTimeAsMs`? That's more user-friendly.

server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
```diff
@@ -135,6 +167,54 @@ class LivyServer extends Logging {
...
+        warn("Fail to run kinit command.")
+        kinitFailCount += 1
+        false
+    }
+  }
+
+  def startKinitDaemon(keytab: String, principal: String): Unit = {
+    executor.schedule(
+      new Runnable() {
+        def run(): Unit = {
```
vanzin: `override`

server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
```diff
@@ -135,6 +167,54 @@ class LivyServer extends Logging {
...
+          if (runKinit(keytab, principal)) {
+            // schedule another kinit run with a fixed delay.
+            executor.schedule(
+              new Runnable() {
+                def run(): Unit = {
+                  runKinit(keytab, principal)
```
vanzin: So, if I read things correctly, this will login once, then login a second time after one hour, and then stop.

Don't you need to schedule `this` instead of this different runnable?

server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
```diff
@@ -135,6 +167,54 @@ class LivyServer extends Logging {
...
+            if (kinitFailCount >= kinitFailThreshold) {
+              error(s"Exit LivyServer after ${kinitFailThreshold} times failures running kinit.")
+              if (server.server.getState == "STARTED") {
+                stop()
+              } else {
+                sys.exit(1)
+              }
+            } else {
+              executor.submit(new Runnable() {
+                def run(): Unit = runKinit(keytab, principal)
```
vanzin: Similarly here; you want to schedule the task that does accounting and re-schedules a login task, not this custom task that only runs kinit and then doesn't do anything else.

vanzin:
Almost there. I think the logic in `startKinitDaemon` is not really correct.

server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
```diff
@@ -135,6 +163,47 @@ class LivyServer extends Logging {
...
+      new Runnable() {
+        override def run(): Unit = {
+          if (runKinit(keytab, principal)) {
+            // schedule another kinit run with a fixed delay.
+            executor.schedule(this, refreshInterval, TimeUnit.MILLISECONDS)
+          } else {
+            // schedule another retry at once or fail the livy server if too many times kinit fail
+            if (kinitFailCount >= kinitFailThreshold) {
+              error(s"Exit LivyServer after ${kinitFailThreshold} times failures running kinit.")
+              if (server.server.getState == "STARTED") {
```
vanzin: This should use `isStarted`. I'll fix during merge.

vanzin:
I'm slightly worried about the aggressive retry with no waits - KDCs generally don't like that, and waiting a little might be the difference between killing the server and running kinit successfully. But in general I expect failures to log in to be rare, so we can tweak that later if needed.

Merging.

PR #146, opened 2016-06-08
Adding user jars to Scala interpreter's class path.

tc0312:
Without this fix, if user imports a package using SparkConf: "spark.jars.packages": "com.databricks:spark-csv_2.10:1.4.0"
Then "import com.databricks.spark.csv._" in the interpreter, it will throw an error because the Scala interpreter cannot see the Spark CSV package.

Also made some small changes in the test framework.

### DIFF
--- a/integration-test/src/main/scala/com/cloudera/livy/test/framework/BaseIntegrationTestSuite.scala
+++ b/integration-test/src/main/scala/com/cloudera/livy/test/framework/BaseIntegrationTestSuite.scala
@@ -105,9 +105,10 @@ abstract class BaseIntegrationTestSuite extends FunSuite with Matchers {
 
   class LivyRestClient(httpClient: AsyncHttpClient, livyEndpoint: String) {
 
-    def startSession(kind: Kind): Int = {
+    def startSession(kind: Kind, sparkConf: Map[String, String] = Map()): Int = {
       val requestBody = new CreateInteractiveRequest()
       requestBody.kind = kind
+      requestBody.conf = sparkConf
 
       val rep = httpClient.preparePost(s"$livyEndpoint/sessions")
         .setBody(mapper.writeValueAsString(requestBody))
--- a/integration-test/src/main/scala/com/cloudera/livy/test/framework/MiniCluster.scala
+++ b/integration-test/src/main/scala/com/cloudera/livy/test/framework/MiniCluster.scala
@@ -19,10 +19,12 @@
 package com.cloudera.livy.test.framework
 
 import java.io._
+import javax.servlet.http.HttpServletResponse
 
 import scala.concurrent.duration._
 import scala.language.postfixOps
 
+import com.ning.http.client.AsyncHttpClient
 import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.fs.Path
 import org.apache.hadoop.hdfs.MiniDFSCluster
@@ -248,6 +250,14 @@ class MiniCluster(config: Map[String, String]) extends Cluster with MiniClusterU
 
     val props = loadProperties(confFile)
     livyUrl = props("livy.server.serverUrl")
+
+    // Wait until Livy server responds.
+    val httpClient = new AsyncHttpClient()
+    eventually(timeout(30 seconds), interval(1 second)) {
+      val res = httpClient.prepareGet(livyUrl).execute().get()
+      assert(res.getStatusCode() == HttpServletResponse.SC_OK)
+    }
+
     livy = Some(localLivy)
   }
 
--- a/repl/src/main/scala/com/cloudera/livy/repl/SparkInterpreter.scala
+++ b/repl/src/main/scala/com/cloudera/livy/repl/SparkInterpreter.scala
@@ -19,10 +19,11 @@
 package com.cloudera.livy.repl
 
 import java.io._
+import java.net.URLClassLoader
+import java.nio.file.Paths
 
 import scala.tools.nsc.Settings
 import scala.tools.nsc.interpreter.{JPrintWriter, Results}
-import scala.tools.nsc.settings.MutableSettings
 import scala.util.{Failure, Success, Try}
 
 import org.apache.spark.{SparkConf, SparkContext}
@@ -32,6 +33,8 @@ import org.json4s.{DefaultFormats, Extraction}
 import org.json4s.JsonAST._
 import org.json4s.JsonDSL._
 
+import com.cloudera.livy.Logging
+
 object SparkInterpreter {
   private val MAGIC_REGEX = "^%(\\w+)\\W*(.*)".r
 }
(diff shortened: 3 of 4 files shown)


integration-test/src/main/scala/com/cloudera/livy/test/framework/BaseIntegrationTestSuite.scala
```diff
@@ -97,17 +97,18 @@ abstract class BaseIntegrationTestSuite extends FunSuite with Matchers {
     }
   }
 
-  test("initialize test cluster") {
+  override def beforeAll() {
```
vanzin: Do you really need this change? I changed this because, if the cluster fails to start, `beforeAll` makes is really, really hard to debug it, since the exception happens during discovery and scalatest doesn't generated good logs in that case.

integration-test/src/main/scala/com/cloudera/livy/test/framework/MiniCluster.scala
```diff
@@ -248,6 +250,14 @@ class MiniCluster(config: Map[String, String]) extends Cluster with MiniClusterU
 
     val props = loadProperties(confFile)
     livyUrl = props("livy.server.serverUrl")
+
+    // Wait until Livy server responses.
```
vanzin: nit: "responds"

repl/src/main/scala/com/cloudera/livy/repl/SparkInterpreter.scala
```diff
@@ -83,6 +86,33 @@ class SparkInterpreter(conf: SparkConf) extends Interpreter {
...
+      var classLoader = Thread.currentThread().getContextClassLoader
+      while (classLoader != null) {
+        if (classLoader.getClass.getSimpleName == "MutableURLClassLoader") {
+          val extraJarPath = classLoader.asInstanceOf[URLClassLoader].getURLs()
+            // Check if the file exists. Otherwise an exception will be thrown.
+            .filter(u => u.getProtocol == "file" && new File(u.getPath).isFile)
+            // Livy rsc and repl are also in the extra jars list. Filter them out.
+            .filterNot(u => Paths.get(u.toURI).getFileName.toString.startsWith("livy-"))
+
+          extraJarPath.foreach(p => info(s"Adding $p to Scala interpreter's class path..."))
```
vanzin: debug?

repl/src/main/scala/com/cloudera/livy/repl/SparkInterpreter.scala
```diff
@@ -83,6 +86,33 @@ class SparkInterpreter(conf: SparkConf) extends Interpreter {
...
+      // "spark.jars.packages": "com.databricks:spark-csv_2.10:1.4.0"
+      // then "import com.databricks.spark.csv._" in the interpreter, it will throw an error.
+      //
+      // Adding them to the interpreter manually to fix this issue.
+      var classLoader = Thread.currentThread().getContextClassLoader
+      while (classLoader != null) {
+        if (classLoader.getClass.getSimpleName == "MutableURLClassLoader") {
+          val extraJarPath = classLoader.asInstanceOf[URLClassLoader].getURLs()
+            // Check if the file exists. Otherwise an exception will be thrown.
+            .filter(u => u.getProtocol == "file" && new File(u.getPath).isFile)
```
vanzin: nit: I prefer `.filter { u => ... }` (since that's the style Spark also uses...)

repl/src/main/scala/com/cloudera/livy/repl/SparkInterpreter.scala
```diff
@@ -83,6 +86,33 @@ class SparkInterpreter(conf: SparkConf) extends Interpreter {
...
+      // them using Scala import statement.
+      //
+      // For instance: If we import a package using SparkConf:
+      // "spark.jars.packages": "com.databricks:spark-csv_2.10:1.4.0"
+      // then "import com.databricks.spark.csv._" in the interpreter, it will throw an error.
+      //
+      // Adding them to the interpreter manually to fix this issue.
+      var classLoader = Thread.currentThread().getContextClassLoader
+      while (classLoader != null) {
+        if (classLoader.getClass.getSimpleName == "MutableURLClassLoader") {
```
vanzin: Could you match the whole class name, just in case?

Also, I know that's currently not used by interactive sessions, but Livy allows adding jars to running sessions (that's exposed in the REST API); it would be good at some point to allow that to work for interactive sessions too. No need to do anything here, but filing a bug would be nice.

integration-test/src/main/scala/com/cloudera/livy/test/framework/BaseIntegrationTestSuite.scala
tc0312: I did this because when I run a specific test e.g. `mvn -Dsuites="com.cloudera.livy.test.InteractiveIT user"`
Test fails because test("initialize test cluster") isn't executed.

Since this is less trivial than it looks, I will do this in a separate CR.

vanzin: I see... maybe a `before` block would work better?

tc0312: > really hard to debug it, since the exception happens during discovery and scalatest doesn't generated good logs in that case.

I injected an exception in beforeAll and it's giving me the full stacktrace. What is your expectation of good logs? Do you mean info()/note() are not showing up for beforeAll()?
`*** RUN ABORTED *** (14 seconds, 853 milliseconds)
  java.lang.Exception: throw form runLivy()
  at com.cloudera.livy.test.framework.MiniCluster.runLivy(MiniCluster.scala:254)
  at com.cloudera.livy.test.framework.MiniCluster.deploy(MiniCluster.scala:234)
  at com.cloudera.livy.test.framework.Cluster$.liftedTree1$1(Cluster.scala:102)
  at com.cloudera.livy.test.framework.Cluster$.cluster$lzycompute(Cluster.scala:90)
  at com.cloudera.livy.test.framework.Cluster$.cluster(Cluster.scala:88)
  at com.cloudera.livy.test.framework.Cluster$.get(Cluster.scala:116)
  at com.cloudera.livy.test.framework.BaseIntegrationTestSuite.beforeAll(BaseIntegrationTestSuite.scala:103)
  at org.scalatest.BeforeAndAfterAll$class.beforeAll(BeforeAndAfterAll.scala:187)
  at com.cloudera.livy.test.framework.BaseIntegrationTestSuite.beforeAll(BaseIntegrationTestSuite.scala:49)
  at org.scalatest.BeforeAndAfterAll$class.run(BeforeAndAfterAll.scala:253)
`

vanzin: I don't remember exact details; but I believe some specific types of errors caused things to be really hard to debug. IIRC beforeAll runs in a static initialization context, and if certain exceptions happen at that point you have no idea what's going on.

I think using `before { }` to initialize the cluster should be enough, right? Only the first invocation takes long, all others are pretty much no-ops. (If doing that also keep the existing "initialize" test with an empty body, just to keep test running times accurate-ish.)

tc0312: That should work. I will do it in a separate PR.

vanzin:
Merging, thanks!

PR #147, opened 2016-06-09
LIVY-163. Fixed missing Python traceback in interactive mode.

tc0312:
- Also stripped the frame of fake_shell in the traceback. It should be transparent to users.

### DIFF
--- a/repl/src/test/scala/com/cloudera/livy/repl/PythonInterpreterSpec.scala
+++ b/repl/src/test/scala/com/cloudera/livy/repl/PythonInterpreterSpec.scala
@@ -21,7 +21,6 @@ package com.cloudera.livy.repl
 import org.apache.spark.SparkConf
 import org.json4s.{DefaultFormats, JValue}
 import org.json4s.JsonDSL._
-import org.scalatest.Outcome
 
 class PythonInterpreterSpec extends BaseInterpreterSpec {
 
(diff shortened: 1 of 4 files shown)


repl/src/test/scala/com/cloudera/livy/repl/PythonInterpreterSpec.scala
```diff
@@ -150,6 +149,7 @@ class PythonInterpreterSpec extends BaseInterpreterSpec {
       "name 'x' is not defined",
       List(
         "Traceback (most recent call last):\n",
+        "  File \"<stdin>\", line 1, in <module>\n",
```
vanzin: I wonder how useful this is? Is there a proper correlation between the line shown here and anything the user has submitted?

From what I remember there is not a 1:1 relationship between a statement submission and a "line" sent to the interpreter.

tc0312: The stack trace is useful when users are using some complicated library (Pandas) and it returns an error.

vanzin: Sorry, I was going to clarify. The stack trace is ok, I meant the frames mentioning `<stdin>` and line numbers.

tc0312: Changed the clean up. Traceback of
```python
def func1():
  raise Exception("message")
def func2():
  func1()
func2()
```
will become
```python
Traceback (most recent call last):
  File "<stdin>", line 4, in func2
  File "<stdin>", line 2, in func1
Exception: message"
```

I don't want to do too many changes at once and will leave line numbers as is. We can revisit line number later.

vanzin:
Merging.

PR #148, opened 2016-06-10
DTTAHOE-77: Create SQL Kind Session on Livy Job Server

meisam:
Implementing a Livy SparkSql Interpreter.

This leaves the SparkInterpreter untouched and confines most of the changes to the livy-repl component.
The SparkSqlInterpreter is based on SparkInterpreter and needs to be polished.
Your feedback is really appreciated.


ksakellis:
Please update this pr to point to the livy bug: LIVY-19

meisam:
I updated the pull request description and reworded the commits to link to LIVY-19.

repl/src/main/scala/com/cloudera/livy/repl/SparkSqlInterpreter.scala
```diff
@@ -0,0 +1,301 @@
...
+
+    executeLines(code.trim.split("\n").toList, Interpreter.ExecuteSuccess(JObject(
+      (TEXT_PLAIN, JString(""))
+    )))
+  }
+
+  override def close(): Unit = synchronized {
+    if (hiveContext != null) {
+      // clean up and close hive context here
+    }
```
zjffdu: maybe set hiveContext as null

repl/src/main/scala/com/cloudera/livy/repl/SparkSqlInterpreter.scala
```diff
@@ -0,0 +1,301 @@
...
+      // Call sparkIMain.setContextClassLoader() to make sure SparkContext and repl are using the
+      // same ClassLoader. Otherwise if someone defined a new class in interactive shell,
+      // SparkContext cannot see them and will result in job stage failure.
+      val setContextClassLoaderMethod = sparkIMain.getClass().getMethod("setContextClassLoader")
+      setContextClassLoaderMethod.setAccessible(true)
+      setContextClassLoaderMethod.invoke(sparkIMain)
+
+      sparkContext = SparkContext.getOrCreate(conf)
+      hiveContext = new HiveContext(sparkContext)
+
```
zjffdu: I think sometimes just SQLContext is sufficient, especailly when there's no hive-site.xml on classpath. Link with #132 which is related .

zjffdu:
Another comment is that maybe we can reuse the SparkContext in SparkInterpreter. We can just add sql support in rest api. e.g. 
```
curl   http://localhost:8998/sessions/0/statements -X POST -H 'Content-Type: application/json' -d '{"sql":"select * from table_1"}'
```

So that for tools like hue and zeppelin, different paragraph can share data easily as they share the same SparkContext.

vanzin:
Hi @meisam,

Before we can even look at this patch, we need a signed ICLA and, if it applies, a signed CCLA from your employer. Please check the wiki: https://github.com/cloudera/livy/wiki/Contributing-to-Livy

vanzin:
> maybe we can reuse the SparkContext in SparkInterpreter

Without having looked at the code, I like that idea.

meisam:
@vanzin 
Where can I find the link to the CCLA? I'll work with my manager to have it signed and sent.

vanzin:
It's at the link I posted above.

meisam:
@zjffdu I was thinking more along these lines

`curl   http://localhost:8998/sessions/0/sql-statements -X POST -H 'Content-Type: application/json' -d '{"code":"select * from table_1"}'`

This keeps `com.cloudera.livy.ExecuteRequest` untouched.

zjffdu:
@meisam  That make sense.

ksakellis:
@meisam Please print, sign and email to cla@cloudera.com both the ICLA:
https://github.com/cloudera/livy/wiki/Individual-Contributor-License-Agreement-(ICLA)
and the CCLA with your name on the Schedule A signed by your employer:
https://github.com/cloudera/livy/wiki/Corporate-Contributor-License-Agreement-(CCLA)

meisam:
@ksakellis I mailed the ICLA and CCLA to cla@cloudera.com last week.

tc0312:
Livy interpreters support 2 magics: [`%json` and `%table`](https://github.com/cloudera/livy/blob/master/repl/src/main/scala/com/cloudera/livy/repl/AbstractSparkInterpreter.scala#L63).
I think we can add another magic: `%sql` to support this use  without adding adding a new REST API.

zjffdu:
Before that, I think we should implement sparkcontext sharing across language #178 , because it is very often user want to access table that is created in spark/pyspark/sparkr interpreter.

tc0312:
Sorry let me clarify myself. I'm purposing to add `%sql` magic to all interpreters. Instead of creating a new interpreter. It shouldn't depend on #178.

meisam:
@tc0312 How would a %sql magic work with SQL code that spans over multiple lines or queries that contains comments in them?

tc0312:
It's up 2 us to define how %sql magic works. We can define it in a way that it works with multiple lines and comments.

meisam:
@tc0312 I guess I should clarify myself. My question is, which approach is easier to implement? Having a %sql magic? Or having a separate SQL interpreter?

meisam:
@tc0312 Actually there's a third approach that @zjffdu suggested: '{"sql":"select * from table_1"}'

tc0312:
I think they are both easy to implement. I would prefer a magic approach because Livy already has magics and cloudera/hue is using it.

sven0726:
What was the final result of this issue?

zjffdu:
It needs more careful design in https://issues.cloudera.org/browse/LIVY-194
We'd like to share sparkcontext across scala/python/R/sql

prabhu1984:
Hi @zjffdu,

Enabling shared spark context across scala/python/R is next phase and it may take sometime.

But, when we support Scala, Python and R interpreters already, can we also add SQL interpreter?

When we SQL interpreter, it will be very easy for users to submit SQL statements directly to interpreter instead of wrapping it up with HiveContext (<2.0) or Spark Session (>2.0) or using SQL magic everytime.

Please let me know your thoughts.

Prabhu

PR #149, opened 2016-06-10
Batch submit to support setting python 2 or 3

KevinGrealish:
Add ability to set Python version for batch submit. See: [https://issues.cloudera.org/browse/LIVY-159](https://issues.cloudera.org/browse/LIVY-159)

Added new fields to batch submit that echo the environment variables that are accepted by spark-submit; PYSPARK_PYTHON and PYSPARK_DRIVER_PYTHON. To prevent ability to execute arbitrary code via these, only "python2" and "python3" are permitted.

### DIFF
--- a/server/src/main/scala/com/cloudera/livy/server/batch/BatchSession.scala
+++ b/server/src/main/scala/com/cloudera/livy/server/batch/BatchSession.scala
@@ -55,7 +55,7 @@ class BatchSession(
     request.queue.foreach(builder.queue)
     request.name.foreach(builder.name)
     request.pyspark_python.foreach(builder.pysparkPython)
-    request.pyspark_python.foreach(builder.pysparkDriverPython) // Always set Driver also to cancel current setting.
+    request.pyspark_python.foreach(builder.pysparkDriverPython) // Always set to replace current.
     request.pyspark_driver_python.foreach(builder.pysparkDriverPython)
     builder.redirectOutput(Redirect.PIPE)
     builder.redirectErrorStream(true)
--- a/server/src/main/scala/com/cloudera/livy/server/batch/BatchSession.scala
+++ b/server/src/main/scala/com/cloudera/livy/server/batch/BatchSession.scala
@@ -38,6 +38,10 @@ class BatchSession(
     val conf = prepareConf(request.conf, request.jars, request.files, request.archives,
       request.pyFiles)
     require(request.file != null, "File is required.")
+    require(!request.pyspark_python.exists(s => s != "python2" && s != "python3"),
+      "pyspark_python can only be set to python2 or python3.")
+    require(!request.pyspark_driver_python.exists(s => s != "python2" && s != "python3"),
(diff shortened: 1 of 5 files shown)


vanzin:
Hi @KevinGrealish, we need you to sign an ICLA before we can merge patches from you. MS has already signed a CCLA so you're good on that front.

You can find the info on the wiki: https://github.com/cloudera/livy/wiki/Contributing-to-Livy

KevinGrealish:
I thought the MS one covered me. Are you saying you need both?

vanzin:
Let me ask around. I was under the impression or legal people required both. :-/

server/src/main/scala/com/cloudera/livy/server/batch/BatchSession.scala
```diff
@@ -38,6 +38,10 @@ class BatchSession(
     val conf = prepareConf(request.conf, request.jars, request.files, request.archives,
       request.pyFiles)
     require(request.file != null, "File is required.")
+    require(!request.pyspark_python.exists(s => s != "python2" && s != "python3"),
```
felixcheung: I think in a lot of environment, Python 2 is simply `python` (eg. CentOS) - could this support that?

ksakellis:
@KevinGrealish You will need to get both the ICLA:
https://github.com/cloudera/livy/wiki/Individual-Contributor-License-Agreement-(ICLA)
and get another CCLA:
https://github.com/cloudera/livy/wiki/Corporate-Contributor-License-Agreement-(CCLA)
signed by your employer Microsoft with your name under the contributors list. 

Each contributor needs to have their name listed on the CCLA from their employer and have an ICLA signed. Please let me know if you have any questions.

sairayapureddy:
Can any one suggest me a documentation or resource for following requirement.
I need to invoke python scripts present in spark from node js server (using livy restful API). I need to check the status of submitted job. This seems basic but I am a starter and started exploring from 2 days.

Stibbons:
Hello. I am not part of the project but very interested in it. Just to give my feedback on this pull request

For the moment we cannot define easily the python environment on Spark, so it might be pretty common to have let's say N environment preinstalled on the Spark cluster's executor. For example, one "Python 2 with pandas" environment, one "Python 3 with tensorflow", one "Anaconda Python 2 with pretty much all libraries", etc... Each are installed on the same folder  on each executor.

To deploy our spark job, we need a way to make them point to the right environment, using the PYSPARK_PYTHON environment variable for instance. If you limit the possible values, your restrict the use of the system python and prevent usage of preinstalled virtual environments.

Let me know what you think.
Regards

KevinGrealish:
That was one of the reasons this change was not made and instead the Spark config was fixed instead so that the python environment variables can be set to any value and Spark will respect them (the conf values were ignored in YARN cluster mode if they were already set system environment vars). See:
https://issues.apache.org/jira/browse/SPARK-16110
https://github.com/apache/spark/pull/13824/files

PR #150, opened 2016-06-10
LIVY-164. Fixed and cleaned up stack trace return from SparkInterpreter.

tc0312:
- Moved stack trace from "evalue" to "traceback".
- Filtered out internal frames from stack trace.

### DIFF
--- a/repl/src/main/scala/com/cloudera/livy/repl/SparkInterpreter.scala
+++ b/repl/src/main/scala/com/cloudera/livy/repl/SparkInterpreter.scala
@@ -39,6 +39,7 @@ object SparkInterpreter {
   private val EXCEPTION_STACK_TRACE_REGEX = """(.+?)\n((?:\tat .+?\n?)*)""".r
   private val KEEP_NEWLINE_REGEX = """(?=\n)""".r
   private val MAGIC_REGEX = "^%(\\w+)\\W*(.*)".r
+  val USER_CODE_FRAME_NAME = "<user code>"
 }
 
 /**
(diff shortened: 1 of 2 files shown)


vanzin:
Hey @tc0312, do you mind filing bugs for these patches you're sending? thanks

repl/src/main/scala/com/cloudera/livy/repl/SparkInterpreter.scala
```diff
@@ -298,7 +300,26 @@ class SparkInterpreter(conf: SparkConf) extends Interpreter with Logging {
...
+                stdout match {
+                  case EXCEPTION_STACK_TRACE_REGEX(ename, tracebackLines) =>
+                    var traceback = KEEP_NEWLINE_REGEX.pattern.split(tracebackLines)
+                    val interpreterFrameIdx = traceback.indexWhere(_.contains("$iwC$$iwC.<init>"))
+                    if (interpreterFrameIdx >= 0) {
+                      traceback = traceback
+                        // Remove Interpreter frames
+                        .take(interpreterFrameIdx)
+                        // Replace weird internal class name
+                        .map(_.replace("$iwC$$iwC", "UserCode"))
```
vanzin: Could you use `<user code>` instead? `UserCode` looks too much like a class name.

vanzin: (You could also make it a constant and avoid copy & pasting it in the test code.)

vanzin:
One minor issue otherwise looks good.

vanzin:
Merging.

PR #151, opened 2016-06-11
LIVY-92. Scala-friendly API for Livy client sessions

manikandan89:
- Added a Scala API which is a wrapper over the existing Java API
- Jobs can be submitted with a Scala-friendly JobContext eliminating the overhead of creating new job instances
- The submit API returns a Scala JobHandle which supports the existing Java Listener callbacks


scala-api/src/main/scala/com/cloudera/livy/scalaapi/ScalaJobHandle.scala
```diff
@@ -0,0 +1,123 @@
...
+import scala.concurrent.{CanAwait, ExecutionContext, Future, TimeoutException}
+import scala.concurrent.duration.Duration
+import scala.util.Try
+
+import com.cloudera.livy.JobHandle
+import com.cloudera.livy.JobHandle.{Listener, State}
+
+class ScalaJobHandle[T] private[livy] (jobHandle: JobHandle[T]) extends Future[T] {
+
+  def getState(): State = jobHandle.getState
```
vanzin: `def state: State = jobHandle.getState()` is more Scala-ish.

scala-api/src/main/scala/com/cloudera/livy/scalaapi/ScalaJobHandle.scala
```diff
@@ -0,0 +1,123 @@
...
+      override def onJobFailed(job: JobHandle[T], cause: Throwable): Unit = {
+        val onJobFailedTask = new Runnable {
+          override def run(): Unit = func(Try(getJavaFutureResult(job)))
+        }
+        executor.execute(onJobFailedTask)
+      }
+    })
+  }
+
+  def onJobQueued[U](func: Unit => Unit)(implicit executor: ExecutionContext): Unit = {
```
vanzin: I think this (and other similar callbacks) should be `def onJobQueued[U](func: => Unit)`. That makes the syntax for the caller more natural. With your code you need this:

     handle.onJobQueued { _ => doSomething() }

With `=> Unit` you just write this:

     handle.onJobQueued { doSomething() }

It's a minor difference but helps a lot with readability (and requires typing fewer unnecessary operators).

scala-api/src/main/scala/com/cloudera/livy/scalaapi/ScalaJobHandle.scala
```diff
@@ -0,0 +1,123 @@
...
+  private def getJavaFutureResultAfterGivenWaitTime(atMost: Duration): T = {
+    if (!atMost.isFinite()) {
+      getJavaFutureResult(jobHandle)
+    } else {
+      getJavaFutureResult(jobHandle, atMost)
+    }
+  }
+}
+
+private abstract class AbstractScalaJobHandleListener[T] private[livy] extends Listener[T] {
```
vanzin: `private[livy]` not needed here.

scala-api/src/main/scala/com/cloudera/livy/scalaapi/package.scala
```diff
@@ -0,0 +1,40 @@
...
+
+package com.cloudera.livy
+
+import java.util.concurrent.{ExecutionException, Future => JFuture, TimeUnit}
+
+import scala.concurrent.duration.Duration
+
+package object scalaapi {
+
+  implicit class ScalaWrapper private[livy] (livyJavaClient: LivyClient) {
```
vanzin: Pretty sure you don't want `private[livy]` here or nobody outside Livy's code base will be able to call it.

scala-api/src/main/scala/com/cloudera/livy/scalaapi/package.scala
```diff
@@ -0,0 +1,40 @@
...
+    def asScalaClient: LivyScalaClient = new LivyScalaClient(livyJavaClient)
+  }
+
+  def getJavaFutureResult[T](jFuture: JFuture[T], atMost: Duration = Duration.Undefined): T = {
```
vanzin: Add `private[livy]` to this one.

scala-api/src/test/scala/com/cloudera/livy/scalaapi/ScalaClientTest.scala
```diff
@@ -0,0 +1,245 @@
...
+      }
+      case Failure(e) => {
+        if (!e.getMessage.contains("CustomTestFailureException")) {
+          hasTestPassed = false
+          testFailureMessage = "Test did not throw expected exception - CustomFailureException"
+        }
+        ScalaClientTest.notify(lock)
+      }
+    }
+    ScalaClientTest.wait(lock)
```
vanzin: This is not thread-safe. If the callback runs before this line is executed, this call will hang forever since it will miss the notify.

You need to register the callback while holding the lock, or you need to use something like a semaphore.

scala-api/src/test/scala/com/cloudera/livy/scalaapi/ScalaClientTest.scala
```diff
@@ -0,0 +1,245 @@
...
+    if (!hasTestPassed) fail(testFailureMessage)
```
vanzin: Minor, but this would be cleaner:

```
var testFailure: Option[String] = None

// When test fails:
testFailure = Some("Error")

testFailure.foreach(fail)
```

scala-api/src/test/scala/com/cloudera/livy/scalaapi/ScalaClientTest.scala
```diff
@@ -0,0 +1,245 @@
...
+  test("test add file") {
+    configureClient(true)
+    var file: File = null
+    file = File.createTempFile("test", ".file")
+    val fileStream = new FileOutputStream(file)
+    fileStream.write("test file".getBytes("UTF-8"))
+    fileStream.close
+    val addFileFuture = client.addFile(new URI("file:" + file.getAbsolutePath()))
+    Await.ready(addFileFuture, 5 second)
+    val sFuture = client.submit(
```
vanzin:
```
val sFuture = client.submit { context =>
  ScalaClientTest.fileOperation(false, file.getName, context)
}
```

scala-api/src/test/scala/com/cloudera/livy/scalaapi/ScalaClientTest.scala
```diff
@@ -0,0 +1,245 @@
...
+      context => ScalaClientTest.fileOperation(false, file.getName, context)
+    )
+    val output = Await.result(sFuture, 10 second)
+    assert(output === "test file")
+  }
+
+  test("test add jar") {
+    configureClient(true)
+    var jar: File = null
```
vanzin: `val jar = File.createTempFile("test", ".resource")`

scala-api/src/test/scala/com/cloudera/livy/scalaapi/ScalaClientTest.scala
```diff
@@ -0,0 +1,245 @@
...
+    jar = File.createTempFile("test", ".resource")
+    val jarFile = new JarOutputStream(new FileOutputStream(jar))
+    jarFile.putNextEntry(new ZipEntry("test.resource"))
+    jarFile.write("test resource".getBytes("UTF-8"))
+    jarFile.closeEntry()
+    jarFile.close()
+    val addJarFuture = client.addJar(new URI("file:" + jar.getAbsolutePath()))
+    Await.ready(addJarFuture, 5 second)
+    val sFuture = client.submit(
```
vanzin: See previous style nit.

scala-api/src/test/scala/com/cloudera/livy/scalaapi/ScalaClientTest.scala
```diff
@@ -0,0 +1,245 @@
...
+            testFailureMessage = "onComplete has not returned the expected message"
+          }
+        }
+        case Failure(e) => {
+          hasTestPassed = false;
+          testFailureMessage = "onComplete should not have triggered Failure callback"
+        }
+      }
+    }
+    if (!hasTestPassed) fail(testFailureMessage)
```
vanzin: This is not thread-safe. Because you're using the global execution context, callbacks may not have run when you get to this point. So if the test fails, you might miss that and report a false position.

Also, you should keep a count of how many callbacks you expect to run, since that's what you're testing here. This is another place where a semaphore would be useful.

scala-api/src/test/scala/com/cloudera/livy/scalaapi/ScalaClientTest.scala
```diff
@@ -0,0 +1,245 @@
...
+      conf.put(SparkLauncher.DRIVER_EXTRA_CLASSPATH, classpath)
+      conf.put(SparkLauncher.EXECUTOR_EXTRA_CLASSPATH, classpath)
+    }
+    conf.put(LIVY_JARS.key, "")
+    conf
+  }
+
+  def fileOperation(isResource: Boolean, fileName: String, context: ScalaJobContext): String = {
+    val arr = Seq(1)
+    val rdd = context.sc.parallelize(arr).map(value => {
```
vanzin:
```
.map { value =>
  // code
}
```

scala-api/src/test/scala/com/cloudera/livy/scalaapi/ScalaJobHandleTest.scala
```diff
@@ -0,0 +1,203 @@
...
+        }
+        ScalaClientTest.notify(lock)
+      }
+      case Failure(e) => {
+        hasTestPassed = false;
+        testFailureMessage = "onComplete should not have triggered Failure callback"
+        ScalaClientTest.notify(lock)
+      }
+    }
+    ScalaClientTest.wait(lock)
```
vanzin: Same thread-safety issue here.

scala-api/src/test/scala/com/cloudera/livy/scalaapi/ScalaJobHandleTest.scala
```diff
@@ -0,0 +1,203 @@
...
+      }
+      case Failure(e) => {
+        if (!e.isInstanceOf[CustomTestFailureException]) {
+          hasTestPassed = false
+          testFailureMessage = "Test did not throw expected exception - CustomFailureException"
+        }
+        ScalaClientTest.notify(lock)
+      }
+    }
+    ScalaClientTest.wait(lock)
```
vanzin: And here.

scala-api/src/test/scala/com/cloudera/livy/scalaapi/ScalaJobHandleTest.scala
```diff
@@ -0,0 +1,203 @@
...
+    val lock = new Object
+    val testScalaHandle = new ScalaJobHandle(jobHandleStub)
+    testScalaHandle onJobCancelled  {
+      case true => ScalaClientTest.notify(lock)
+      case false => {
+        hasTestPassed = false
+        ScalaClientTest.notify(lock)
+      }
+    }
+    ScalaClientTest.wait(lock)
```
vanzin: And here.

scala-api/src/test/scala/com/cloudera/livy/scalaapi/ScalaJobHandleTest.scala
```diff
@@ -0,0 +1,203 @@
...
+    var hasTestPassed = false
+    val lock = new Object
+    val testScalaHandle = new ScalaJobHandle(jobHandleStub)
+    testScalaHandle onJobQueued {
+      case () => {
+        hasTestPassed = true
+        ScalaClientTest.notify(lock)
+      }
+    }
+    ScalaClientTest.wait(lock)
```
vanzin: And here.

scala-api/src/test/scala/com/cloudera/livy/scalaapi/ScalaJobHandleTest.scala
```diff
@@ -0,0 +1,203 @@
...
+    var hasTestPassed = false
+    val lock = new Object
+    val testScalaHandle = new ScalaJobHandle(jobHandleStub)
+    testScalaHandle onJobStarted {
+      case () => {
+        hasTestPassed = true
+        ScalaClientTest.notify(lock)
+      }
+    }
+    ScalaClientTest.wait(lock)
```
vanzin: And here.

scala-api/src/test/scala/com/cloudera/livy/scalaapi/ScalaClientTest.scala
```diff
@@ -0,0 +1,243 @@
...
+        lock.countDown()
+      }
+      case Failure(e) => {
+        if (!e.getMessage.contains("CustomTestFailureException")) {
+          testFailure = Some("Test did not throw expected exception - CustomFailureException")
+        }
+        lock.countDown()
+      }
+    }
+    if (lock.getCount == 1) lock.await()
```
vanzin: You can just call `lock.await()`. But you should use a timeout, because if the callback never fires, this will hang forever.

scala-api/src/test/scala/com/cloudera/livy/scalaapi/ScalaClientTest.scala
```diff
@@ -0,0 +1,243 @@
...
+    testFailure.foreach(fail(_))
```
vanzin: this can just be `testFailure.foreach(fail)`.

scala-api/src/test/scala/com/cloudera/livy/scalaapi/ScalaClientTest.scala
```diff
@@ -0,0 +1,243 @@
...
+  }
+
+  test("Successive onComplete callbacks") {
+    var testFailure: Option[String] = None
+    configureClient(true)
+    val future = client.run(ScalaClientTest.helloJob)
+    var count: Integer = 0
+    breakable {
+      for (i <- 0 to 2) {
+        val lock = new CountDownLatch(1)
```
vanzin: Why use a separate latch for each iteration? Just have a single latch outside of the loop, with the total count. That avoids having to use weird things like "breakable".

scala-api/src/test/scala/com/cloudera/livy/scalaapi/ScalaClientTest.scala
```diff
@@ -0,0 +1,243 @@
...
+        future onComplete {
+          case Success(t) => {
+            if (!t.equals("hello")) testFailure = Some("Expected message not returned")
```
vanzin: nit: expand blocks like this.

```
if (condition) {
  doSomething()
}
```

The extra verbosity helps with readability.

scala-api/src/test/scala/com/cloudera/livy/scalaapi/ScalaJobHandleTest.scala
```diff
@@ -0,0 +1,195 @@
...
+    }
+    var hasTestPassed = false
+    val lock = new CountDownLatch(1)
+    val testScalaHandle = new ScalaJobHandle(jobHandleStub)
+    testScalaHandle onJobQueued {
+      hasTestPassed = true
+      lock.countDown()
+    }
+    if (lock.getCount == 1) lock.await()
+    if (!hasTestPassed) fail("Callback not triggered")
```
vanzin: This check is not necessary; if the callback is not called, the previous statement will block forever. Which is why you should use a timeout in the call to `await` (and assert that it returns true).

vanzin:
Your test code has a lot of duplication, especially around checking whether the tests succeeded or not; if you try to come up with some shared code to handle that, it would be both be nicer, and easier to fix any issues since there would be a single place to change.

scala-api/src/main/scala/com/cloudera/livy/scalaapi/ScalaJobHandle.scala
```diff
@@ -0,0 +1,123 @@
...
+
+  @throws(classOf[InterruptedException])
+  @throws(classOf[TimeoutException])
+  override def ready(atMost: Duration)(implicit permit: CanAwait): ScalaJobHandle.this.type = {
+    getJavaFutureResultAfterGivenWaitTime(atMost)
+    this
+  }
+
+  private def getJavaFutureResultAfterGivenWaitTime(atMost: Duration): T = {
+    if (!atMost.isFinite()) {
```
vanzin: `getJavaFutureResult` already handles this right? No need to duplicate the check. Then you don't need this method.

scala-api/src/main/scala/com/cloudera/livy/scalaapi/LivyScalaClient.scala
```diff
@@ -0,0 +1,88 @@
...
+import java.util.concurrent.{Executors, Future => JFuture, ScheduledFuture, TimeUnit}
+
+import scala.concurrent._
+import scala.util.Try
+
+import com.cloudera.livy._
+
+class LivyScalaClient(livyJavaClient: LivyClient) {
+
+  private val executor = Executors.newSingleThreadScheduledExecutor()
```
vanzin: Missed this before; but it's better to create an executor with a custom `ThreadFactory` that creates daemon threads. There's an example in `HttpClient.java`.

scala-api/src/main/scala/com/cloudera/livy/scalaapi/LivyScalaClient.scala
```diff
@@ -0,0 +1,88 @@
...
+    new PollingContainer(livyJavaClient.uploadFile(file)).poll()
+
+  def addFile(uRI: URI): Future[_] = new PollingContainer(livyJavaClient.addFile(uRI)).poll()
+
+  private class PollingContainer[T] private[livy] (jFuture: JFuture[T]) extends Runnable {
+
+    private val initialDelay = 1
+    private val longDelay = 1
+    private var scheduledFuture: ScheduledFuture[_] = _
+    val promise = Promise[T]
```
vanzin: This should be private.

scala-api/src/test/scala/com/cloudera/livy/scalaapi/ScalaClientTestUtils.scala
```diff
@@ -0,0 +1,57 @@
...
+    val partitions = Math.min(r.nextInt(10) + 1, count)
+    val buffer = new ArrayBuffer[Int]()
+    for (a <- 1 to count) {
+      buffer += r.nextInt()
+    }
+    context.sc.parallelize(buffer, partitions).count()
+  }
+
+  def assertAwait(lock: CountDownLatch): Unit = {
+    if (lock.getCount == 1) {
```
vanzin: This check is not needed (and in fact is wrong, given that in some tests the count might be more than 1).

scala-api/src/test/scala/com/cloudera/livy/scalaapi/ScalaClientTestUtils.scala
```diff
@@ -0,0 +1,57 @@
...
+      assert(lock.await(Timeout, TimeUnit.SECONDS) == true)
+    }
+  }
+
+  def assertTestPassed(future: Future[Any], expectedValue: Any): Unit = {
```
vanzin: Minor, but to get help from the compiler, this should  be:

    def assertTestPassed[T](future: Future[T], expectedValue: T): Unit = {

vanzin:
A few minor things left, but overall looks good.

vanzin:
Alright, I'm gonna merge this. There's still some cleanup that can be done here, and there's work around usability and documentation that needs to be done, but that can be done separately.

PR #154, opened 2016-06-17
LIVY-174. CSRF protection in livy rest api

zjffdu:
Add CSRFFilter in livy server. By default it is false, so it won't bring incomptability and break the existing application.

### DIFF
--- a/conf/livy.conf
+++ b/conf/livy.conf
@@ -38,3 +38,7 @@
 # default it's empty, meaning users can only reference remote URIs when starting their
 # sessions.
 # livy.file.local-dir-whitelist =
+
+# Whether to enable csrf protection, by default it is false. If it is enabled, client should add
+# http-header "X-Requested-By" in request if the http method is POST/DELETE/PATCH.
+# livy.server.csrf_protection.enabled =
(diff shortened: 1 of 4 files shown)


vanzin:
Could you add the new config and a quick description in `conf/livy.conf`?

It would also be cool to automatically enable this for new API versions - but that would require creating a new API version, so probably better to leave it as is for now.

conf/livy.conf
```diff
@@ -38,3 +38,7 @@
 # default it's empty, meaning users can only reference remote URIs when starting their
 # sessions.
 # livy.file.local-dir-whitelist =
+
+# Whether to enable csrf protection, by default it is false. If it is enabled, client should add
+# http-header "X-Requested-By" in request if the http method is POST/DELETE/PATCH.
```
vanzin: nit: s/PATCH/PUT (I don't think we have any "patch" endpoints). I'll fix it during merge.

vanzin:
Cool, merging.

vanzin:
Actually... could you update the client-http code to always set the header? That way it will work regardless of the server-side setting.

conf/livy.conf
zjffdu: I think both PATCH/PUT are not used in livy api. But I just notice they are defined in `JsonServlet` and only be used in `TestJsonServlet`. So I will just keep them here in case they will be used in api in future.

vanzin:
Merging.

PR #155, opened 2016-06-17
LIVY-177. Added unicode support output for PySpark session.

tc0312:
fake_shell is using cStringIO to capture stdout and stderr.
cStringIO doesn't support unicode. Replacing it with io.StringIO.
io.StringIO supports unicode only, convert every output to utf8.

### DIFF
--- a/repl/src/main/resources/fake_shell.py
+++ b/repl/src/main/resources/fake_shell.py
@@ -18,14 +18,14 @@ import ast
 import cStringIO
 import datetime
 import decimal
+import io
 import json
 import logging
 import sys
 import traceback
-import StringIO
-import base64
 import os
 import re
+import StringIO
 
 logging.basicConfig()
 LOG = logging.getLogger('fake_shell')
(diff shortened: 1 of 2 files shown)


vanzin:
LGTM, merging. At some point I should look at making sure scalastyle ignores the contents of strings and comments (I just don't want non-ASCII in regular source code).

qwemicheal:
I am building Livy using the latest 0.4 snapshot code and failed with the following output:
`Python2InterpreterSpec:  
**- should print unicode correctly *** FAILED *** (103 milliseconds)
  ExecuteSuccess(JObject(List((text/plain,JString(âº))))) did not equal ExecuteSuccess(JObject(List((text/plain,JString(☺))))) (PythonInterpreterSpec.scala:260)
**`

On PythonInterpreterSpec.scala:260 I find :
`intp.execute("""print("\xE2\x98\xBA")""") should equal(Interpreter.ExecuteSuccess(`

I have both python 2&3 installed. In python 2 ,`print("\xE2\x98\xBA"` can actually print a smiley face as the following:

`Python 2.7.12 (default, Nov 19 2016, 06:48:10) 
[GCC 5.4.0 20160609] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> print("\xE2\x98\xBA"
... )
☺
`

To successfully build Livy, I have to comment out that line of code. 
Is there something wrong with this test case ?

PR #163, opened 2016-07-09
LIVY-184. Better YARN integration for Batch Session.

tc0312:
When livy is running with YARN master:
- Stop and log session using YARN API.
- Exposed app id.
- Livy now requires Spark 1.6+. Added a check so Livy won't start if it's running with older Spark.

Test changes:
- Added a test to verify killing the batch session will kill its YARN app.
- Added a test to verify killing the YARN app will change the batch session state.
- Added YarnClient integration in test framework.
- Added halt mode in SimpleSparkApp in test lib for kill session test.
- Added a method to dump logs when Batch test fails.
- Added a Clock interface to allow unit testing to mock out time.

### DIFF
--- a/integration-test/src/main/scala/com/cloudera/livy/test/framework/BaseIntegrationTestSuite.scala
+++ b/integration-test/src/main/scala/com/cloudera/livy/test/framework/BaseIntegrationTestSuite.scala
@@ -105,11 +105,6 @@ abstract class BaseIntegrationTestSuite extends FunSuite with Matchers with Befo
     livyClient = new LivyRestClient(httpClient, livyEndpoint)
   }
 
-  test("initialize test cluster") {
-    // Empty test case to separate time spent on creating cluster in beforeAll() and
-    // executing actual test cases.
-  }
-
   class LivyRestClient(httpClient: AsyncHttpClient, livyEndpoint: String) {
 
     def startSession(kind: Kind, sparkConf: Map[String, String] = Map()): Int = {
--- a/integration-test/src/main/scala/com/cloudera/livy/test/framework/BaseIntegrationTestSuite.scala
+++ b/integration-test/src/main/scala/com/cloudera/livy/test/framework/BaseIntegrationTestSuite.scala
@@ -46,7 +46,7 @@ object BaseIntegrationTestSuite {
   val ERROR = "error"
 }
 
-abstract class BaseIntegrationTestSuite extends FunSuite with Matchers with BeforeAndAfter {
+abstract class BaseIntegrationTestSuite extends FunSuite with Matchers with BeforeAndAfterAll {
   import BaseIntegrationTestSuite._
 
   var cluster: Cluster = _
@@ -97,15 +97,17 @@ abstract class BaseIntegrationTestSuite extends FunSuite with Matchers with Befo
     }
   }
 
-  before {
+  // We need beforeAll() here because BatchIT's beforeAll() has to be executed after this.
+  // Please create an issue if this breaks test logging for cluster creation.
+  protected override def beforeAll() = {
     cluster = Cluster.get()
     httpClient = new AsyncHttpClient()
     livyClient = new LivyRestClient(httpClient, livyEndpoint)
   }
 
   test("initialize test cluster") {
-    // Empty test case to separate time spent on creating cluster in before() and executing actual
-    // test cases.
+    // Empty test case to separate time spent on creating cluster in beforeAll() and
+    // executing actual test cases.
   }
 
   class LivyRestClient(httpClient: AsyncHttpClient, livyEndpoint: String) {
--- a/integration-test/src/main/scala/com/cloudera/livy/test/framework/Cluster.scala
+++ b/integration-test/src/main/scala/com/cloudera/livy/test/framework/Cluster.scala
@@ -28,6 +28,7 @@ import scala.util.Try
 import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.fs.FileSystem
 import org.apache.hadoop.fs.Path
+import org.apache.hadoop.yarn.client.api.YarnClient
 
 import com.cloudera.livy.Logging
 
@@ -48,20 +49,26 @@ trait Cluster {
 
   def doAsClusterUser[T](task: => T): T
 
-  lazy val hadoopConf = {
-    val conf = new Configuration(false)
-    configDir().listFiles().foreach { f =>
-      if (f.getName().endsWith(".xml")) {
-        conf.addResource(new Path(f.toURI()))
-      }
-    }
-    conf
-  }
+  lazy val coreSiteConf = loadConf(s"${configDir().getCanonicalPath}/core-site.xml")
+
+  lazy val yarnSiteConf = loadConf(s"${configDir().getCanonicalPath}/yarn-site.xml")
 
   lazy val fs = doAsClusterUser {
-    FileSystem.get(hadoopConf)
+    FileSystem.get(coreSiteConf)
   }
 
+  lazy val yarnClient = doAsClusterUser {
+    val c = YarnClient.createYarnClient()
+    c.init(yarnSiteConf)
+    c.start()
+    c
+  }
+
+  private def loadConf(confPath: String): Configuration = {
+    val conf = new Configuration(false)
+    conf.addResource(new Path(confPath))
+    conf
+  }
 }
 
 object Cluster extends Logging {
--- a/integration-test/src/main/scala/com/cloudera/livy/test/framework/Cluster.scala
+++ b/integration-test/src/main/scala/com/cloudera/livy/test/framework/Cluster.scala
@@ -49,25 +49,31 @@ trait Cluster {
 
   def doAsClusterUser[T](task: => T): T
 
-  lazy val coreSiteConf = loadConf(s"${configDir().getCanonicalPath}/core-site.xml")
+  lazy val hadoopConf = {
+    val conf = new Configuration(false)
+    configDir().listFiles().foreach { f =>
+      if (f.getName().endsWith(".xml")) {
+        conf.addResource(new Path(f.toURI()))
+      }
+    }
+    conf
+  }
 
-  lazy val yarnSiteConf = loadConf(s"${configDir().getCanonicalPath}/yarn-site.xml")
+  lazy val yarnConf = {
+    val conf = new Configuration(false)
+    conf.addResource(new Path(s"${configDir().getCanonicalPath}/yarn-site.xml"))
+    conf
+  }
 
   lazy val fs = doAsClusterUser {
-    FileSystem.get(coreSiteConf)
+    FileSystem.get(hadoopConf)
   }
 
   lazy val yarnClient = doAsClusterUser {
     val c = YarnClient.createYarnClient()
-    c.init(yarnSiteConf)
+    c.init(yarnConf)
     c.start()
     c
-  }
-
-  private def loadConf(confPath: String): Configuration = {
-    val conf = new Configuration(false)
-    conf.addResource(new Path(confPath))
-    conf
   }
 }
 
--- a/integration-test/src/test/scala/com/cloudera/livy/test/BatchIT.scala
+++ b/integration-test/src/test/scala/com/cloudera/livy/test/BatchIT.scala
@@ -57,21 +57,21 @@ class BatchIT extends BaseIntegrationTestSuite with BeforeAndAfterAll {
 
   test("submit spark app") {
     assume(testLibPath != null, "Test lib not uploaded.")
-    val output = newOutputPath
+    val output = newOutputPath()
     val result = runSpark(classOf[SimpleSparkApp], args = List(output))
 
     dumpLogOnFailure(result.id) {
-      assert(result.state === SessionState.Success().toString)
+      assert(result.state === SessionState.Success().toString())
       assert(cluster.fs.isDirectory(new Path(output)))
     }
   }
 
   test("submit an app that fails") {
     assume(testLibPath != null, "Test lib not uploaded.")
-    val output = newOutputPath
+    val output = newOutputPath()
     val result = runSpark(classOf[FailingApp], args = List(output))
     // At this point the application has exited. State should be 'dead' instead of 'error'.
-    assert(result.state === SessionState.Dead().toString)
+    assert(result.state === SessionState.Dead().toString())
 
     // The file is written to make sure the app actually ran, instead of just failing for
     // some other reason.
@@ -80,9 +80,9 @@ class BatchIT extends BaseIntegrationTestSuite with BeforeAndAfterAll {
 
   pytest("submit a pyspark application") {
     val hdfsPath = uploadResource("pytest.py")
-    val output = newOutputPath
+    val output = newOutputPath()
     val result = runScript(hdfsPath, args = List(output))
-    assert(result.state === SessionState.Success().toString)
+    assert(result.state === SessionState.Success().toString())
     assert(cluster.fs.isDirectory(new Path(output)))
   }
 
@@ -91,12 +91,12 @@ class BatchIT extends BaseIntegrationTestSuite with BeforeAndAfterAll {
   ignore("submit a SparkR application") {
     val hdfsPath = uploadResource("rtest.R")
     val result = runScript(hdfsPath)
-    assert(result.state === SessionState.Success().toString)
+    assert(result.state === SessionState.Success().toString())
   }
 
   test("deleting a session should kill YARN app") {
     assume(testLibPath != null, "Test lib not uploaded.")
-    val output = newOutputPath
+    val output = newOutputPath()
     val batchId = runSpark(classOf[SimpleSparkApp], List(output, "false"), waitForExit = false).id
 
     dumpLogOnFailure(batchId, cleanup = false) {
@@ -112,7 +112,7 @@ class BatchIT extends BaseIntegrationTestSuite with BeforeAndAfterAll {
 
   test("killing YARN app should change batch state to dead") {
     assume(testLibPath != null, "Test lib not uploaded.")
-    val output = newOutputPath
+    val output = newOutputPath()
     val batchId = runSpark(classOf[SimpleSparkApp], List(output, "false"), waitForExit = false).id
 
     dumpLogOnFailure(batchId) {
@@ -125,7 +125,7 @@ class BatchIT extends BaseIntegrationTestSuite with BeforeAndAfterAll {
           .getFinalApplicationStatus == FinalApplicationStatus.KILLED,
           "YARN app should be killed.")
         val batchInfo = getBatchSessionInfo(batchId)
-        assert(batchInfo.state == SessionState.Dead().toString, "Batch state should be dead.")
+        assert(batchInfo.state == SessionState.Dead().toString(), "Batch state should be dead.")
         assert(batchInfo.log.contains("Application killed by user."),
           "Batch log doesn't contain yarn final diagnostics.")
       }
@@ -157,12 +157,12 @@ class BatchIT extends BaseIntegrationTestSuite with BeforeAndAfterAll {
       val appReport = cluster.yarnClient.getApplicationReport(ConverterUtils.toApplicationId(appId))
       assert(appReport != null, "appReport shouldn't be null")
 
-      appReport.getDiagnostics
-    } getOrElse ""
+      appReport.getDiagnostics()
+    }.getOrElse("")
   }
 
-  private def newOutputPath: String = {
-    cluster.hdfsScratchDir().toString + "/" + UUID.randomUUID().toString
+  private def newOutputPath(): String = {
+    cluster.hdfsScratchDir().toString() + "/" + UUID.randomUUID().toString()
   }
 
   private def uploadResource(name: String): String = {
@@ -175,7 +175,7 @@ class BatchIT extends BaseIntegrationTestSuite with BeforeAndAfterAll {
       in.close()
       out.close()
     }
-    hdfsPath.toUri.getPath
+    hdfsPath.toUri().getPath()
   }
 
   private def runScript(script: String, args: List[String] = Nil): SessionInfo = {
@@ -202,7 +202,7 @@ class BatchIT extends BaseIntegrationTestSuite with BeforeAndAfterAll {
     val r = httpClient.prepareGet(s"$livyEndpoint/batches/$batchId")
       .execute()
       .get()
-    assert(r.getStatusCode === SC_OK)
+    assert(r.getStatusCode() === SC_OK)
     mapper.readValue(r.getResponseBodyAsStream(), classOf[SessionInfo])
   }
 
@@ -210,7 +210,7 @@ class BatchIT extends BaseIntegrationTestSuite with BeforeAndAfterAll {
     val r = httpClient.prepareDelete(s"$livyEndpoint/batches/$batchId")
       .execute()
       .get()
-    assert(r.getStatusCode === SC_OK)
+    assert(r.getStatusCode() === SC_OK)
   }
 
   private def startBatch(request: CreateBatchRequest): SessionInfo = {
@@ -233,14 +233,14 @@ class BatchIT extends BaseIntegrationTestSuite with BeforeAndAfterAll {
       val batch = getBatchSessionInfo(batchId)
 
       // If batch state transits to any error state, fail test immediately.
-      if (batch.state == SessionState.Error().toString ||
-        batch.state == SessionState.Dead().toString) {
+      if (batch.state == SessionState.Error().toString() ||
+        batch.state == SessionState.Dead().toString()) {
         throw new TestPendingException {
           override def getMessage = s"Session shouldn't be in a terminal state: ${batch.state}"
         }
       }
 
-      assert(batch.state === SessionState.Running().toString)
+      assert(batch.state === SessionState.Running().toString())
 
       ConverterUtils.toApplicationId(batch.appId)
     }
@@ -248,7 +248,7 @@ class BatchIT extends BaseIntegrationTestSuite with BeforeAndAfterAll {
 
   private def waitAndDeleteBatch(batchId: Int): SessionInfo = {
     val terminalStates = Set(SessionState.Error(), SessionState.Dead(), SessionState.Success())
-      .map (_.toString)
+      .map(_.toString())
 
     var finished = false
     try {
--- a/integration-test/src/test/scala/com/cloudera/livy/test/BatchIT.scala
+++ b/integration-test/src/test/scala/com/cloudera/livy/test/BatchIT.scala
@@ -32,7 +32,6 @@ import org.apache.hadoop.yarn.api.records.{ApplicationId, FinalApplicationStatus
 import org.apache.hadoop.yarn.util.ConverterUtils
 import org.scalatest.BeforeAndAfterAll
 import org.scalatest.concurrent.Eventually._
-import org.scalatest.exceptions.TestPendingException
 
 import com.cloudera.livy.client.common.HttpMessages._
 import com.cloudera.livy.server.batch.CreateBatchRequest
@@ -50,13 +49,7 @@ class BatchIT extends BaseIntegrationTestSuite with BeforeAndAfterAll {
     testLibPath = uploadToHdfs(new File(testLib))
   }
 
-  test("upload test lib") {
-    // Empty test case to separate time spent on uploading test lib in beforeAll() and
-    // executing actual test cases.
-  }
-
   test("submit spark app") {
-    assume(testLibPath != null, "Test lib not uploaded.")
     val output = newOutputPath()
     val result = runSpark(classOf[SimpleSparkApp], args = List(output))
 
@@ -67,7 +60,6 @@ class BatchIT extends BaseIntegrationTestSuite with BeforeAndAfterAll {
   }
 
   test("submit an app that fails") {
-    assume(testLibPath != null, "Test lib not uploaded.")
     val output = newOutputPath()
     val result = runSpark(classOf[FailingApp], args = List(output))
     // At this point the application has exited. State should be 'dead' instead of 'error'.
@@ -95,12 +87,11 @@ class BatchIT extends BaseIntegrationTestSuite with BeforeAndAfterAll {
   }
 
   test("deleting a session should kill YARN app") {
-    assume(testLibPath != null, "Test lib not uploaded.")
     val output = newOutputPath()
     val batchId = runSpark(classOf[SimpleSparkApp], List(output, "false"), waitForExit = false).id
 
     dumpLogOnFailure(batchId, cleanup = false) {
-      val appId = waitBatchForRunning(batchId)
+      val appId = waitUntilRunning(batchId)
 
       // Delete the session then verify the YARN app state is KILLED.
       deleteBatch(batchId)
@@ -111,12 +102,11 @@ class BatchIT extends BaseIntegrationTestSuite with BeforeAndAfterAll {
   }
 
   test("killing YARN app should change batch state to dead") {
-    assume(testLibPath != null, "Test lib not uploaded.")
     val output = newOutputPath()
     val batchId = runSpark(classOf[SimpleSparkApp], List(output, "false"), waitForExit = false).id
 
     dumpLogOnFailure(batchId) {
-      val appId = waitBatchForRunning(batchId)
+      val appId = waitUntilRunning(batchId)
 
       // Kill the YARN app and check batch state should be KILLED.
       cluster.yarnClient.killApplication(appId)
@@ -228,16 +218,14 @@ class BatchIT extends BaseIntegrationTestSuite with BeforeAndAfterAll {
     mapper.readValue(response.getResponseBodyAsStream(), classOf[SessionInfo])
   }
 
-  private def waitBatchForRunning(batchId: Int): ApplicationId = {
+  private def waitUntilRunning(batchId: Int): ApplicationId = {
     eventually {
       val batch = getBatchSessionInfo(batchId)
 
       // If batch state transits to any error state, fail test immediately.
       if (batch.state == SessionState.Error().toString() ||
         batch.state == SessionState.Dead().toString()) {
-        throw new TestPendingException {
-          override def getMessage = s"Session shouldn't be in a terminal state: ${batch.state}"
-        }
+        fail(s"Session shouldn't be in a terminal state: ${batch.state}")
       }
 
       assert(batch.state === SessionState.Running().toString())
--- a/integration-test/src/test/scala/com/cloudera/livy/test/BatchIT.scala
+++ b/integration-test/src/test/scala/com/cloudera/livy/test/BatchIT.scala
@@ -24,10 +24,15 @@ import javax.servlet.http.HttpServletResponse._
 
 import scala.concurrent.duration._
 import scala.language.postfixOps
+import scala.util.control.Exception.allCatch
 
 import org.apache.commons.io.IOUtils
 import org.apache.hadoop.fs.Path
+import org.apache.hadoop.yarn.api.records.{ApplicationId, FinalApplicationStatus}
+import org.apache.hadoop.yarn.util.ConverterUtils
+import org.scalatest.BeforeAndAfterAll
 import org.scalatest.concurrent.Eventually._
+import org.scalatest.exceptions.TestPendingException
 
 import com.cloudera.livy.client.common.HttpMessages._
 import com.cloudera.livy.server.batch.CreateBatchRequest
@@ -35,27 +40,38 @@ import com.cloudera.livy.sessions.SessionState
 import com.cloudera.livy.test.apps._
 import com.cloudera.livy.test.framework.BaseIntegrationTestSuite
 
-class BatchIT extends BaseIntegrationTestSuite {
+class BatchIT extends BaseIntegrationTestSuite with BeforeAndAfterAll {
+  implicit val patienceConfig = PatienceConfig(timeout = 2 minutes, interval = 1 second)
 
   private var testLibPath: String = _
 
-  test("upload test lib") {
+  protected override def beforeAll() = {
+    super.beforeAll()
     testLibPath = uploadToHdfs(new File(testLib))
+  }
+
+  test("upload test lib") {
+    // Empty test case to separate time spent on uploading test lib in beforeAll() and
+    // executing actual test cases.
   }
 
   test("submit spark app") {
     assume(testLibPath != null, "Test lib not uploaded.")
-    val output = newOutputPath()
+    val output = newOutputPath
     val result = runSpark(classOf[SimpleSparkApp], args = List(output))
-    assert(result.state === SessionState.Success().toString)
-    assert(cluster.fs.isDirectory(new Path(output)))
+
+    dumpLogOnFailure(result.id) {
+      assert(result.state === SessionState.Success().toString)
+      assert(cluster.fs.isDirectory(new Path(output)))
+    }
   }
 
   test("submit an app that fails") {
     assume(testLibPath != null, "Test lib not uploaded.")
-    val output = newOutputPath()
+    val output = newOutputPath
     val result = runSpark(classOf[FailingApp], args = List(output))
-    assert(result.state === SessionState.Error().toString)
+    // At this point the application has exited. State should be 'dead' instead of 'error'.
+    assert(result.state === SessionState.Dead().toString)
 
     // The file is written to make sure the app actually ran, instead of just failing for
     // some other reason.
@@ -64,7 +80,7 @@ class BatchIT extends BaseIntegrationTestSuite {
 
   pytest("submit a pyspark application") {
     val hdfsPath = uploadResource("pytest.py")
-    val output = newOutputPath()
+    val output = newOutputPath
     val result = runScript(hdfsPath, args = List(output))
     assert(result.state === SessionState.Success().toString)
     assert(cluster.fs.isDirectory(new Path(output)))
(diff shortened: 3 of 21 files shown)


zjffdu:
Using yarn app tags seems a little complicated to me. 
@tc0312 @vanzin  I create another PR #167 for resolving this issue.  In that PR, I use the SparkAppHandle to get the notification of app Id update and monitor the app state in a separated thread.

tc0312:
There's a race condition that leaks application for using SparkListener API.

We described that case in [HA design doc](https://docs.google.com/document/d/1SIK2_ropGGKkirhJBerawmAvDWb9RTG5GWtlG1RafJk)

Copying the paragraph here:

> If Livy server crashed between it persisted the initial state and retrieved the Yarn application ID, it doesn't have a way to know if the job has been accepted by Yarn or not during recovery. To solve this Livy will generate a unique tag (could be a GUID), persist it in the store with Livy batch ID and submit a job with that tag to Yarn. After crash, it can query yarn for status using that if Yarn application ID is not yet persisted in the metadata. If Yarn has an app with that tag, we will get the Yarn application ID and resume to normal recovery. If Yarn does not have it, we know the previous submit failed, and we can mark the job status as error. This tagging and tag-querying capability is only provided by Spark 1.6 so we will have a dependency on that version.

tc0312:
Ready for review.

integration-test/src/main/scala/com/cloudera/livy/test/framework/Cluster.scala
```diff
@@ -48,20 +49,26 @@ trait Cluster {
...
-    configDir().listFiles().foreach { f =>
-      if (f.getName().endsWith(".xml")) {
-        conf.addResource(new Path(f.toURI()))
-      }
-    }
-    conf
-  }
+  lazy val coreSiteConf = loadConf(s"${configDir().getCanonicalPath}/core-site.xml")
+
+  lazy val yarnSiteConf = loadConf(s"${configDir().getCanonicalPath}/yarn-site.xml")
```
vanzin: Is there an issue with creating the yarn client when all configuration is loaded? The previous code does not hardcode config file names, which I prefer.

integration-test/src/test/scala/com/cloudera/livy/test/BatchIT.scala
```diff
@@ -24,38 +24,54 @@ import javax.servlet.http.HttpServletResponse._
...
 import com.cloudera.livy.test.framework.BaseIntegrationTestSuite
 
-class BatchIT extends BaseIntegrationTestSuite {
+class BatchIT extends BaseIntegrationTestSuite with BeforeAndAfterAll {
+  implicit val patienceConfig = PatienceConfig(timeout = 2 minutes, interval = 1 second)
 
   private var testLibPath: String = _
 
-  test("upload test lib") {
+  protected override def beforeAll() = {
```
vanzin: What is this change accomplishing?

integration-test/src/test/scala/com/cloudera/livy/test/BatchIT.scala
```diff
@@ -64,7 +80,7 @@ class BatchIT extends BaseIntegrationTestSuite {
 
   pytest("submit a pyspark application") {
     val hdfsPath = uploadResource("pytest.py")
-    val output = newOutputPath()
+    val output = newOutputPath
```
vanzin: No need to change these things, and I really prefer method calls to have the `()` in them.

integration-test/src/test/scala/com/cloudera/livy/test/BatchIT.scala
```diff
@@ -78,12 +94,79 @@ class BatchIT extends BaseIntegrationTestSuite {
...
+      assert(appId != null, "appId shouldn't be null")
+
+      val appReport = cluster.yarnClient.getApplicationReport(ConverterUtils.toApplicationId(appId))
+      assert(appReport != null, "appReport shouldn't be null")
+
+      appReport.getDiagnostics
+    } getOrElse ""
+  }
+
+  private def newOutputPath: String = {
```
vanzin: This is more of a style convention; but if a method is not a simple property getter, it's good practice to use `()` since it makes it a little clearer that the method actually does things.

integration-test/src/test/scala/com/cloudera/livy/test/BatchIT.scala
```diff
@@ -92,57 +175,94 @@ class BatchIT extends BaseIntegrationTestSuite {
       in.close()
       out.close()
     }
-    hdfsPath.toUri().getPath()
+    hdfsPath.toUri.getPath
```
vanzin: Please avoid these noisy changes that don't really change anything.

server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
```diff
@@ -165,15 +174,30 @@ class LivyServer extends Logging {
...
   /**
+   * Throw an exception if Spark version is not supported.
+   * @param version Spark version
+   */
+  private[server] def testSparkVersion(version: String): Unit = {
+    val versionPattern = """(\d)+\.(\d)+(?:\.\d*)?""".r
+    val (minMajor, minMinor) = (1, 6)
+    val supportedVersion = version match {
+      case versionPattern(major, minor) =>
+        major.toInt > minMajor || major.toInt == minMajor && minor.toInt >= minMinor
```
vanzin: Well, Livy doesn't work with Spark 2, so might as well enforce that too.

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
```diff
@@ -0,0 +1,240 @@
...
+
+import org.apache.hadoop.yarn.api.records.{ApplicationId, ApplicationReport, FinalApplicationStatus, YarnApplicationState}
+import org.apache.hadoop.yarn.client.api.YarnClient
+import org.apache.hadoop.yarn.conf.YarnConfiguration
+
+import com.cloudera.livy.Logging
+import com.cloudera.livy.util.LineBufferedProcess
+
+object SparkYarnApp extends Logging {
+  private class NonDaemonThreadFactory extends ThreadFactory {
```
vanzin: Code similar to this is peppered in a bunch of places... the RSC's `Util` class has a `newDaemonThreadFactory` method, maybe you can just call that. Later we can clean up things and move that method to a more "shared" location.

vanzin:
I'll take a closer look at this tomorrow... I see a lot of style issues (like `while(foo)` and multi-line parameter lists indented differently), it would be nice to keep things in line with the rest of the code.

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
```diff
@@ -0,0 +1,240 @@
...
+    }
+  }
+
+  // Create a non daemon thread pool to run getAppIdFromTagAsync().
+  // getAppIdFromTagAsync() might take a while and should not use the global default thread pool.
+  private implicit val ec = ExecutionContext.fromExecutor(
+    Executors.newCachedThreadPool(new NonDaemonThreadFactory()))
+
+  // It takes at least 5 seconds to see the newly submitted app.
+  private val APP_TAG_TO_ID_TIMEOUT = 30 seconds
```
vanzin: Reading the Spark code (see `Client.scala` in the `yarn` module), first Spark will upload all needed files to the cluster, then it will start the AM, which is where the tags are set. If the upload part takes long, then you might hit this timeout.

So is a timeout the right thing here? I understand that Spark currently doesn't have a mechanism to time out a submission (maybe it should, since it knows more about what's going on?), so maybe this is a necessary evil. But at least it should be configurable.

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
```diff
@@ -0,0 +1,240 @@
...
+  private val KILL_TIMEOUT = APP_TAG_TO_ID_TIMEOUT
+  private val POLL_INTERVAL = 1 second
+
+  // YarnClient is thread safe. Create once, share it across threads.
+  private[this] lazy val yarnClient = YarnClient.createYarnClient()
```
vanzin: You can replace all this (down to the `getYarnClient` method) with:

```
lazy val yarnClient = {
   val client = YarnClient.createYarnClient()
   client.init(new YarnConfiguration())
   client.start()
   client
}
```

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
```diff
@@ -0,0 +1,240 @@
...
+  def getYarnClient(): YarnClient = synchronized {
+    if (!yarnClientCreated) {
+      yarnClient.init(new YarnConfiguration())
+      yarnClient.start()
+      yarnClientCreated = true
+    }
+    yarnClient
+  }
+
+  def getAppIdFromTagAsync(appTag: String): Future[ApplicationId] =
```
vanzin: nit: `Async` is implicit by the method returning a `Future`, so no need to be in the name.

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
```diff
@@ -0,0 +1,240 @@
...
+   *               It will be converted to lower case to match YARN's behaviour.
+   * @return ApplicationId or the failure.
+   */
+  @tailrec
+  private def getAppIdFromTag(
+    appTag: String,
+    deadline: Deadline = APP_TAG_TO_ID_TIMEOUT.fromNow): Try[ApplicationId] = {
+    val appTagLowerCase = appTag.toLowerCase
+
+    // FIXME Should not loop thru all YARN applications but YarnClient doesn't offer an API.
```
vanzin: There are overridden `getApplications`  methods that allow you to filter by app type and state, that should be a little better.

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
```diff
@@ -0,0 +1,240 @@
...
+    // Consider calling rmClient in YarnClient directly.
+    getYarnClient().getApplications().asScala.find(_.getApplicationTags.contains(appTagLowerCase))
+      match {
+        case Some(app) => Success(app.getApplicationId)
+        case None =>
+          if (deadline.isOverdue) {
+            Failure(new Exception(s"No YARN application is tagged with $appTagLowerCase."))
+          } else {
+            blocking { Clock.sleep(POLL_INTERVAL.toMillis) }
```
vanzin: I generally dislike code that blocks threads in a thread pool. That can be replaced by:

- using a `Promise` to signal the completion of the task to the caller (`getAppIdFromTagAsync` can still return a future)
- using a `ScheduledExecutorService` instead of a plain one, and re-scheduling a task instead of sleeping

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
```diff
@@ -0,0 +1,240 @@
...
+    yarnClient: => YarnClient = SparkYarnApp.getYarnClient()) // For unit test.
+  extends SparkApp
+  with Logging {
+  import SparkYarnApp.ec
+
+  private var state: SparkApp.State = SparkApp.State.STARTING
+  private var yarnDiagnostics: IndexedSeq[String] = IndexedSeq.empty[String]
+
+  override def log(): IndexedSeq[String] =
+    (process map (_.inputLines) getOrElse ArrayBuffer.empty[String]) ++ yarnDiagnostics
```
vanzin: Please use periods for method calls.

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
```diff
@@ -0,0 +1,240 @@
...
+
+  override def kill(): Unit = synchronized {
+    if (isRunning) {
+      process foreach (_.destroy())
```
vanzin: `process.foreach(...)`

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
```diff
@@ -0,0 +1,240 @@
...
+
+      try {
+        Await.result(appIdFuture map (yarnClient.killApplication), SparkYarnApp.KILL_TIMEOUT)
```
vanzin: `appIdFuture.map(...)` etc etc. I'll refrain from pointing these out from now on, but please fix all of them.

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
vanzin: Shouldn't you do this after you try to kill the application through YARN? Otherwise you might be making the timeout case more common by killing the spark-submit process mid-submission.

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
```diff
@@ -0,0 +1,240 @@
...
+      case YarnApplicationState.FAILED => SparkApp.State.FAILED
+      case YarnApplicationState.KILLED => SparkApp.State.KILLED
+    }
+  }
+
+  // Exposed for unit test.
+  // TODO Instead of spawning a thread for every session, create a centralized thread and
+  // batch YARN queries.
+  private[utils] val yarnAppMonitorThread = new Thread(s"yarnAppMonitorThread-$this") {
+    override def run() = {
```
vanzin: `: Unit = {`

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
```diff
@@ -0,0 +1,240 @@
...
+        try {
+          // Wait for spark-submit to finish submitting the app to YARN.
+          process map (_.waitFor()) match {
+            case (None | Some(0)) => Await.result(appIdFuture, SparkYarnApp.POLL_INTERVAL)
+            case Some(exitCode) =>
+              throw new Exception(s"spark-submit exited with code $exitCode}.\n" +
+                s"${process.get.inputLines mkString "\n"}")
+          }
+        } catch {
+          case e: TimeoutException => waitForAppId()
```
vanzin: Not sure I understand why you have a timeout if you're calling the method recursively once the `Await` call times out?

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
```diff
@@ -0,0 +1,240 @@
...
+        }
+      }
+
+      try {
+        val appId = waitForAppId()
+
```
vanzin: nit: these blank lines don't really help readability

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
```diff
@@ -0,0 +1,240 @@
...
+        }
+      case YarnApplicationState.FAILED => SparkApp.State.FAILED
+      case YarnApplicationState.KILLED => SparkApp.State.KILLED
+    }
+  }
+
+  // Exposed for unit test.
+  // TODO Instead of spawning a thread for every session, create a centralized thread and
+  // batch YARN queries.
+  private[utils] val yarnAppMonitorThread = new Thread(s"yarnAppMonitorThread-$this") {
```
vanzin: Having a utility method to start a daemon thread would be nice, since you do this in at least a couple of places. Something like:

    def startDaemonThread(name: String)(fn: => Unit): Thread = {
      val thread = new Thread(new Runnable() {
        override def run(): Unit = fn
      }
      ...
    }

server/src/main/scala/com/cloudera/livy/utils/SparkProcApp.scala
```diff
@@ -0,0 +1,65 @@
...
+  override def log(): IndexedSeq[String] = process.inputLines
+
+  private def changeState(newState: SparkApp.State.Value) = {
+    if (state != newState) {
+      listener.foreach(_.stateChanged(state, newState))
+      state = newState
+    }
+  }
+
+  val waitThread = new Thread(new Runnable {
```
vanzin: private?

server/src/main/scala/com/cloudera/livy/utils/SparkProcApp.scala
```diff
@@ -0,0 +1,65 @@
...
+    listener: Option[SparkAppListener])
+  extends SparkApp with Logging{
+
+  private var state = SparkApp.State.STARTING
+
+  override def kill(): Unit = {
+    if (process.isAlive) {
+      process.destroy()
+    }
+  }
```
vanzin: Should you join `waitThread` here?

server/src/main/scala/com/cloudera/livy/utils/SparkApp.scala
```diff
@@ -0,0 +1,92 @@
...
+   *
+   * @param uniqueAppTag A tag that can uniquely identify the application.
+   */
+  def create(
+      uniqueAppTag: String,
+      process: LineBufferedProcess,
+      livyConf: LivyConf,
+      listener: Option[SparkAppListener]): SparkApp = {
+    if (livyConf.isRunningOnYarn) {
+      new SparkYarnApp(getAppIdFromTagAsync(uniqueAppTag), Some(process), listener)
```
vanzin: Can `getAppIdFromTagAsync` be handled internally by `SparkYarnApp`? Then you only need to pass it the unique tag, and that could simplify some code in the other class.

For example, you could use the unique tag when setting the name for `yarnAppMonitorThread`.

At the very least it would cleanup the internal API here a bit.

vanzin: For example, by moving `getAppIdFromTagAsync` to `SparkYarnApp`, you probably don't need the executor that maps tags to appIds at all. Each `yarnAppMonitorThread` can do that itself instead of waiting on a future for the app id.

server/src/main/scala/com/cloudera/livy/utils/SparkApp.scala
```diff
@@ -0,0 +1,92 @@
...
+
+trait SparkAppListener {
+  /** Fired when appId is known, even during recovery. */
+  def appIdKnown(appId: String): Unit = {}
+
+  /** Fired when the app state in the cluster changes. */
+  def stateChanged(oldState: SparkApp.State, newState: SparkApp.State): Unit = {}
+}
+
+// This class looks quite similar to SparkLauncher. Consider adding recovery to SparkLauncher.
```
vanzin: Not sure I understand this comment. Nothing here looks like `SparkLauncher`.

server/src/main/scala/com/cloudera/livy/utils/Clock.scala
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * A lot of Livy code relies on time related functions like Thread.sleep.
+ * To timing effects from unit test, this class is created to mock out time.
+ *
+ * Code in Livy should not call Thread.sleep() directly. It should call this class instead.
+ */
+object Clock {
+  private var _sleep: Long => Unit = Thread.sleep
+
+  def setSleepMethod(sleep: Long => Unit): Unit = _sleep = sleep
```
vanzin: One thing I noticed is that your tests don't reset the sleep method to its original value. Oops.

Instead, a better API would be:

```
def withSleepMethod(sleep: Long => Unit)(fn: => Unit): Unit = {
   // set up sleep method, then
   try { fn } finally { /* reset sleep method */
}
```

server/src/main/scala/com/cloudera/livy/sessions/Session.scala
```diff
@@ -47,26 +47,36 @@ abstract class Session(val id: Int, val owner: String, val livyConf: LivyConf) e
...
+
+  def recordActivity(): Unit = {
+    _lastActivity = System.nanoTime()
+  }
 
   def state: SessionState
 
   def stop(): Future[Unit] = Future {
     try {
+      info(s"Stopping $this...")
```
vanzin: In general I avoid `info` logs for operations triggered by users. These in particular feel more like `debug` to me.

integration-test/src/main/scala/com/cloudera/livy/test/framework/Cluster.scala
tc0312: core-site.xml contains invalid value for yarn.resourcemanager.address.

integration-test/src/test/scala/com/cloudera/livy/test/BatchIT.scala
tc0312: To make BatchIT to upload test lib even when running a selected test case.

integration-test/src/test/scala/com/cloudera/livy/test/BatchIT.scala
tc0312: What you are suggesting is against the recommended Scala style but I don't have a strong opinion on this. Will fix.

server/src/main/scala/com/cloudera/livy/sessions/Session.scala
tc0312: When something goes wrong in production, how do we debug Livy without logging user operations?

server/src/main/scala/com/cloudera/livy/utils/SparkApp.scala
tc0312: For recovery, most sessions have appId persisted in the state store. Those sessions can be constructed directly from appId without the lookup from appTag. This is the motivation of extracting the lookup into a future.

vanzin: I think there's still value in encapsulating that in `SparkYarnApp`. For example, you could handle the use case you mention by having two factory methods (`fromTag` and `fromAppId`) and leave the how of translating a tag to an app id to the internal code.

server/src/main/scala/com/cloudera/livy/sessions/Session.scala
vanzin: There should be an access log for that. I don't think we expose that functionality currently, and it's something we should, but that's how to properly log user operations.

integration-test/src/test/scala/com/cloudera/livy/test/BatchIT.scala
tc0312: Fixed

integration-test/src/test/scala/com/cloudera/livy/test/BatchIT.scala
tc0312: Fixed

server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
tc0312: Added

server/src/main/scala/com/cloudera/livy/sessions/Session.scala
tc0312: Since we don't have access log now and I don't want to make this PR bigger, I will leave these here for now.

server/src/main/scala/com/cloudera/livy/utils/Clock.scala
tc0312: Fixed

server/src/main/scala/com/cloudera/livy/utils/SparkApp.scala
tc0312: Removed

server/src/main/scala/com/cloudera/livy/utils/SparkProcApp.scala
tc0312: Fixed

server/src/main/scala/com/cloudera/livy/utils/SparkProcApp.scala
tc0312: Fixed

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
tc0312: Fixed

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
tc0312: Fixed

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
tc0312: Fixed

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
tc0312: Fixed

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
tc0312: Fixed

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
tc0312: Fixed

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
tc0312: Fixed

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
tc0312: Fixed

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
tc0312: Fixed

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
tc0312: Fixed

server/src/main/scala/com/cloudera/livy/utils/SparkApp.scala
tc0312: Fixed.

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
tc0312: Added it to LivyConf.

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
tc0312: Fixed this by using yarn app monitoring thread.

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
tc0312: Added filtering by app type.  Can't filter by state because Livy needs to show state of completed apps.

integration-test/src/main/scala/com/cloudera/livy/test/framework/BaseIntegrationTestSuite.scala
```diff
@@ -97,15 +97,17 @@ abstract class BaseIntegrationTestSuite extends FunSuite with Matchers with Befo
...
+  protected override def beforeAll() = {
     cluster = Cluster.get()
     httpClient = new AsyncHttpClient()
     livyClient = new LivyRestClient(httpClient, livyEndpoint)
   }
 
   test("initialize test cluster") {
-    // Empty test case to separate time spent on creating cluster in before() and executing actual
-    // test cases.
+    // Empty test case to separate time spent on creating cluster in beforeAll() and
```
vanzin: This isn't actually needed with `beforeAll`, since that code runs as part of test discovery. So if making that change, this "test" can go away.

integration-test/src/test/scala/com/cloudera/livy/test/BatchIT.scala
```diff
@@ -24,38 +24,54 @@ import javax.servlet.http.HttpServletResponse._
...
 
   private var testLibPath: String = _
 
-  test("upload test lib") {
+  protected override def beforeAll() = {
+    super.beforeAll()
     testLibPath = uploadToHdfs(new File(testLib))
   }
 
+  test("upload test lib") {
```
vanzin: Same here ("test" not needed with `beforeAll`).

integration-test/src/test/scala/com/cloudera/livy/test/BatchIT.scala
vanzin: You could have used a `before` block for that:

```
before {
  if (testLibPath == null) {
    // upload
  }
}
```

integration-test/src/test/scala/com/cloudera/livy/test/BatchIT.scala
```diff
@@ -24,38 +24,54 @@ import javax.servlet.http.HttpServletResponse._
...
+    // Empty test case to separate time spent on uploading test lib in beforeAll() and
+    // executing actual test cases.
+  }
+
   test("submit spark app") {
     assume(testLibPath != null, "Test lib not uploaded.")
```
vanzin: You can also remove these `assume` calls.

integration-test/src/test/scala/com/cloudera/livy/test/BatchIT.scala
```diff
@@ -99,50 +182,87 @@ class BatchIT extends BaseIntegrationTestSuite {
...
+
+  private def waitBatchForRunning(batchId: Int): ApplicationId = {
+    eventually {
+      val batch = getBatchSessionInfo(batchId)
 
-    val batchInfo = mapper.readValue(response.getResponseBodyAsStream(), classOf[SessionInfo])
+      // If batch state transits to any error state, fail test immediately.
+      if (batch.state == SessionState.Error().toString() ||
+        batch.state == SessionState.Dead().toString()) {
+        throw new TestPendingException {
```
vanzin: The scaladoc for this exception sounds like it's not really meant to indicate test failure. Why not use a common exception like `IllegalStateException` which, as bonus, has a proper constructor that takes a message?

vanzin: I see this is because of using `eventually` above. Anyway this will affect the reporting for this test, since per the javadoc it doesn't seem like it would show up as a proper test failure.

integration-test/src/test/scala/com/cloudera/livy/test/BatchIT.scala
```diff
@@ -99,50 +182,87 @@ class BatchIT extends BaseIntegrationTestSuite {
...
 
-    assert(response.getStatusCode() === SC_CREATED)
+    withClue(response) {
+      assert(response.getStatusCode === SC_CREATED)
+    }
+
+    mapper.readValue(response.getResponseBodyAsStream(), classOf[SessionInfo])
+  }
+
+  private def waitBatchForRunning(batchId: Int): ApplicationId = {
```
vanzin: The name of this method is super weird. Maybe `waitUntilRunning` or something?

integration-test/src/test/scala/com/cloudera/livy/test/BatchIT.scala
```diff
@@ -99,50 +182,87 @@ class BatchIT extends BaseIntegrationTestSuite {
...
+    eventually {
```
vanzin: Not explicit timeout values?

server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
```diff
@@ -241,15 +250,34 @@ class LivyServer extends Logging {
...
+   */
+  private[server] def testSparkVersion(version: String): Unit = {
+    val versionPattern = """(\d)+\.(\d)+(?:\.\d*)?""".r
+    // This is exclusive. Version which equals to this will be rejected.
+    val (maxMajor, maxMinor) = (2, 0)
+    val (minMajor, minMinor) = (1, 6)
+
+    val supportedVersion = version match {
+      case versionPattern(major, minor) =>
+        (major.toInt > minMajor || major.toInt == minMajor && minor.toInt >= minMinor) &&
```
vanzin: If you want to keep this code generic, I suggest using this:

```
scala> import scala.math.Ordering.Implicits._
import scala.math.Ordering.Implicits._

scala> val vmin = (1, 6)
vmin: (Int, Int) = (1,6)

scala> val vmax = (2, 0)
vmax: (Int, Int) = (2,0)

scala> (1, 6) >= vmin
res0: Boolean = true
```

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
```diff
@@ -0,0 +1,230 @@
...
+    // Consider calling rmClient in YarnClient directly.
+    val appType = Set("SPARK").asJava
+    yarnClient.getApplications(appType).asScala.find(_.getApplicationTags.contains(appTagLowerCase))
+    match {
+      case Some(app) => app.getApplicationId
+      case None =>
+        if (deadline.isOverdue) {
+          throw new Exception(s"No YARN application is tagged with $appTagLowerCase.")
+        } else {
+          Clock.sleep(pollInterval.toMillis)
```
vanzin: This still suffers from the problem of blocking threads while an application is starting. It's preferred to just re-schedule a task instead, which doesn't block any threads.

server/src/test/scala/com/cloudera/livy/utils/SparkYarnAppSpec.scala
```diff
@@ -0,0 +1,197 @@
...
+  private def cleanupThread(t: Thread)(f: => Unit) = {
+    noCatch.andFinally { t.interrupt() } (f)
+  }
+
+  private def mockSleep(ms: Long) = {
+    Thread.`yield`()
+  }
+
+  private def withSleepMethod(mockSleep: Long => Unit)(f: => Unit): Unit = {
+    noCatch.andFinally { Clock.setSleepMethod(Thread.sleep) } {
```
vanzin: What I really mean was to have the `Clock` API expose this method, instead of `setSleepMethod` which is prone to misuse. Also, this Scala construct is pretty ugly, the usual Java construct is much cleaner:

```
try {
  sleepMethod = foo
  f
} finally {
  sleepMethod = Thread.sleep
}
```

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
```diff
@@ -0,0 +1,230 @@
...
+  private def getYarnDiagnostics(appReport: ApplicationReport): IndexedSeq[String] = {
+    Option(appReport.getDiagnostics)
+      .filter(_.nonEmpty)
+      .map[IndexedSeq[String]]("YARN Diagnostics:" +: _.split("\n"))
+      .getOrElse(IndexedSeq.empty)
+  }
+
+  private def isRunning: Boolean = {
+    state != SparkApp.State.FAILED && state != SparkApp.State.FINISHED &&
+    state != SparkApp.State.KILLED
```
vanzin: nit: indent one level

server/src/main/scala/com/cloudera/livy/LivyConf.scala
```diff
@@ -73,6 +73,12 @@ object LivyConf {
   val KINIT_FAIL_THRESHOLD =
     LivyConf.Entry("livy.server.launch.kerberos.kinit_fail_threshold", 5)
 
+  // If Livy can't find the yarn app within this time, consider it lost.
+  val YARN_APP_LOOKUP_TIMEOUT = Entry("livy.server.yarn.app-lookup-timeout", "30s")
+
+  // How often Livy polls YARN to refresh YARN app state.
+  val YARN_POLL_INTERVAL = Entry("livy.server.yarn.poll-interval", "1s")
```
vanzin: `1s` seems pretty aggressive, especially since the heartbeat time for AMs is generally much higher than that (Spark is 5s by default IIRC). Might be good to tune this down, especially since currently you have one thread per session doing the polling.

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
vanzin: BTW you could avoid this altogether if you try to merge this logic with the logic in `yarnAppMonitorThread` (`waitForAppId`). That way, you can use a blocking call as much as you want, since in that code you have one thread per session.

server/src/test/scala/com/cloudera/livy/utils/SparkYarnAppSpec.scala
tc0312: The Clock interface is designed to swap out Thread.sleep globally in Livy so unit test can mock them out. It's a test hook. setSleepMethod() shouldn't be called in production code at all. I don't see any real benefit of resetting the sleep method in production code.

If we wanna make this cleaner, I would argue Clock should be DI-ed into classes, but what you are suggesting defeats the purpose of having a Clock interface.

vanzin: I don't see why it defeats anything. You're adding this clock implementation because *specific* unit tests need it. But, if you do what you did before and forget to clean up, you'll affect *all* unit tests that run after yours.

The interface I'm proposing has exactly the same functionality you seek, but it avoids the problem of forgetting to reset the clock for other tests. So I don't see why it defeats anything.

integration-test/src/main/scala/com/cloudera/livy/test/framework/BaseIntegrationTestSuite.scala
tc0312: Fixed.

integration-test/src/test/scala/com/cloudera/livy/test/BatchIT.scala
tc0312: Fixed.

integration-test/src/test/scala/com/cloudera/livy/test/BatchIT.scala
tc0312: Fixed.

integration-test/src/test/scala/com/cloudera/livy/test/BatchIT.scala
tc0312: Fixed

integration-test/src/test/scala/com/cloudera/livy/test/BatchIT.scala
tc0312: I found that defining
```
implicit val patienceConfig = PatienceConfig(timeout = 2 minutes, interval = 1 second)
```
in the test suite is less verbose than specifying timeout values explicitly everywhere.
I think the implicit timeout value (2 minutes) works well here.

integration-test/src/test/scala/com/cloudera/livy/test/BatchIT.scala
tc0312: Removed. There's no way to break early in scalatest's eventually.

server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
tc0312: Fixed.

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
tc0312: That's the motivation of my change. appId is passed as a call by name parameter.
`
    appId: => ApplicationId,
`

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
tc0312: Fixed

server/src/main/scala/com/cloudera/livy/LivyConf.scala
tc0312: Changed it to 5s to match Spark's default.

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
vanzin: My bad. In that case you don't need the executor context instantiated in this object, do you?

In general I think this code is a little too complicated and hard to follow. I've thought about some ways to make it easier, my current preferred way is:

- have a base class with an abstract `def getAppId(): String`
- have an implementation that has a known app id defined in the constructor (I think you don't need that right now?)
- have a separate implementation where that method's implementation does the equivalent of:

```
override def getAppId(): String = {
  val exitCode = process.waitFor()
  if (exitCode != 0) {
    throw some exception 
  }

  // add code that maps the tag to the app id here
}
```

The base class's `yarnAppMonitorThread` would call `getAppId()` as the first thing it does. You can use a promise for that so that you can keep the current `kill()` implementation. This way the interface becomes cleaner - the non-tag version of the class doesn't need to know about processes and what not, and it's easier to figure out the sequence of actions.

I'm not 100% happy with that, though, so open to suggestions.

tc0312: Mind if you pinpoint which part of the code is hard to follow so I can address that?

I still think my approach is kinda similar to the way you prefer.
Instead of creating 2 subclasses, I am using 2 constructors/factory methods. SparkYarnApp works exactly the same way no matter it's constructed from tag or id. The difference is only in how the object is constructed. Having 2 constructors/factory methods captures the this difference. Also using 2 subclasses is more bloated than using just 2 constructors/factory methods. I need to understand why you think it's hard to follow to address your concern.

vanzin: `SparkYarnApp` has code to wait for a process when a process does not exist in the recovery case. The "process does not exist" is handled when matching for a process result, which is weird.

The actual sequence of operations is "wait for process -> match tag to app id -> monitor app through yarn", and that logic is scattered all over the place. For example, the "wait for process" part is tightly coupled with the code that matches tags to app ids. It's a simple change to decouple those, and that would make the code clearer. No process? Skip any process-related code. Process fails? Throw an exception. Any other state? Ok, *now* do tag matching.

Maybe two classes is overkill, but the current code really needs some rearranging so that it's easier to follow the flow of operations.

vanzin: A more explicit suggestions:

- make SparkYarnApp take both an appId parameter (as `Option[String]`) and a `sessionTag` parameter. This is useful in recovery in case Livy crashes after the app has started but before it can fetch the app id.
- in the code that monitors the app, do something like this:

```
process.foreach { p =>
  // wait for process, throw if error
}

val appId = appIdOption.getOrElse {
  // code that matches session tag to actual app ID
}

// code that monitors app
```

Then everything lives in one class, all code is nicely decoupled (could even be in separate methods for easier unit testing), and the flow of operations is clear.

tc0312: I treat companion object as the static part of the class. Do you consider it as a separate class?

vanzin: I do, in Scala. They're separated enough in the code (you even have to explicitly import things from companion objects) that I don't consider them part of the same entity.

tc0312: I see your point now. Will fix.

integration-test/src/test/scala/com/cloudera/livy/test/BatchIT.scala
```diff
@@ -75,15 +83,80 @@ class BatchIT extends BaseIntegrationTestSuite {
...
     val result = runScript(hdfsPath)
-    assert(result.state === SessionState.Success().toString)
+    assert(result.state === SessionState.Success().toString())
+  }
+
+  test("deleting a session should kill YARN app") {
+    val output = newOutputPath()
+    val batchId = runSpark(classOf[SimpleSparkApp], List(output, "false"), waitForExit = false).id
+
+    dumpLogOnFailure(batchId, cleanup = false) {
```
vanzin: Should `cleanup` really be `false` here?

For example, if `waitUntilRunning` throws, you'll potentially leave the underlying batch session alive, which might affect subsequent tests can cause cascading failures.

The follow up question would be: does it even make sense to have `dumpLogOnFailure` have a `cleanup` parameter? Maybe it should always try to clean up the batch and ignore any errors.

server/src/main/scala/com/cloudera/livy/utils/SparkProcApp.scala
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * Provide a class to control a Spark application using spark-submit.
+ *
+ * @param process The spark-submit process launched the Spark application.
+ */
+class SparkProcApp (
+    process: LineBufferedProcess,
+    listener: Option[SparkAppListener])
+  extends SparkApp with Logging{
```
vanzin: nit: space before `{`

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
```diff
@@ -0,0 +1,228 @@
...
+   *
+   * @param appTag The application tag tagged on the target application.
+   *               If the tag is not unique, it returns the first application it found.
+   *               It will be converted to lower case to match YARN's behaviour.
+   * @return ApplicationId or the failure.
+   */
+  @tailrec
+  private def getAppIdFromTag(
+    appTag: String,
+    pollInterval: Duration = getYarnPollInterval(livyConf),
```
vanzin: Do these need to be parameters? Seems like they could just local variables.

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
```diff
@@ -0,0 +1,228 @@
...
+      // Wait for spark-submit to finish submitting the app to YARN.
+      process.foreach { p =>
+        val exitCode = p.waitFor()
+        if (exitCode != 0) {
+          throw new Exception(s"spark-submit exited with code $exitCode}.\n" +
+            s"${process.get.inputLines.mkString("\n")}")
+        }
+      }
+
+      // If appId is not know, query YARN by appTag to get it.
```
vanzin: "known"

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
```diff
@@ -0,0 +1,228 @@
...
+   * @param appTag The application tag tagged on the target application.
+   *               If the tag is not unique, it returns the first application it found.
+   *               It will be converted to lower case to match YARN's behaviour.
+   * @return ApplicationId or the failure.
+   */
+  @tailrec
+  private def getAppIdFromTag(
+    appTag: String,
+    pollInterval: Duration = getYarnPollInterval(livyConf),
+    deadline: Deadline = getYarnTagToAppIdTimeout(livyConf).fromNow): ApplicationId = {
```
vanzin: I see this one can't be a local variable; but I'd prefer an explicit parameter instead of one with a default value.

vanzin:
A few minor things, otherwise looks good.

integration-test/src/test/scala/com/cloudera/livy/test/BatchIT.scala
tc0312: Fixed.

server/src/main/scala/com/cloudera/livy/utils/SparkProcApp.scala
tc0312: Fixed.

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
tc0312: Fixed.

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
tc0312: Fixed.

vanzin:
I noticed that `SparkYarnAppSpec` does not cover the tag-to-app-id translation code; since that's partially covered by the integration tests, I think it's ok for now. It'd be nice to add unit tests later though, especially to exercise the retry / timeout paths.

Merging.

PR #165, opened 2016-07-11
Added example module

manikandan89:
- Separate module for examples
- Added a simple WordCount example using Scala API

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -43,7 +43,7 @@
 
   <properties>
     <hadoop.version>2.6.0-cdh5.5.0</hadoop.version>
-    <spark.version>1.5.0-cdh5.5.0</spark.version>
+    <spark.version>1.6.2</spark.version>
     <commons-codec.version>1.9</commons-codec.version>
     <httpclient.version>4.5.2</httpclient.version>
     <httpcore.version>4.4.4</httpcore.version>


examples/src/main/scala/com/cloudera/livy/examples/WordCountApp.scala
```diff
@@ -0,0 +1,180 @@
...
+   * 2) Libraries to be added to classpath - livy-api-version-SNAPSHOT.jar,
+   * livy-scala-api-version-SNAPSHOT.jar, livy-client-http-version-SNAPSHOT.jar
+   *
+   * @param args
+   * arg(0) - Livy server url
+   * arg(1) - Output path to save the text read from the stream
+   */
+  def main(args: Array[String]) {
+    if (args.length < 2) {
+      throw new IllegalArgumentException("Number of args is less")
```
vanzin: nit: "Incorrect arguments." or something along those lines. You should also be checking for exactly 2, since that's what you expect.

(In fact you could just omit this block since if you don't have exactly 2 arguments, L165 will throw an exception.)

pom.xml
```diff
@@ -176,7 +176,7 @@
     <module>client-common</module>
     <module>client-http</module>
     <module>core</module>
-    <module>coverage</module>
```
vanzin: You shouldn't remove this.

examples/src/main/scala/com/cloudera/livy/examples/WordCountApp.scala
```diff
@@ -0,0 +1,177 @@
...
+  }
+
+  private def stopClient(shutdownContext: Boolean): Unit = {
+    scalaClient.stop(shutdownContext)
+  }
+
+  /**
+   * Main method of the WordCount App
+   * STEPS FOR EXECUTION
+   * 1) 'livy.file.local-dir-whitelist' config has to be set in livy.conf file for the jars to be
```
vanzin: This is not true, since you're uploading the jars.

examples/src/main/scala/com/cloudera/livy/examples/WordCountApp.scala
```diff
@@ -0,0 +1,177 @@
...
+   * added to the session
+   *
+   * 2) Libraries to be added to classpath - livy-api-version-SNAPSHOT.jar,
```
vanzin: Instead of this comment, better to put the needed dependencies in `compile` scope instead of `provided`. Also, the file names you're referencing don't exist, which is another reason to not add this comment.

examples/src/main/scala/com/cloudera/livy/examples/WordCountApp.scala
```diff
@@ -0,0 +1,177 @@
...
+   * arg(0) - Livy server url
+   * arg(1) - Output path to save the text read from the stream
+   */
+  def main(args: Array[String]) {
+    val Array(uri, outputPath) = args
+    try {
+      init(uri)
+      uploadRelevantJarsForJobExecution()
+      println("Calling processStreamingWordCount")
+      val handle1 = processStreamingWordCount("localhost", 8086, outputPath)
```
vanzin: "localhost" and "8086" should be configurable, no? Also, you need to document what is expected to be running on that host/port so that people really know how to run this. The `processStreamingWordCount` method alludes to reading words from that place, but it doesn't say how; is it an http server? Something else?

PR #166, opened 2016-07-11
Fixed Travis config

tc0312:
- Don't run integration test during Travis's dependency install step.
- Fixed TestSparkClient to not dump tons of debug logs.
- Use Travis cache.


vanzin:
Merging.

PR #167, opened 2016-07-12
LIVY-139. Livy server needs notification when remote session dies

zjffdu:
We see 2 bugs in livy. 
* livy session state will remain as it was when yarn app is killed/failed
* yarn app will be always in ACCEPT state when session creation timeout. (It happens when no sufficient resource capacity)  

This PR will resolve these 2 issues, but only limit to yarn mode.

### DIFF
--- a/rsc/src/main/java/com/cloudera/livy/rsc/ContextLauncher.java
+++ b/rsc/src/main/java/com/cloudera/livy/rsc/ContextLauncher.java
@@ -44,6 +44,7 @@ import static java.nio.file.attribute.PosixFilePermission.*;
 
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.util.concurrent.Promise;
+import org.apache.spark.launcher.SparkAppHandle;
 import org.apache.spark.launcher.SparkLauncher;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -69,9 +70,11 @@ class ContextLauncher {
   private static final String SPARK_ARCHIVES_KEY = "spark.yarn.dist.archives";
   private static final String SPARK_HOME_ENV = "SPARK_HOME";
 
-  static Promise<ContextInfo> create(RSCClientFactory factory, RSCConf conf)
+  static Promise<ContextInfo> create(RSCClientFactory factory,
+                                     RSCConf conf,
+                                     RSCAppListener appListener)
       throws IOException {
-    ContextLauncher launcher = new ContextLauncher(factory, conf);
+    ContextLauncher launcher = new ContextLauncher(factory, conf, appListener);
     return launcher.promise;
   }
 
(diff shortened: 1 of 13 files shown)


zjffdu:
I will create a separated thread for monitoring yarn app state. As SparkAppListener won't get app state notification because I will disconnect with the spark process after the RSC connection is established. If I don't disconnect it, there would be a large amount of useless log which will mess the livy server log.

rsc/src/main/java/com/cloudera/livy/rsc/ContextLauncher.java
```diff
@@ -44,6 +44,7 @@
 
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.util.concurrent.Promise;
+import org.apache.spark.launcher.SparkAppHandle;
```
tc0312: As we discussed before, this will introduce with Livy session recovery/HA.

rsc/src/main/java/com/cloudera/livy/rsc/RSCAppListener.java
```diff
@@ -0,0 +1,61 @@
...
+ */
+
+
+package com.cloudera.livy.rsc;
+
+import org.apache.spark.launcher.SparkAppHandle;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class RSCAppListener implements SparkAppHandle.Listener {
```
tc0312: Since SparkLauncher API is not HA capable. In https://github.com/cloudera/livy/pull/163, I defined a similar listener called SparkAppListener in SparkApp.

I think we can achieve the same goal by extending SparkAppListener.

server/src/main/scala/com/cloudera/livy/sessions/SessionManager.scala
```diff
@@ -41,7 +47,12 @@ class SessionManager[S <: Session](val livyConf: LivyConf) extends Logging {
   private[this] final val sessionTimeout =
     TimeUnit.MILLISECONDS.toNanos(livyConf.getTimeAsMs(SessionManager.SESSION_TIMEOUT))
 
+  private val yarnClient = YarnClient.createYarnClient()
```
tc0312: It takes a while to initialize yarn client. Especially when no YARN is running at the default port on the same machine.
I would not initialize yarn client unless Livy is running on YARN.
I implemented these handling in https://github.com/cloudera/livy/pull/163, so if you can wait a bit, you can get these for free.

tc0312:
vanzin@ posted this comment in JIRA:

> That's outside the scope of what this bug is about though. This bug is very simple: currently, RSC disconnects from Livy server, Livy server does not update session state, which still shows up as "idle" or whatever it was before. Nothing more.
Anything outside of that scope if not what this bug is trying to address. If you want to reproduce this, start a session and then kill it using YARN (not via Livy) and see how the Livy server reacts to that.

vanzin:
Yes, Alex is right, I was going to comment on his recovery work and the reasons for why the code is like it is but he already did. Your changes only cover the non-recovery scenario; you cannot use SparkLauncher with an already-running context (needed for recovery), which is gonna be needed when he's done with his work.

zjffdu:
Thanks @tc0312 @vanzin , close this PR.

PR #168, opened 2016-07-12
LIVY-188. 	change kryo serde method usage

purechoc:
currently kryo serde method usage has bug when SerDe long string include unicode.

may be this is kryo bug.
but, may be this is wrong method usage.

so using byte array instead of ByteBufferInputStream.

### DIFF
--- a/client-common/src/main/java/com/cloudera/livy/client/common/Serializer.java
+++ b/client-common/src/main/java/com/cloudera/livy/client/common/Serializer.java
@@ -22,7 +22,6 @@ import java.io.ByteArrayOutputStream;
 import java.nio.ByteBuffer;
 
 import com.esotericsoftware.kryo.Kryo;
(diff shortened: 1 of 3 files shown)


client-common/src/main/java/com/cloudera/livy/client/common/Serializer.java
```diff
@@ -59,7 +59,9 @@ protected Kryo initialValue() {
   }
 
   public Object deserialize(ByteBuffer data) {
-    Input kryoIn = new Input(new ByteBufferInputStream(data));
```
vanzin: Could you remove the `ByteBufferInputStream` import since it's now unused? LGTM otherwise.

purechoc: thanks to review
removed unused class.

vanzin:
Merging.

PR #169, opened 2016-07-13
LIVY-153. Upgraded Spark from 1.5 to 1.6.2.

tc0312:
Changed default Spark version to 1.6.2 (Apache build).


vanzin:
Updating the Spark version in the build is not necessary to support `spark.yarn.tags`. All you need is to set `SPARK_HOME` to a Spark version that supports the feature.

I'll merge the PR but I'll change the title to remove the misleading text.

PR #170, opened 2016-07-23
LIVY-195. Killing interactive session while spark-submit is running.

tc0312:


rsc/src/test/java/com/cloudera/livy/rsc/TestSparkClient.java
```diff
@@ -346,6 +338,65 @@ void config(Properties conf) {
...
+            throw e;
+          } else {
+            Thread.sleep(1000);
+          }
+        }
+      }
+
+      waitForCalled.countDown();
+    } catch (Exception e) {
+      // JUnit prints not so useful backtraces in test summary reports, and we don't see the
```
vanzin: Are you sure this is not the surefire plugin trying to make output more readable?

It has an option (`-DtrimStackTrace=false`) to disable processing of the stack traces.

rsc/src/test/java/com/cloudera/livy/rsc/TestSparkClient.java
```diff
@@ -346,6 +338,65 @@ void config(Properties conf) {
...
+
+      client = new LivyClientBuilder(false).setURI(new URI("rsc:/"))
+              .setAll(conf)
+              .build();
+
+      client.stop(true);
+
+      // The FutureListener that calls child.kill() is running in a different thread.
+      // To avoid race condition, add a retry block around verify.
+      DateTime deadline = new DateTime().plusSeconds(10);
```
vanzin: Seems like this would be simpler if you used the `CountDownLatch` approach you used for `waitFor`. Just set up an answer for the `destroy` method too, triggering a different latch instance.

rsc/src/test/java/com/cloudera/livy/rsc/TestSparkClient.java
```diff
@@ -346,6 +338,65 @@ void config(Properties conf) {
...
+          waitForCalled.await();
+          return null;
+        }
+      });
+
+      ContextLauncher.mockSparkSubmit = mockSparkSubmit;
+      conf.put(TEST_MOCK_SPARK_SUBMIT.key(), "true");
+
+      client = new LivyClientBuilder(false).setURI(new URI("rsc:/"))
+              .setAll(conf)
```
vanzin: Since I had some comments, might as well. super nit: indent these just one level from previous line.

vanzin:
Looks good, I just think the test can be simplified a little.

rsc/src/test/java/com/cloudera/livy/rsc/TestSparkClient.java
tc0312: I copied and paste it from runBypassTest().

rsc/src/test/java/com/cloudera/livy/rsc/TestSparkClient.java
tc0312: Fixed.

rsc/src/test/java/com/cloudera/livy/rsc/TestSparkClient.java
tc0312: Fixed.

rsc/src/main/java/com/cloudera/livy/rsc/ContextLauncher.java
```diff
@@ -186,7 +195,9 @@ private static ChildProcess startDriver(final RSCConf conf, Promise<?> promise)
 
     final File confFile = writeConfToFile(conf);
 
-    if (conf.getBoolean(CLIENT_IN_PROCESS)) {
+    if (conf.getBoolean(TEST_MOCK_SPARK_SUBMIT)) {
```
vanzin: You could avoid the conf value and just check that `mockSparkSubmit` is `null`.

rsc/src/test/java/com/cloudera/livy/rsc/TestSparkClient.java
```diff
@@ -346,6 +337,48 @@ void config(Properties conf) {
...
+
+      waitForCalled.countDown();
+      destroyCalled.countDown();
+    } catch (Exception e) {
+      // JUnit prints not so useful backtraces in test summary reports, and we don't see the
+      // actual source line of the exception, so print the exception to the logs.
+      LOG.error("Test threw exception.", e);
+      throw e;
+    } finally {
+      stubStream.close();
```
vanzin: Should reset `ContextLauncher.mockSparkSubmit` here, just in case.

rsc/src/test/java/com/cloudera/livy/rsc/TestSparkClient.java
```diff
@@ -346,6 +337,48 @@ void config(Properties conf) {
...
+      when(mockSparkSubmit.getInputStream()).thenReturn(stubStream);
+      when(mockSparkSubmit.getErrorStream()).thenReturn(stubStream);
+
+      // Block waitFor until process.destroy is called.
+      CountDownLatch waitForCalled = new CountDownLatch(1);
+      when(mockSparkSubmit.waitFor()).thenAnswer(new BlockingAnswer<>(waitForCalled, 1));
+
+      // Block waitFor until process.destroy is called.
+      CountDownLatch destroyCalled = new CountDownLatch(1);
+      Mockito.doAnswer(new BlockingAnswer<>(destroyCalled, null)).when(mockSparkSubmit).destroy();
```
vanzin: nit: `Mockito.` unnecessary.

rsc/src/test/java/com/cloudera/livy/rsc/TestSparkClient.java
```diff
@@ -346,6 +337,48 @@ void config(Properties conf) {
...
+      conf.put(TEST_MOCK_SPARK_SUBMIT.key(), "true");
+
+      client = new LivyClientBuilder(false).setURI(new URI("rsc:/"))
+        .setAll(conf)
+        .build();
+
+      client.stop(true);
+
+      waitForCalled.countDown();
+      destroyCalled.countDown();
```
vanzin: I'm confused, what is being tested now?

You want the `client.stop(true)` call to cause `Process.destroy()` to be called, right? So the answer in your mocked `destroy()` method should count down, and here you should instead wait for that to happen with a reasonable timeout.

vanzin:
I think the test is not quite right now.

I also feel like the static mocked process is a little hacky, and that it would be better to test ContextLauncher directly and use a mockito `spy` to override certain behaviors. But that's a more complicated change, so the current hack is ok until I decide it bothers me more than it does now.

rsc/src/main/java/com/cloudera/livy/rsc/RSCConf.java
```diff
@@ -69,7 +69,9 @@
     RPC_SECRET_RANDOM_BITS("secret.bits", 256),
 
     SASL_MECHANISMS("rpc.sasl.mechanisms", "DIGEST-MD5"),
-    SASL_QOP("rpc.sasl.qop", null);
+    SASL_QOP("rpc.sasl.qop", null),
+
+    TEST_MOCK_SPARK_SUBMIT("client.do_not_use.mock_spark_submit", false);
```
vanzin: Not being used anymore.

rsc/src/test/java/com/cloudera/livy/rsc/TestSparkClient.java
```diff
@@ -346,6 +336,60 @@ void config(Properties conf) {
...
+      final CountDownLatch waitForCalled = new CountDownLatch(1);
+      when(mockSparkSubmit.waitFor()).thenAnswer(new Answer<Integer>() {
+        @Override
+        public Integer answer(InvocationOnMock invocation) throws Throwable {
+          waitForCalled.await();
+          return 0;
+        }
+      });
+
+      // Block waitFor until process.destroy is called.
```
vanzin: Comment is wrong. The count down here is to notify the test that destroy was called.

vanzin:
A couple of minor nits, otherwise LGTM.

vanzin:
Merging.

PR #171, opened 2016-07-26
LIVY-198 & LIVY-199

paulmw:


server/src/main/scala/com/cloudera/livy/server/AccessFilter.scala
```diff
@@ -0,0 +1,49 @@
...
+import javax.servlet.http.{HttpServletRequest, HttpServletResponse}
+
+import com.cloudera.livy.LivyConf
+
+class AccessFilter extends Filter {
+
+  val allowedUsers = scala.collection.mutable.Set[String]()
+
+  override def init(filterConfig: FilterConfig): Unit = {
+    val users = filterConfig.getInitParameter(LivyConf.ACCESS_CONTROL_USERS.key)
```
vanzin: I'd suggest doing something similar to the `superusers` method in `LivyConf`. Then you can reuse some of the code for parsing lists too, and avoid the mutable set (not a big deal but I prefer to avoid mutable types unless really needed).

server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
```diff
@@ -122,6 +118,13 @@ class LivyServer extends Logging {
       server.context.addFilter(csrfHolder, "/*", EnumSet.allOf(classOf[DispatcherType]))
     }
 
+    if (livyConf.getBoolean(ACCESS_CONTROL_ENABLED)) {
+      info("Access control is enabled.")
```
vanzin: If `AUTH_TYPE` is not defined in the conf you should probably error out, or at least print a warning, since this won't do anything in that case.

server/src/main/scala/com/cloudera/livy/server/AccessFilter.scala
```diff
@@ -0,0 +1,49 @@
...
+    val httpRequest = request.asInstanceOf[HttpServletRequest]
+    val remoteUser = httpRequest.getRemoteUser
+    if(allowedUsers.contains(remoteUser)) {
+      chain.doFilter(request, response)
+    } else {
+      response.asInstanceOf[HttpServletResponse].sendError(HttpServletResponse.SC_UNAUTHORIZED, "User not authorised to use Livy.")
+    }
+  }
+
+  override def destroy(): Unit = {}
```
vanzin: not needed?

vanzin: nevermind, servlet api (was thinking of the wrong class).

server/src/main/scala/com/cloudera/livy/server/AccessFilter.scala
vanzin: (You can pass `LivyConf` in the constructor, from the `LivyServer` instance.)

vanzin:
Looks alright. You're probably triggering style checks, though; run `mvn verify` to see those.

I think impersonation should default to "on" when kerberos is enabled, but we can change that later.

vanzin:
(Oh, also, please write a descriptive PR title since we use that as the commit message.)

paulmw:
Closing this PR for now to split the issues into separate PRs and base them from separate branches.

PR #172, opened 2016-07-27
LIVY-198. Remove forced impersonation when using Kerberos.

paulmw:
Changing file upload to run as current user if impersonation is disabled.


server/src/main/scala/com/cloudera/livy/sessions/Session.scala
```diff
@@ -31,6 +31,7 @@ import org.apache.hadoop.fs.permission.FsPermission
 import org.apache.hadoop.security.UserGroupInformation
 
 import com.cloudera.livy.{LivyConf, Logging, Utils}
+import com.cloudera.livy.LivyConf._
```
vanzin: I'd just do `LivyConf.IMPERSONATION_ENABLED` since that's the only reference in this file...

vanzin:
Looks good, just a minor style thing.

vanzin:
That's not really what I meant in my feedback (I meant you didn't need the import in the first place), but I'll fix that during merge.

PR #173, opened 2016-07-27
LIVY-199. Add a simple user whitelist capability.

paulmw:
Adding allowed users to LivyConf.
Adding a servlet filter to control who has access.
Validating auth type when enabling access control.

### DIFF
--- a/server/src/main/scala/com/cloudera/livy/LivyConf.scala
+++ b/server/src/main/scala/com/cloudera/livy/LivyConf.scala
@@ -56,7 +56,7 @@ object LivyConf {
   val IMPERSONATION_ENABLED = Entry("livy.impersonation.enabled", false)
   val SUPERUSERS = Entry("livy.superusers", null)
 
-  val ACCESS_CONTROL_ENABLED = LivyConf.Entry("livy.server.access_control.enabled", false)
+  val ACCESS_CONTROL_ENABLED = Entry("livy.server.access_control.enabled", false)
   val ACCESS_CONTROL_USERS = Entry("livy.server.access_control.users", null)
 
   val AUTH_TYPE = Entry("livy.server.auth.type", null)
@@ -114,7 +114,7 @@ class LivyConf(loadDefaults: Boolean) extends ClientConf[LivyConf](null) {
   import LivyConf._
 
   private lazy val _superusers = configToSeq(SUPERUSERS)
-  private lazy val _allowedUsers = configToSeq(ACCESS_CONTROL_USERS)
+  private lazy val _allowedUsers = configToSeq(ACCESS_CONTROL_USERS).toSet
 
   lazy val hadoopConf = new Configuration()
   lazy val localFsWhitelist = configToSeq(LOCAL_FS_WHITELIST).map { path =>
@@ -158,8 +158,8 @@ class LivyConf(loadDefaults: Boolean) extends ClientConf[LivyConf](null) {
   /** Return the list of superusers. */
   def superusers(): Seq[String] = _superusers
 
-  /** Return the list of users allowed to use Livy via SPNEGO. */
-  def allowedUsers(): Seq[String] = _allowedUsers
+  /** Return the set of users allowed to use Livy via SPNEGO. */
+  def allowedUsers(): Set[String] = _allowedUsers
 
   private val configDir: Option[File] = {
     sys.env.get("LIVY_CONF_DIR")
--- a/server/src/main/scala/com/cloudera/livy/LivyConf.scala
+++ b/server/src/main/scala/com/cloudera/livy/LivyConf.scala
@@ -56,6 +56,9 @@ object LivyConf {
   val IMPERSONATION_ENABLED = Entry("livy.impersonation.enabled", false)
   val SUPERUSERS = Entry("livy.superusers", null)
 
+  val ACCESS_CONTROL_ENABLED = LivyConf.Entry("livy.server.access_control.enabled", false)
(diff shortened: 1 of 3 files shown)


server/src/main/scala/com/cloudera/livy/LivyConf.scala
```diff
@@ -56,6 +56,9 @@ object LivyConf {
   val IMPERSONATION_ENABLED = Entry("livy.impersonation.enabled", false)
   val SUPERUSERS = Entry("livy.superusers", null)
 
+  val ACCESS_CONTROL_ENABLED = LivyConf.Entry("livy.server.access_control.enabled", false)
```
vanzin: `LivyConf.` is not necessary.

server/src/main/scala/com/cloudera/livy/server/AccessFilter.scala
```diff
@@ -0,0 +1,48 @@
...
+  private lazy val allowedUsers = livyConf.allowedUsers().toSet
+
+  override def init(filterConfig: FilterConfig): Unit = {}
+
+  override def doFilter(request: ServletRequest,
+                        response: ServletResponse,
+                        chain: FilterChain): Unit = {
+    val httpRequest = request.asInstanceOf[HttpServletRequest]
+    val remoteUser = httpRequest.getRemoteUser
+    if(allowedUsers.contains(remoteUser)) {
```
vanzin: nit: `if (`

server/src/main/scala/com/cloudera/livy/server/AccessFilter.scala
```diff
@@ -0,0 +1,48 @@
...
+package com.cloudera.livy.server
+
+import javax.servlet._
+import javax.servlet.http.{HttpServletRequest, HttpServletResponse}
+
+import com.cloudera.livy.LivyConf
+
+class AccessFilter(livyConf: LivyConf) extends Filter {
+
+  private lazy val allowedUsers = livyConf.allowedUsers().toSet
```
vanzin: If you make the `LivyConf` method return a set, you don't need this field.

server/src/main/scala/com/cloudera/livy/server/AccessFilter.scala
```diff
@@ -0,0 +1,48 @@
...
+  override def doFilter(request: ServletRequest,
+                        response: ServletResponse,
+                        chain: FilterChain): Unit = {
+    val httpRequest = request.asInstanceOf[HttpServletRequest]
+    val remoteUser = httpRequest.getRemoteUser
+    if(allowedUsers.contains(remoteUser)) {
+      chain.doFilter(request, response)
+    } else {
+      val httpServletResponse = response.asInstanceOf[HttpServletResponse]
+        httpServletResponse.sendError(HttpServletResponse.SC_UNAUTHORIZED,
```
vanzin: nit: indentation is wrong

server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
```diff
@@ -122,6 +122,18 @@ class LivyServer extends Logging {
       server.context.addFilter(csrfHolder, "/*", EnumSet.allOf(classOf[DispatcherType]))
     }
 
+    if (livyConf.getBoolean(ACCESS_CONTROL_ENABLED)) {
+      if (livyConf.get(AUTH_TYPE) == KerberosAuthenticationHandler.TYPE) {
```
vanzin: Should be enough to check that it's not `null`, in case we ever add another auth type.

server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
```diff
@@ -122,6 +122,18 @@ class LivyServer extends Logging {
...
+        info("Access control is enabled.")
+        val accessHolder = new FilterHolder(new AccessFilter(livyConf))
+        server.context.addFilter(accessHolder, "/*", EnumSet.allOf(classOf[DispatcherType]))
+      } else {
+        throw new IllegalArgumentException (s"Access control was requested but could " +
```
vanzin: nit: no space before `(`

Style for multiple lines is:

```
throw new IllegalArgumentException(s"This is line 1" +
  "This is line 2" +
  "This is line 3")
```

vanzin:
A few style issues, otherwise looks ok. I'll make a note on LIVY-128 to add tests for this when we add tests for security in general.

server/src/main/scala/com/cloudera/livy/LivyConf.scala
paulmw: Got it. Also true of CSRF_PROTECTION, but I'll leave that out of this commit.

server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
```diff
@@ -150,6 +150,18 @@ class LivyServer extends Logging {
...
     }
 
+    if (livyConf.getBoolean(ACCESS_CONTROL_ENABLED)) {
+      if (livyConf.get(AUTH_TYPE) != null) {
+        info("Access control is enabled.")
+        val accessHolder = new FilterHolder(new AccessFilter(livyConf))
+        server.context.addFilter(accessHolder, "/*", EnumSet.allOf(classOf[DispatcherType]))
+      } else {
+        throw new IllegalArgumentException(s"Access control was requested but could " +
+          s"not be enabled, since auth type is" +
```
vanzin: "since authentication is disabled". I'll fix it during merge.

PR #176, opened 2016-08-09
LIVY-173. Examples module

manikandan89:
Simple WordCount example using Scala-API

### DIFF
--- a/examples/src/main/scala/com/cloudera/livy/examples/WordCountApp.scala
+++ b/examples/src/main/scala/com/cloudera/livy/examples/WordCountApp.scala
@@ -32,23 +32,23 @@ import com.cloudera.livy.scalaapi.{LivyScalaClient, ScalaJobHandle, _}
 
 /**
  *  A WordCount example using Scala-API which reads text from a stream and saves
- *  it as data frames. The word with maximum count is the result
+ *  it as data frames. The word with maximum count is the result.
  */
 object WordCountApp {
 
   var scalaClient: LivyScalaClient = null
 
   /**
-   *  Initializes the scala client with the given url
-   *  @param url The livy server url
+   *  Initializes the Scala client with the given url.
+   *  @param url The Livy server url.
    */
   def init(url: String): Unit = {
     scalaClient = new LivyClientBuilder(false).setURI(new URI(url)).build().asScalaClient
   }
 
   /**
-   *  Uploads the scala-api Jar and the examples Jar from the target directory
-   *  @throws FileNotFoundException If either of scala-api Jar or examples Jar is not found
+   *  Uploads the Scala-API Jar and the examples Jar from the target directory.
+   *  @throws FileNotFoundException If either of Scala-API Jar or examples Jar is not found.
    */
   @throws(classOf[FileNotFoundException])
   def uploadRelevantJarsForJobExecution(): Unit = {
@@ -77,15 +77,15 @@ object WordCountApp {
   }
 
   /**
-   * Submits a spark streaming job to the livy server
+   * Submits a spark streaming job to the livy server.
    *
    * The streaming job reads data from the given host and port. The data read
    * is saved in json format as data frames in the given output path file. If the file is present
    * it appends to it, else creates a new file. For simplicity, the number of streaming batches
-   * are 2 with each batch for 20 seconds. The Timeout of the streaming job is set to 40 seconds
-   * @param host Hostname that Spark Streaming context has to connect for receiving data
-   * @param port Port that Spark Streaming context has to connect for receiving data
-   * @param outputPath Output path to save the processed data read by the Spark Streaming context
+   * are 2 with each batch for 20 seconds. The Timeout of the streaming job is set to 40 seconds.
+   * @param host Hostname that Spark Streaming context has to connect for receiving data.
+   * @param port Port that Spark Streaming context has to connect for receiving data.
+   * @param outputPath Output path to save the processed data read by the Spark Streaming context.
    */
   def processStreamingWordCount(
       host: String,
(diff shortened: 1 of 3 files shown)


examples/src/main/scala/com/cloudera/livy/examples/WordCountApp.scala
```diff
@@ -0,0 +1,243 @@
...
+      val sqlctx = context.sqlctx
+      try {
+        val rdd = sqlctx.read.json(inputPath)
+        rdd.registerTempTable("words")
+        val result = sqlctx.sql("select word, count(word) as word_count from words " +
+          "group by word order by word_count desc limit 1")
+        result.first().toString()
+      } catch {
+        case exception: IOException => {
+          throw new IOException("The input path does not have any data frame", exception)
```
vanzin: nit: the exception could be for a completely different reason than what the message says. better to just let the exception propagate without trying to guess the reason.

examples/src/main/scala/com/cloudera/livy/examples/WordCountApp.scala
```diff
@@ -0,0 +1,243 @@
...
+   * @param args
+   *
+   * REQUIRED ARGUMENTS
+   * arg(0) - Livy server url.
+   * arg(1) - Output path to save the text read from the stream.
+   *
+   * OPTIONAL ARGUMENTS(Default values: arg(2) = 'localhost', arg(3) = 8086)
+   * arg(2) - Socket stream host that spark streaming will listen to.
+   * arg(3) - Socket stream port that spark streaming will listen to.
+   * Optional arguments need to be passed in a key=value way
```
vanzin: If they're passed as `key=value` then they don't need positional markers like `arg(2)`, right?

So just remove the previous lines and say "Remaining arguments are treated as `key=value` pairs. The following keys are recognized:"

examples/src/main/scala/com/cloudera/livy/examples/WordCountApp.scala
```diff
@@ -0,0 +1,243 @@
...
+   *    To pass data to localhost and port 8086 provide the following command
+   *    nc -kl 8086
+   *
+   * 3) The text can be provided as paragraphs as the app will tokenize the data and filter spaces.
+   *
+   * 4) Execute the application jar file with the required and optional arguments either using
+   * mvn or scala.
+   *
+   * Example execution:
+   * scala -cp /pathTo/livy-api-*version*-SNAPSHOT.jar:/pathTo/livy-client-http-*version*
```
vanzin: I'd remove `-SNAPSHOT` from the example since when there's a real release that won't be there.

examples/src/main/scala/com/cloudera/livy/examples/WordCountApp.scala
```diff
@@ -0,0 +1,243 @@
...
+   * -SNAPSHOT.jar:/pathTo/livy-examples-*version*-SNAPSHOT.jar:/pathTo/livy-scala-api-*version*
+   * -SNAPSHOT.jar com.cloudera.livy.examples.WordCountApp http://livy-host:8998 /outputFilePath
+   * host=myhost port=8080
+   */
+  def main(args: Array[String]) {
```
vanzin: nit: `: Unit = {`

examples/src/main/scala/com/cloudera/livy/examples/WordCountApp.scala
```diff
@@ -0,0 +1,243 @@
...
+
```
vanzin: nit: remove this line

examples/src/main/scala/com/cloudera/livy/examples/WordCountApp.scala
```diff
@@ -0,0 +1,243 @@
...
+    var socketStreamHost: String = "localhost"
+    var socketStreamPort: Int = 8086
+    var url = ""
+    var outputFilePath = ""
+    def parseOptionalArg(arg: String): Unit = {
+      val Array(argKey, argValue) = arg.split("=")
+      if (argKey == "host") {
```
vanzin: you can use `argKey match { case "host" => ...` instead, and you should also error out if an unknown argument is provided.

examples/src/main/scala/com/cloudera/livy/examples/WordCountApp.scala
```diff
@@ -0,0 +1,243 @@
...
+        socketStreamHost = argValue
+      } else if (argKey == "port") {
+        socketStreamPort = argValue.toInt
+      }
+    }
+    def parseRequiredArgs(requiredArgs: Array[String]): Unit = {
+      url = requiredArgs(0)
+      outputFilePath = requiredArgs(1)
+    }
+    args.length match {
```
vanzin: Seems like you don't need this match. Just call `parseRequiredArgs` and assert that it contains at least 2 items (`require(args.length >= 2)`), and the process anything after the second index with `parseOptionalArg`.

vanzin:
Could use some simplification in handling the command line options, but the example code itself is fine.

examples/src/main/scala/com/cloudera/livy/examples/WordCountApp.scala
```diff
@@ -0,0 +1,222 @@
...
+    var outputFilePath = ""
+    def parseOptionalArg(arg: String): Unit = {
+      val Array(argKey, argValue) = arg.split("=")
+      argKey match {
+        case "host" => socketStreamHost = argValue
+        case "port" => socketStreamPort = argValue.toInt
+        case _ => throw new IllegalArgumentException("Invalid key for optional arguments")
+      }
+    }
+    def parseRequiredArgs(requiredArgs: Array[String]): Unit = {
```
vanzin: This doesn't need to be a separate function anymore. Just inline the code.

examples/src/main/scala/com/cloudera/livy/examples/WordCountApp.scala
```diff
@@ -0,0 +1,222 @@
...
+import java.io.{File, FileNotFoundException, IOException}
+import java.net.URI
+
+import org.apache.spark.storage.StorageLevel
+import scala.concurrent.Await
+import scala.concurrent.duration._
+import scala.language.postfixOps
+
+import com.cloudera.livy.LivyClientBuilder
+import com.cloudera.livy.scalaapi.{LivyScalaClient, ScalaJobHandle, _}
```
vanzin: Since you're importing `_`, the others are redundant.

vanzin:
Merging.

PR #177, opened 2016-08-10
LIVY-201. Pick up R executable from configuration first

zjffdu:
For now, I can not run sparkr in cluster mode as livy pick up R executable from environment variable DRIVER_R, but spark didn't pass DRIVER_R to AM in yarn-cluster mode (This can be fixed in spark side). And since spark introduce spark.sparkr.r.command & spark.r.command for configuring R executable. Livy should also take this configuration over environment variable DRIVER_R.

### DIFF
--- a/repl/src/main/scala/com/cloudera/livy/repl/SparkRInterpreter.scala
+++ b/repl/src/main/scala/com/cloudera/livy/repl/SparkRInterpreter.scala
@@ -87,7 +87,11 @@ object SparkRInterpreter {
     try {
       // Wait for RBackend initialization to finish
       initialized.tryAcquire(backendTimeout, TimeUnit.SECONDS)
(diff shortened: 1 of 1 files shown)


vanzin:
Spark's docs also list `spark.r.driver.command`, should you also check for that?

zjffdu:
Thanks @vanzin , I added spark.r.driver.command and also change the order. Although the spark doc says `spark.r.driver.command` will be ignored in cluster mode, I still think it is better to make it take priority in livy for sematic reason. I think in spark yarn-client and yarn-cluster just represent 2 execution mode (interactive & batch) for python/R. In livy we can run PythonInterpreter/SparkRInterpreter in yarn-cluster mode, we do still have driver but just only run it on remote node.

vanzin:
Merging.

PR #180, opened 2016-08-18
LIVY-203 Spark version check fails on CDH

Jimvin:
- Fixed regex to also match the Spark version returned by Cloudera CDH
- Reduced the minimum required version of 1.5; Spark 1.6 was not introduced to CDH until 5.7.x

### DIFF
--- a/server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
+++ b/server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
@@ -263,10 +263,10 @@ class LivyServer extends Logging {
    * @param version Spark version
    */
   private[server] def testSparkVersion(version: String): Unit = {
-    val versionPattern = """(\d)+\.(\d)+(?:\.\d*)?""".r
+    val versionPattern = """(\d)+\.(\d)+(?:\.\d*).*?""".r
     // This is exclusive. Version which equals to this will be rejected.
     val maxVersion = (2, 0)
-    val minVersion = (1, 6)
+    val minVersion = (1, 5)
 
     val supportedVersion = version match {
       case versionPattern(major, minor) =>
--- a/server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
+++ b/server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
@@ -251,7 +251,7 @@ class LivyServer extends Logging {
    */
   private[server] def testSparkSubmit(livyConf: LivyConf): Unit = {
     try {
-      testSparkVersion(sparkSubmitVersion(livyConf))
+      testSparkVersion(sparkSubmitVersion(livyConf), livyConf.get(LIVY_SPARK_MASTER))
     } catch {
       case e: IOException =>
         throw new IOException("Failed to run spark-submit executable", e)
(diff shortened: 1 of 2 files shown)


tc0312:
YARN integration does not work with Spark 1.5. Do we really need the latest Livy to support CDH older than 5.7.x?

Jimvin:
YARN integration does work in the build of Spark 1.5 shipped with CDH 5.5. Since there are a lot of people using CDH 5.5 it makes sense to allow it to run.

tc0312:
Are you talking about LIVY-184? The code I added requires "spark.yarn.tags" SparkConf support. Which's added in 1.6. Are you sure Livy works with Spark 1.5?

If working with CDH 5.5 is a requirement, we need to do an additional check in the code to disable YARN integration for Spark 1.5.

Jimvin:
From what I can see the YARN integration does not seem to break Livy for CDH 5.5 and the property spark.yarn.tags will be ignored. Perhaps we can add a deprecation warning for Spark 1.5 if it's not broken but missing YARN integration.

tc0312:
Did you set livy.spark.master = yarn-cluster when you test your test?

Jimvin:
No, you are quite right that YARN integration does not work with Spark 1.5 on CDH 5.5. You get the expected error when Livy tries to find the YARN application based on the tag name:

ERROR SparkYarnApp: Error whiling refreshing YARN state: java.lang.Exception: No YARN application is tagged with livy-session-2-0epbo8qc.

tc0312:
Mind if you add the check I mentioned above?

Jimvin:
I've added a check to disable yarn-master mode for Spark 1.5. Oddly, despite the error message about the YARN tags my test job ran successfully. I'm not certain if not having tag supports actually breaks Livy for Spark 1.5, but the check in is there nonetheless.

tc0312:
Mind if fix the test break?
https://travis-ci.org/cloudera/livy/builds/159856046

Jimvin:
If fixed the tests and added a tests for yarn-master mode being used with Spark < 1.6.

server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
```diff
@@ -275,6 +277,17 @@ class LivyServer extends Logging {
       case _ => false
     }
     require(supportedVersion, s"Unsupported Spark version $version.")
+
+    // Spark >= 1.6 require for Livy yarn-master mode
+    if (sparkMaster == "yarn-master") {
```
vanzin: I'm pretty sure this can be just "yarn" (and have `LIVY_SPARK_DEPLOY_MODE` set to "cluster"). So maybe just check for `startsWith` instead of equality.

Also there's no such thing as "yarn-master".

server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
```diff
@@ -275,6 +277,17 @@ class LivyServer extends Logging {
...
+      val supportedVersion = version match {
+        case versionPattern(major, minor) =>
+          val v = (major.toInt, minor.toInt)
+          v >= minYarnMasterVersion
+        case _ => false
+      }
+      require(supportedVersion, s"YARN master mode unsupported in Spark version $version.")
```
vanzin: I think instead of this it would be better to fall back to using `SparkProcApp` instead of `SparkYarnApp` if version is older than 1.6. But maybe there's something I'm missing that @tc0312 can comment on.

tc0312:
#278 fixed this. Closing.

PR #181, opened 2016-08-20
LIVY-205. sparkr.zip is unzipped on executor side

zjffdu:
This PR is to unzipped sparkr.zip on executor side.

### DIFF
--- a/server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
+++ b/server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
@@ -109,7 +109,7 @@ class InteractiveSession(
       case SparkR() =>
         val sparkRArchive = if (!LivyConf.TEST_MODE) findSparkRArchive() else None
         sparkRArchive.foreach { archive =>
-          builderProperties.put(RSCConf.Entry.SPARKR_PACKAGE.key(), archive)
+          builderProperties.put(RSCConf.Entry.SPARKR_PACKAGE.key(), archive + "#sparkr")
         }
       case _ =>
     }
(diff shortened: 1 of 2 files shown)


zjffdu:
@vanzin Although this PR can make it work, but it needs some configuration change on spark-defaults.conf. Because livy use `R` to launch R script while spark use `Rscript`, is there any reason why livy doesn't use `Rscript` ?

zjffdu:
Never mind, I think I may made a mistake in #177.  R is used in spark as well for interactive mode while RScript is used for batch mode.

zjffdu:
@vanzin  I have a Spark PR (https://github.com/apache/spark/pull/14744) to allow setting sparkr shell command through configuration. And I updated this livy PR based on the spark PR.

zjffdu:
ping @vanzin

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
```diff
@@ -292,7 +292,7 @@ class InteractiveSession(
         val path = Seq(sparkHome, "R", "lib", "sparkr.zip").mkString(File.separator)
         val rArchivesFile = new File(path)
         require(rArchivesFile.exists(), "sparkr.zip not found; cannot run sparkr application.")
-        rArchivesFile.getAbsolutePath()
+        rArchivesFile.getAbsolutePath() + "#sparkr"
```
vanzin: Don't you need to add this suffix when the user sets `SPARKR_PACKAGE` too? Avoids having to force the user to do it manually.

zjffdu: Yeah, right.

zjffdu:
ping @vanzin

vanzin:
Merging.

PR #182, opened 2016-08-24
LIVY-192. Python Client API - Server side changes

manikandan89:
- Server side changes to handle requests from the Python Client API
- Added Integration tests


integration-test/src/test/scala/com/cloudera/livy/test/JobApiIT.scala
```diff
@@ -204,6 +200,62 @@ class JobApiIT extends BaseIntegrationTestSuite with BeforeAndAfterAll with Logg
...
+    val builder = new ProcessBuilder(Seq("python", createPyTestsForPythonAPI().toString).asJava)
+
+    val env = builder.environment()
+    env.put("LIVY_END_POINT", livyEndpoint)
+    env.put("ADD_FILE_URL", addFilePath)
+    env.put("ADD_PYFILE_URL", addPyFilePath)
+    env.put("UPLOAD_FILE_URL", uploadFilePath)
+    env.put("UPLOAD_PYFILE_URL", uploadPyFilePath)
+
+    builder.redirectOutput(new File("src/test/resources/pytest_results.txt"))
```
vanzin: Create this under `sys.props("java.io.tmpdir")` instead of polluting the source directory.

repl/src/main/scala/com/cloudera/livy/repl/BypassPySparkJob.scala
```diff
@@ -0,0 +1,34 @@
...
+
+class BypassPySparkJob(
+    serializedJob: Array[Byte],
+    pysparkJobProcessor: PySparkJobProcessor) extends Job[Array[Byte]] {
+
+  override def call(jc: JobContext): Array[Byte] = {
+    val resultByteArray = pysparkJobProcessor.processBypassJob(serializedJob)
+    val resultString = new String(resultByteArray, "UTF-8")
+    if (resultString.startsWith("Error job")) {
+      throw new Exception(resultString)
```
vanzin: It would be nice to throw a more specific exception, even if it's a silly `PythonJobException` that you declare in this package.

repl/src/main/scala/com/cloudera/livy/repl/BypassPySparkJob.scala
```diff
@@ -0,0 +1,34 @@
...
+import com.cloudera.livy.{Job, JobContext}
+
+class BypassPySparkJob(
+    serializedJob: Array[Byte],
+    pysparkJobProcessor: PySparkJobProcessor) extends Job[Array[Byte]] {
+
+  override def call(jc: JobContext): Array[Byte] = {
+    val resultByteArray = pysparkJobProcessor.processBypassJob(serializedJob)
+    val resultString = new String(resultByteArray, "UTF-8")
+    if (resultString.startsWith("Error job")) {
```
vanzin: Since you're returning this "Error job" string to the client, it would be nice to phrase it better. e.g. "Client job error: <exception string>"

repl/src/main/scala/com/cloudera/livy/repl/BypassPySparkJob.scala
```diff
@@ -0,0 +1,34 @@
...
+
+import com.cloudera.livy.{Job, JobContext}
+
+class BypassPySparkJob(
+    serializedJob: Array[Byte],
+    pysparkJobProcessor: PySparkJobProcessor) extends Job[Array[Byte]] {
+
+  override def call(jc: JobContext): Array[Byte] = {
+    val resultByteArray = pysparkJobProcessor.processBypassJob(serializedJob)
+    val resultString = new String(resultByteArray, "UTF-8")
```
vanzin: nit: use `StandardCharsets.UTF_8` instead of the string.

integration-test/src/test/resources/test_python_api.py
```diff
@@ -0,0 +1,223 @@
...
+        poll_request_uri = livy_end_point + "/sessions/" + str(session_id) + \
+                           "/jobs/" + str(job_id)
+        poll_header = {'X-Requested-By': 'livy'}
+        poll_response = requests.request('GET', poll_request_uri, headers=poll_header)
+        poll_time *= 2
+
+    assert poll_response.json()['id'] == job_id
+    assert poll_response.status_code == httplib.OK
+    if not is_error_job:
+        result = poll_response.json()['result']
```
vanzin: This code made it a little hard to debug things when the call actually failed unexpectedly.

Can you add an assert that in this case the error is none? That way if there's an error it will show up in the log file.

vanzin:
Ran tests locally with mini cluster, real spark, and real cluster, all passing. Merging.

PR #185, opened 2016-08-31
LIVY-204: statement history kept in memory indefinitely

paulmw:
Storing statements in a better structure that allows for slicing and removal.
Adding support for statement deletion.

### DIFF
--- a/server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
+++ b/server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
@@ -26,6 +26,7 @@ import java.util.concurrent.atomic.AtomicLong
 
 import scala.annotation.tailrec
 import scala.collection.JavaConverters._
+import scala.collection.immutable.TreeMap
 import scala.collection.mutable
 import scala.concurrent.{Future, _}
 import scala.util.{Failure, Success, Try}
@@ -169,7 +170,7 @@ class InteractiveSession(
 
 
   private[this] var _executedStatements = 0
-  private[this] var _statements = IndexedSeq[Statement]()
+  private[this] var _statements = TreeMap[Int, Statement]()
 
   override def logLines(): IndexedSeq[String] = IndexedSeq()
 
@@ -181,7 +182,7 @@ class InteractiveSession(
     transition(SessionState.Dead())
   }
 
-  def statements: IndexedSeq[Statement] = _statements
+  def statements: TreeMap[Int, Statement] = _statements
 
   def interrupt(): Future[Unit] = {
     stop()
@@ -200,9 +201,21 @@ class InteractiveSession(
     val statement = new Statement(_executedStatements, content, future)
 
     _executedStatements += 1
-    _statements = _statements :+ statement
+    addStatement(statement)
 
     statement
+  }
+
+  def addStatement(statement: Statement): Unit = {
+    _statements.synchronized {
+      _statements = _statements + (statement.id -> statement)
+    }
+  }
+
+  def removeStatement(statementId: Int): Unit = {
+    _statements.synchronized {
+      _statements = _statements - statementId
+    }
   }
 
   def runJob(job: Array[Byte]): Long = {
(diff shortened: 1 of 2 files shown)


vanzin:
Error is legitimate:

```
[31m- should setup and tear down an interactive session *** FAILED *** (718 milliseconds)[0m
[31m  500 did not equal 200 Unexpected response status: 500 != 200 ("java.lang.ClassCastException: scala.collection.immutable.Vector cannot be cast to scala.collection.immutable.TreeMap") (BaseJsonServletSpec.scala:104)[0m
```

tc0312:
It seems this approach requires the user to delete statement explicitly.
If we add a feature to keep just last N statements, users don't have to worry about deleting old statements. They will just disappear. Do you think this works better for your use case?
The N value could be a per session property or a LivyConf.

paulmw:
Yeah. The approach does require explicit deletes. I was initially thinking about only keeping the last N statements, but that then requires users to consider what N should be. In practice, there are actually a number of strategies you could use, any of which might be valid:

- Keep all statements (the current approach)
- Keep the last N statements
- Keep the last N completed statements
- Keep all statements which have yet to have been retrieved
- Keep statements for the last T seconds

So, classic strategy pattern... a StatementManager trait with multiple implementations, selected through config. 

I backed off of this idea a bit since it was probably over complicating things a bit.. happy to revisit that.

vanzin:
I think there's an underlying design flaw in the interactive session design here. The Livy server shouldn't be storing any of this stuff; it should be stored in the user's session (a.k.a. the Spark application), and the Livy server should just be proxying requests for that data. That way, there's no need to set any limits in the Livy server since sessions would not really cause memory usage to balloon.

That's a more complicated change, though. So in the short term the server needs some kind of limit for how much memory a session is allowed to use. The easiest way is to set an upper limit on the number of statement results stored for each session. Note that "upper limit" means sessions cannot be configured to go above that limit - this should be a server-side configuration that admins only can control.

Explicit deletes are fine if the above is respected.

tc0312:
> I think there's an underlying design flaw in the interactive session design here. The Livy server shouldn't be storing any of this stuff; it should be stored in the user's session (a.k.a. the Spark application), and the Livy server should just be proxying requests for that data.

The old interactive code does that. I've just discovered the new ReplDriver doesn't. The changes mentioned by @vanzin (storing statement results in ReplDriver, livy-server serving as just a proxy) are required for interactive session HA. Without these changes, all statements results are lost when livy-server restarts.

I'm thinking of a way to implement these and will tackle this issue as a part of my work.

tc0312:
The change mentioned by @vanzin has a side effect. If the repl driver crashes, all statement results will be lost because they are stored just in livy-repl but not livy-server. Is this acceptable?

jerryshao:
I think it would be better to store in the driver side, which will be easy for HA reconstruction of metadata.

@tc0312 I think you mixed two things:

1. Livy server HA.
2. Livy repl/driver HA. Driver HA actually is not so important and Spark itself doesn't address this issue except Spark Streaming.

For Livy server HA, it would be better to store the data into repl driver which can be recovered later easily.

And for Livy repl HA, actually it is not so important compared to Livy server HA (SPOF) and the impact is not so huge. Also it might not be so complicated to handle if we have Write Ahead Log like mechanism to replay all the executed statements.

tc0312:
@jerryshao Thanks for your input. Glad to see at least 3 of us agree with this approach. 
Just to elaborate a bit, I'm not trying to address livy-repl HA. My main concern is in current Livy, statement results are cached in livy-server and are available after livy-repl crashes. With my new code, this's no longer true, statement results will be lost. This's a change in behaviour and I worry that it might break some use cases I haven't thought of.

jerryshao:
I think it makes sense to lose these statements after repl is gone, since repl is already gone, so keeping these statements is not so necessary.

tc0312:
I've implemented the proxy change purposed by @vanzin. Will send a PR soon.

tc0312:
#214 will fix this

PR #186, opened 2016-09-01
LIVY-162. Livy to support Spark 2.0 + Scala 2.11 cross build

jerryshao:
This PR adds support to build against Spark 2.0 and Scala 2.11, what this PR addressed:

1. Support building against different Spark version with profile like -Pspark-1.6, -Pspark-2.0.
2. Support building against different version of Scala through profile: -Pscala-2.10 -Pscala-2.11.
3. Refactor SparkInterpret to support different Scala version + Spark version.

What this PR doesn't address:

1. Batch API to support SparkSession. Currently batch API still uses old API for `SparkContext`, `SQLContext` and `HiveContext`.

This PR is tested with following scenarios:

1. Spark 1.6 + Scala 2.10
2. Spark 1.6 + Scala 2.11
3. Spark 2.0 + Scala 2.10
3. Spark 2.0 + Scala 2.11

Please suggest and comment, any feedback is greatly appreciated.


jerryshao:
@vanzin @tc0312 @zjffdu please suggest, thanks a lot.

zjffdu:
@jerryshao Can I run both spark1.6/spark2 in one livy build ? Or do I need to make different build with different profile ?

jerryshao:
>Can I run both spark1.6/spark2 in one livy build 

What's the meaning of one livy build? Are you referring one Livy server to server different Sparks, in this scenario currently I think it is not supported.

If you're referring one build with two deploys, one for Spark 1.6 and another for Spark 2.0. If Spark 1.6 and 2.0 are both using same scala version, then I guess it could be supported (though I haven't tried). But if Spark 1.6 uses Scala 2.10 while Spark 2.0 uses 2.11, then I think it cannot be supported currently.

zjffdu:
I mean running different spark version in one livy server.  Since spark interpreter are in another JVM, so it should be possible to do that.  Let's hear more feedback from others.

client-common/pom.xml
```diff
@@ -20,19 +20,19 @@
...
   <parent>
     <groupId>com.cloudera.livy</groupId>
-    <artifactId>livy-main</artifactId>
+    <artifactId>livy-main_2.10</artifactId>
     <version>0.3.0-SNAPSHOT</version>
   </parent>
 
   <groupId>com.cloudera.livy</groupId>
-  <artifactId>livy-client-common</artifactId>
+  <artifactId>livy-client-common_2.10</artifactId>
```
jerryshao: Do we need to add Scala binary version suffix like `2.10` to this module that is purely Java code? I checked Spark code, it adds suffix for such modules like `launcher`, `network-comm`.

tc0312:
@zjffdu is right about livy-server can be Scala version agnostic.
livy-server doesn't really interact with Spark besides calling spark-submit.
For batch, livy-server calls spark-submit directly.
For interactive, livy-server uses SparkLauncher, a library in Spark, but it's a Java library.
I think we can build livy-server with just Scala 2.11.

jerryshao:
But livy-repl and livy-core is Scala version dependent and according to my test it needs to be uploaded to Spark runtime, then I think it is not Scala version agnostic.

tc0312:
@jerryshao Yup your understanding is correct.

jerryshao:
@tc0312 , would you please elaborate this, sorry I could not catch your point.

>I think for now, we can build everything besides livy-repl (and may be livy-rsc? I'm not sure) with just Scala 2.10 and Spark 1.6. And upgrade to Scala 2.11 and Spark 2.0 as a separate task.

vanzin:
I'm with Alex here. We don't need to builld Livy-the-server against Scala 2.11 or Spark 2 to support those; that's why, when I filed bugs for this, I filed separate ones.

It's ok to support a single Spark version *at runtime* to get the ball rolling (supporting multiple versions at runtime is actually what LIVY-106 is about, btw).

Basically, there's a bunch of different tasks here that are independent and shouldn't be mixed together. They are, in order of what I think is important:

- building the repl code against different versions of scala / spark
- supporting multiple versions of Spark at runtime
- building Livy-the-server with Spark 2.11

It would be nice if this patch covers *just* the first (which is LIVY-162).

In the distant past I've played with how to support building against different Scala versions in the same project, and I came up with this:
https://github.com/vanzin/multi-scala

I think that pattern can help a lot with achieving that first goal.

jerryshao:
Thanks @vanzin for your comments. I will pick the first one for this PR.

* Only change the repl related codes to support different versions of Scala / Spark.
*  Livy-server and other parts be isolated to Scala version (currently Scala 2.10 is enough).

One more thing livy-scala-api should also be built against different scala versions, but I think it could be addressed in other JIRA, since this jar is not packed into livy-server runtime package.

jerryshao:
@vanzin @tc0312 ,
I refactored this PR a bit to only cover "building the repl code against different versions of scala / spark". 

1. For now only `livy-core` and `livy-repl` can be built against different versions of Spark, other modules are still bound to scala 2.10.
2. Still use the profile to support different versions of Spark, like "-Pspark-1.6 or -Pspark-2.0".

Please review and suggest, thanks a lot.

scala/pom.xml
```diff
@@ -0,0 +1,116 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
tc0312: I'm not a Maven expert. Would you mind explaining the purpose of this pom.xml? Thanks.

jerryshao: It is a fake module only used to support different scala version in one build, you could refer to @vanzin 's comment and here https://github.com/vanzin/multi-scala .

tc0312:
Thanks, Jerry.
If my understanding is correct, repl 2.11 is not being integration tested. IT is using repl 2.10.

Would you mind changing it to run IT for both repl? 
Off the top of my head I think we can use Travis matrix to set an environment variable and read it in livy-server. May be you have some better idea.

jerryshao:
Let me see if I could enable cross integration test

jerryshao:
@tc0312 it may not be easy to change to run IT for both repl, it may requires not only scripts or pom changs, some hard coded jar paths should also be updated. I think we'd better minimize the first  step and get this in, then we could do the following things like you mentioned above, what do you think?

tc0312:
Mind if you fix the build break?

jerryshao:
Done, mind reviewing again?

tc0312:
Mind if you create a JIRA for configuring Travis to run integration test with Scala 2.11?

jerryshao:
Sure, I can handle this following tasks in other JIRAs.

repl/src/main/scala/com/cloudera/livy/repl/SparkContextInitializer.scala
```diff
@@ -0,0 +1,123 @@
...
+      }
+    } else {
+      spark = builder.getClass.getMethod("getOrCreate").invoke(builder)
+      info("Created Spark session")
+    }
+
+    sparkContext = spark.getClass.getMethod("sparkContext").invoke(spark)
+      .asInstanceOf[SparkContext]
+
+    bind("spark", spark.getClass.getCanonicalName, spark, List("""@transient"""))
```
jerryshao: Do we really need such hints? Since spark-shell is facing to the end-user, while interpreter in Livy isn't.

repl/scala-2.11/src/main/scala/com/cloudera/livy/repl/SparkInterpreter.scala
```diff
@@ -0,0 +1,182 @@
...
+
+/**
+ * Scala 2.11 version of SparkInterpreter
+ */
+class SparkInterpreter(conf: SparkConf)
+  extends AbstractSparkInterpreter with SparkContextInitializer {
+
+  protected var sparkContext: SparkContext = _
+  private var sparkILoop: SparkILoop = _
+  private var classServer: Object = _
```
jerryshao: Actually in Spark2.0, this is not an Http server, just an plain Netty file server, so it is better not to declare as a http server.

repl/scala-2.11/src/main/scala/com/cloudera/livy/repl/SparkInterpreter.scala
```diff
@@ -0,0 +1,182 @@
...
+
+  private def startHttpServer(outputDir: File): (Object, String) = {
+    try {
+      val securityManager = {
+        val constructor = Class.forName("org.apache.spark.SecurityManager")
+          .getConstructor(classOf[SparkConf])
+        constructor.setAccessible(true)
+        constructor.newInstance(conf).asInstanceOf[Object]
+      }
+      val classServerConstructor = Class.forName("org.apache.spark.HttpServer")
```
jerryshao: I see, let me change it.

jerryshao:
@vazin would you please help to review the current implementation, it would be great to get your feedback, thanks a lot and deeply appreciated.

api/pom.xml
```diff
@@ -15,7 +15,9 @@
   ~ See the License for the specific language governing permissions and
   ~ limitations under the License.
 -->
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0"
```
vanzin: It would be nice to avoid these unnecessary changes.

assembly/pom.xml
```diff
@@ -89,4 +78,41 @@
...
+      <id>targz</id>
+      <properties>
+        <assembly.format>tar.gz</assembly.format>
+      </properties>
+    </profile>
+
+    <profile>
+      <id>scala-2.10</id>
+      <activation>
+        <activeByDefault>true</activeByDefault>
```
vanzin: Do not use `activeByDefault`. It does not do what you think it does.

If you want to choose between 2.10 and 2.11 using profiles, you need to do what Spark does and condition it on a property. (`-Dscala-2.10` in the case of Spark, since 2.11 is the default there)

client-common/pom.xml
```diff
@@ -16,7 +16,9 @@
   ~ See the License for the specific language governing permissions and
   ~ limitations under the License.
 -->
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0"
```
vanzin: Noise... could you revert all these unnecessary changes?

examples/pom.xml
```diff
@@ -19,58 +19,58 @@
...
-        <skip>true</skip>
-        <skipDeploy>true</skipDeploy>
-    </properties>
+  <properties>
+    <skipTests>true</skipTests>
+    <skip>true</skip>
+    <skipDeploy>true</skipDeploy>
+  </properties>
 
-    <dependencies>
```
vanzin: It seems you're just changing indentation here... or are you changing something else too?

Please avoid noisy changes. I know the indentation is not consistent in all pom files but it would be better to fix that separately, otherwise it's hard to know what's changing.

pom.xml
```diff
@@ -963,4 +971,30 @@
     </snapshotRepository>
   </distributionManagement>
 
+  <profiles>
+    <!-- Spark version profile -->
+    <profile>
+      <id>spark-1.6</id>
+      <activation>
+        <activeByDefault>true</activeByDefault>
```
vanzin: Don't use this. See previous comment.

repl/pom.xml
```diff
@@ -19,177 +19,176 @@
...
-                        <spark.master>local</spark.master>
-                    </systemProperties>
-                </configuration>
-            </plugin>
-
-        </plugins>
-    </build>
+  </parent>
+
+  <artifactId>livy-repl-parent</artifactId>
```
vanzin: Again, it's hard to see what's changing here, if anything. Please revert the indentation changes so it's possible to see whether there are any real changes.

repl/scala-2.11/src/main/scala/com/cloudera/livy/repl/SparkInterpreter.scala
```diff
@@ -0,0 +1,183 @@
...
+
+  protected def bind(name: String, tpe: String, value: Object, modifier: List[String]): Unit = {
+    sparkILoop.beQuietDuring {
+      sparkILoop.bind(name, tpe, value, modifier)
+    }
+  }
+
+  private def createTempDir(rootDir: String): File = {
+    try {
+      val method = Class.forName("org.apache.spark.util.Util")
```
vanzin: Why call internal Spark code here?

Just use `java.nio.file.Files.createTempDirectory`.

repl/scala-2.11/src/main/scala/com/cloudera/livy/repl/SparkInterpreter.scala
```diff
@@ -0,0 +1,183 @@
...
+      case NonFatal(e) =>
+        val dir = new File(rootDir, s"spark-${UUID.randomUUID().toString}")
+        dir.mkdir()
+        dir
+    }
+  }
+
+  private def startHttpServer(outputDir: File): Option[(Object, String)] = {
+    try {
+      val httpServerClass = Class.forName("org.apche.spark.HttpServer")
```
vanzin: You have a typo there.

Also, isn't SparkEnv responsible for starting this? Why does it have to be done manually in the 2.11 repl? (See how the 2.10 repl doesn't do this.)

repl/src/main/scala/com/cloudera/livy/repl/SparkContextInitializer.scala
```diff
@@ -0,0 +1,127 @@
...
+    if (isSparkSessionPresent()) {
+      spark2CreateContext(conf)
+    } else {
+      spark1CreateContext(conf)
+    }
+  }
+
+  private def spark1CreateContext(conf: SparkConf): Unit = {
+    val sparkClz = classOf[SparkContext]
+    sparkContext = sparkClz.getMethod("getOrCreate", classOf[SparkConf]).invoke(null, conf)
```
vanzin: Do you need reflection for `SparkContext` methods? Aren't they binary compatible between 1.6 and 2.0? The signature in the source code is the same.

server/pom.xml
```diff
@@ -19,227 +19,227 @@
...
-                        </goals>
-                    </execution>
-                </executions>
-            </plugin>
-
-        </plugins>
-    </build>
+  </parent>
+
+  <artifactId>livy-server</artifactId>
```
vanzin: Same. Can't see what's changing here.

assembly/pom.xml
```diff
@@ -89,4 +78,41 @@
...
+        </dependency>
+      </dependencies>
+    </profile>
+
+    <profile>
+      <id>scala-2.11</id>
+      <dependencies>
+        <dependency>
+          <groupId>${project.groupId}</groupId>
+          <artifactId>livy-repl_2.11</artifactId>
```
vanzin: Is this right?

If I understand correctly, all modules except repl and core are still built with 2.10. Meaning the server code itself is built with Scala 2.10, so the assembly should really depend on the 2.10 artifact.

jerryshao: I see, thanks.

repl/scala-2.11/src/main/scala/com/cloudera/livy/repl/SparkInterpreter.scala
jerryshao: We don't require this in Spark 2.0, but in 1.6 it still needed, please refer to [here](https://github.com/apache/spark/blob/branch-1.6/repl/scala-2.11/src/main/scala/org/apache/spark/repl/Main.scala#L39).

repl/src/main/scala/com/cloudera/livy/repl/SparkContextInitializer.scala
jerryshao: I see, I will change it back.

jerryshao:
@vanzin thanks a lot for your reviewing, I revert most of the unnecessary changes, would you please review again, thanks again.

Regarding test, currently repl unit test can be run with Scala 2.10 and 2.11. But for other tests they're still depending on repl-2.10, so only Scala 2.10 can be verified, this requires code and pom file changes, also some additional scripts may be required, I think we could do it separately.

assembly/pom.xml
jerryshao: I use profile `scala-2.11` and `scala-2.10` to differentiate the repl jar packaged in assembly. If `scala-2.11` is specified, the assembly module will depend on `repl_2.11` and package the dependencies into folder `repl-jars`. If not specified it will always pick `repl_2.10` jars into folder `repl-jars`.

repl/scala-2.11/src/main/scala/com/cloudera/livy/repl/SparkInterpreter.scala
vanzin: Ah. Unfortunate that both repls work so much differently...

assembly/pom.xml
vanzin: That's fine. I just re-read the assembly pom and it doesn't really use the dependencies - they're there for build order only, but packaging is not done based on them. So mixing versions should work fine here.

core/scala-2.11/pom.xml
```diff
@@ -0,0 +1,37 @@
...
+
+  <parent>
+    <groupId>com.cloudera.livy</groupId>
+    <artifactId>livy-core-parent</artifactId>
+    <version>0.3.0-SNAPSHOT</version>
+    <relativePath>../pom.xml</relativePath>
+  </parent>
+
+  <properties>
+    <scala.version>2.11.8</scala.version>
```
vanzin: minor, but could you declare `scala-2.11.version` in the root pom, and just use that variable in these files? Then there's only one place to update the 2.11 version when needed.

pom.xml
```diff
@@ -963,4 +971,37 @@
...
+      <properties>
+        <spark.version>1.6.2</spark.version>
+        <py4j.version>0.9</py4j.version>
+        <netty.version>4.0.23.Final</netty.version>
+        <json4s.version>3.2.10</json4s.version>
+      </properties>
+    </profile>
+
+    <profile>
+      <id>spark-2.0</id>
```
vanzin: You didn't answer my question about the need for this profile. What is it used for? It seems all the 1.6 vs. 2.0 differences are being handled via reflection.

repl/scala-2.11/src/main/scala/com/cloudera/livy/repl/SparkInterpreter.scala
```diff
@@ -0,0 +1,169 @@
...
+      val httpServerConstructor = httpServerClass
+        .getConstructor(classOf[SparkConf],
+          classOf[File],
+          Class.forName("org.apache.spark.SecurityManager"),
+          classOf[Int],
+          classOf[String])
+      httpServerConstructor.setAccessible(true)
+      // Create Http Server
+      val server = httpServerConstructor
+        .newInstance(conf, outputDir, securityManager, new Integer(0), "HTTP server")
```
vanzin: Minor, but should probably note somewhere that `spark.fileserver.port` will be ignored.

server/src/main/scala/com/cloudera/livy/utils/LivySparkUtils.scala
```diff
@@ -0,0 +1,115 @@
...
+
+import scala.math.Ordering.Implicits._
+
+import com.cloudera.livy.LivyConf
+import com.cloudera.livy.util.LineBufferedProcess
+
+object LivySparkUtils {
+
+  /**
+   * Sets the spark-submit path if it's not configured in the LivyConf
```
vanzin: Comment doesn't match method.

server/src/main/scala/com/cloudera/livy/utils/LivySparkUtils.scala
```diff
@@ -0,0 +1,115 @@
...
+    }
+  }
+
+  /**
+   * Throw an exception if Spark version is not supported.
+   * @param version Spark version
+   */
+  def testSparkVersion(version: String): Unit = {
+    // This is exclusive. Version which equals to this will be rejected.
+    val maxVersion = (2, 1)
```
vanzin: With the new code this check will probably get more annoying than useful... perhaps we should switch to just check for the min version. But it's ok for now.

pom.xml
jerryshao: Because Spark 2.0 upgraded some dependencies like Netty, json4s-jackson which are not compatible with Spark 1.6.

1. Currently we will upload netty jar as livy-rsc dependencies to Spark, and this will introduce conflict since we use the old version.
2. json4s-jackson's `render` API signature is changed in Spark's version.

That's why I use separated profiles for Spark 1.6 and 2.0. Yes this can also be avoided either with shading or some other ways. 

@vanzin what is your suggestion? Thanks a lot.

repl/scala-2.11/src/main/scala/com/cloudera/livy/repl/SparkInterpreter.scala
jerryshao: Should it be `spark.replClassServer.port`?

vanzin: Yes. Too many configs. I also see you filed a Spark bug that the setting is not used in the first place, so the point is moot anyway.

server/src/main/scala/com/cloudera/livy/utils/LivySparkUtils.scala
jerryshao: done.

core/scala-2.11/pom.xml
jerryshao: done.

jerryshao:
@vanzin , is there any missing piece I should address?

pom.xml
vanzin: The json4s issue is unfortunate and one of the reasons why I'd like to get rid of that dependency... but Livy shouldn't be overriding Spark's version of netty. It may upload a netty jar with the app, but that should be in the app's class loader, not on the system class loader, so Spark should still use its own.

Anyway, it's fine for now until the issue of supporting multiple versions simultaneously is sorted out.

repl/scala-2.11/src/main/scala/com/cloudera/livy/repl/SparkInterpreter.scala
```diff
@@ -0,0 +1,170 @@
...
+      }
+
+      createSparkContext(conf)
+    }
+
+    sparkContext
+  }
+
+  override def close(): Unit = synchronized {
+    if (sparkHttpServer != null) {
```
vanzin: The start order is http server -> iloop -> context, so the shutdown order should be context -> iloop -> http server.

vanzin:
Alex also had some comments you still haven't addressed.

pom.xml
jerryshao: according to my test, Spark will pick Livy's Netty instead of its own and throw exception because of API incompatibility.

repl/src/main/scala/com/cloudera/livy/repl/SparkContextInitializer.scala
```diff
@@ -0,0 +1,123 @@
...
+    val builderMethod = sparkClz.getMethod("builder")
+    val builder = builderMethod.invoke(sparkObj)
+    builder.getClass.getMethod("config", classOf[SparkConf]).invoke(builder, conf)
+
+    var spark: Object = null
+    if (conf.get("spark.sql.catalogImplementation", "hive").toLowerCase == "hive") {
+      if (sparkClz.getMethod("hiveClassesArePresent").invoke(sparkObj).asInstanceOf[Boolean]) {
+        builder.getClass.getMethod("enableHiveSupport").invoke(builder)
+        spark = builder.getClass.getMethod("getOrCreate").invoke(builder)
+        info("Created Spark session with Hive support")
```
jerryshao: Fixed.

repl/src/main/scala/com/cloudera/livy/repl/SparkContextInitializer.scala
```diff
@@ -0,0 +1,123 @@
...
+      } else {
+        builder.getClass.getMethod("config", classOf[String], classOf[String])
+          .invoke(builder, "spark.sql.catalogImplementation", "in-memory")
+        spark = builder.getClass.getMethod("getOrCreate").invoke(builder)
+        info("Created Spark session")
```
jerryshao: Fixed.

repl/src/main/scala/com/cloudera/livy/repl/SparkContextInitializer.scala
```diff
@@ -0,0 +1,123 @@
...
+      }
+    } else {
+      spark = builder.getClass.getMethod("getOrCreate").invoke(builder)
+      info("Created Spark session")
```
jerryshao: Fixed.

repl/src/main/scala/com/cloudera/livy/repl/SparkContextInitializer.scala
```diff
@@ -0,0 +1,123 @@
...
+    bind("spark", spark.getClass.getCanonicalName, spark, List("""@transient"""))
+    bind("sc", "org.apache.spark.SparkContext", sparkContext, List("""@transient"""))
+
+    execute("import org.apache.spark.SparkContext._")
+    execute("import spark.implicits._")
+    execute("import spark.sql")
+    execute("import org.apache.spark.sql.functions._")
+  }
+
+  private def isSparkSessionPresent: Boolean = {
```
jerryshao: Fixed.

repl/src/main/scala/com/cloudera/livy/repl/AbstractSparkInterpreter.scala
```diff
@@ -379,3 +258,10 @@ class SparkInterpreter(conf: SparkConf) extends Interpreter with Logging {
     output
   }
 }
+
+object AbstractSparkInterpreter {
```
jerryshao: Done.

repl/scala-2.10/src/main/scala/com/cloudera/livy/repl/SparkInterpreter.scala
```diff
@@ -0,0 +1,137 @@
...
+
+  protected def bind(name: String, tpe: String, value: Object, modifier: List[String]): Unit = {
+    sparkIMain.beQuietDuring {
+      sparkIMain.bind(name, tpe, value, modifier)
+    }
+  }
+
+  override def close(): Unit = synchronized {
+    if (sparkContext != null) {
+      sparkContext.stop()
```
jerryshao: Done.

repl/scala-2.11/src/main/scala/com/cloudera/livy/repl/SparkInterpreter.scala
jerryshao: Done.

repl/scala-2.11/src/main/scala/com/cloudera/livy/repl/SparkInterpreter.scala
```diff
@@ -0,0 +1,182 @@
...
+
+  private def createTempDir(rootDir: String): File = {
+    try {
+      val method = Class.forName("org.apache.spark.util.Util")
+        .getMethod("createTempDir", classOf[String], classOf[String])
+      method.setAccessible(true)
+      method.invoke(null, rootDir, "spark").asInstanceOf[File]
+    } catch {
+      case NonFatal(e) =>
+        val file = new File(rootDir, s"spark-${UUID.randomUUID().toString}")
```
jerryshao: Removed this method.

repl/scala-2.11/src/main/scala/com/cloudera/livy/repl/SparkInterpreter.scala
```diff
@@ -0,0 +1,182 @@
...
+        file.mkdir()
+        file
+    }
+  }
+
+  private def startHttpServer(outputDir: File): (Object, String) = {
```
jerryshao: Done.

repl/scala-2.11/src/main/scala/com/cloudera/livy/repl/SparkInterpreter.scala
jerryshao: Fixed.

jerryshao:
Hi @vanzin , I checked again, all the comments should have been addressed, would you please please review again, thanks a lot and appreciate your time.

vanzin:
Looks OK. Probably could use some cleanup but we can do that later. Merging.

PR #187, opened 2016-09-05
LIVY-209. Fix getPathInfo return null issue

jerryshao:
getPathInfo may return null if there's no path info, this will lead to wrong `location` data in the returning headers, so here fix this.

To produce this issue:

```python
>>> import json, pprint, requests, textwrap
>>> host = 'http://localhost:8998'
>>> data = {'kind': 'spark'}
>>> headers = {'Content-Type': 'application/json'}
>>> r = requests.post(host + '/sessions', data=json.dumps(data), headers=headers)
>>> r.json()
{u'kind': u'spark', u'log': [], u'proxyUser': None, u'state': u'starting', u'appId': None, u'owner': None, u'id': 0}
>>> r.headers['location']
'null/sessions/0'
```

the right value should be:

```python
>>> r.headers['location']
'/sessions/0'
```

### DIFF
--- a/server/src/main/scala/com/cloudera/livy/server/SessionServlet.scala
+++ b/server/src/main/scala/com/cloudera/livy/server/SessionServlet.scala
@@ -128,7 +128,7 @@ abstract class SessionServlet[S <: Session](livyConf: LivyConf)
   }
 
   private def getRequestPathInfo(request: HttpServletRequest): String = {
-    if (request.getPathInfo != "/") {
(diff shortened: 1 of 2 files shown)


tc0312:
Thanks for the fix. Mind if you add some unit test?

jerryshao:
Sure, I will try to add a unit test.

tc0312:
LGTM

PR #188, opened 2016-09-06
LIVY-161. Avoid exposing livy internal configurations to Spark

jerryshao:
Livy internal configurations will be existed as system properties in Spark driver process, and these can be fetched through various ways in Spark's runtime, so here propose to remove these Livy only system properties after being set to `RSCConf`.


tc0312:
Mind if you test this with IntegrationTest?
I think we can test this easily and cheaply by piggybacking this InteractiveIT.basic interactive session.
We can execute the following code in the interpreter and the result should be false.
```scala
sc.getConf().contains(<SECRET CONF KEY NAME>)
```

jerryshao:
Unit test added.

tc0312:
Your test is even better. LGTM

PR #189, opened 2016-09-07
LIVY-208. Exposing YARN driver log url and Spark UI url for YARN sessions.

tc0312:
They are exposed as:
- appInfo.driverLogUrl
- appInfo.sparkUiUrl

### DIFF
--- a/integration-test/src/test/scala/com/cloudera/livy/test/BatchIT.scala
+++ b/integration-test/src/test/scala/com/cloudera/livy/test/BatchIT.scala
@@ -40,6 +40,7 @@ import com.cloudera.livy.test.apps._
 import com.cloudera.livy.test.framework.BaseIntegrationTestSuite
 
 class BatchIT extends BaseIntegrationTestSuite with BeforeAndAfterAll {
+  import com.cloudera.livy.utils.AppInfo._
   implicit val patienceConfig = PatienceConfig(timeout = 2 minutes, interval = 1 second)
 
   private var testLibPath: String = _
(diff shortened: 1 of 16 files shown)


tc0312:
@vanzin @jerryshao @zjffdu Mind if you guys take a look?

server/src/main/scala/com/cloudera/livy/utils/SparkApp.scala
```diff
@@ -18,15 +18,25 @@
...
 package com.cloudera.livy.utils
 
+import scala.collection.JavaConverters._
+
 import com.cloudera.livy.LivyConf
 import com.cloudera.livy.util.LineBufferedProcess
 
+case class AppInfo(var driverLogUrl: Option[String] = None, var sparkUiUrl: Option[String] = None) {
+  def asJavaMap: java.util.Map[String, String]
+    = Map("driverLogUrl" -> driverLogUrl.orNull, "sparkUiUrl" -> sparkUiUrl.orNull).asJava
```
jerryshao: nit: move `=` to the above line.

jerryshao: Can we define these keys "driverLogUrl", "sparkUiUrl" in a common place. I found that lots of places use them, and it easily introduces the typo.

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
```diff
@@ -209,16 +211,43 @@ class SparkYarnApp private[utils] (
...
+          changeState(mapYarnState(
+            appReport.getApplicationId,
+            appReport.getYarnApplicationState,
+            appReport.getFinalApplicationStatus))
+
+          val latestAppInfo = {
+            val attempt =
+              yarnClient.getApplicationAttemptReport(appReport.getCurrentApplicationAttemptId)
+            val driverLogUrl =
+              Try(yarnClient.getContainerReport(attempt.getAMContainerId).getLogUrl)
```
jerryshao: Does it mean driver should be run on AM, here the driver log is actually AM log? So in that case, how can we handle yarn client mode, or other cluster manager like Standalone or Mesos?

integration-test/src/test/scala/com/cloudera/livy/test/BatchIT.scala
```diff
@@ -56,6 +56,10 @@ class BatchIT extends BaseIntegrationTestSuite with BeforeAndAfterAll {
     dumpLogOnFailure(result.id) {
       assert(result.state === SessionState.Success().toString())
       assert(cluster.fs.isDirectory(new Path(output)))
+      result.appInfo should contain key ("driverLogUrl")
+      result.appInfo should contain key ("sparkUiUrl")
+      result.appInfo.get("driverLogUrl")
```
zjffdu: duplicated test with `result.appInfo should contain key ("driverLogUrl")`

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
tc0312: You are right and Livy officially supports yarn-cluster mode and local mode only.

tc0312:
Fixed. Please review.

jerryshao:
LGTM.

PR #193, opened 2016-09-19
Minor. Fixed bug in SparkApp that merges user YARN tags and Livy YARN tags incorrectly.

zjffdu:
The user defined tags will be merged into generated tags. The key is not used correctly.


zjffdu:
@tc0312 Please help review.

tc0312:
My bad. Will merge.

PR #194, opened 2016-09-19
LIVY-215. Support canceling of statement

zjffdu:
StatementId is unique in the scope of per session. So I just use the statementId as the jobGroupId.  
I think trait `Interpreter` is not public for extension of other interpreter. So I change `Interpreter` directly to make it support canceling of statement. 

Here's the steps for trying this features.
- curl -X POST --data '{"kind": "spark"}' -H "Content-Type: application/json" localhost:8998/sessions
- curl   http://localhost:8998/sessions/0/statements -X POST -H 'Content-Type: application/json' -d '{"code":"sc.parallelize(0 to 100).map(e=>{Thread.sleep(10000); e}).sum()"}' 
- curl -X POST http://localhost:8998/sessions/0/statements/0/cancel
- curl   http://localhost:8998/sessions/0/statements/0

### DIFF
--- a/server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
+++ b/server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
@@ -430,22 +430,29 @@ class InteractiveSession(
     stop()
   }
 
-  def executeStatement(content: ExecuteRequest): Statement = {
+  def executeStatement(content: ExecuteRequest): Statement = synchronized {
     ensureRunning()
     _state = SessionState.Busy()
     recordActivity()
 
+    val statementId = _executedStatements
     val future = Future {
-      val id = client.get.submitReplCode(content.code)
+      val id = client.get.submitReplCode(content.code, statementId.toString)
       waitForStatement(id)
     }
 
-    val statement = new Statement(_executedStatements, content, future)
+    val statement = new Statement(statementId, content, future)
 
     _executedStatements += 1
     _statements = _statements :+ statement
 
     statement
+  }
+
+  def cancelStatement(statementId: String): Unit = {
+    ensureRunning()
+    recordActivity()
+    client.get.cancelReplCode(statementId)
   }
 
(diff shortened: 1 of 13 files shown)


server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
```diff
@@ -215,24 +215,31 @@ class InteractiveSession(
     stop()
   }
 
-  def executeStatement(content: ExecuteRequest): Statement = {
+  def executeStatement(content: ExecuteRequest): Statement = synchronized {
```
zjffdu: Just to make sure statementId is unique in one session, so I add synchronized here.

zjffdu:
@vanzin @tc0312 @jerryshao Mind to review ?

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
tc0312: I'm a change for session recovery that slashes with this change a lot.
https://github.com/hdinsight/livy/commit/6c7f15b58d0fd6844021930b6d6087aaee8b0a9a

tc0312:
I've a change for session recovery that conflicts with this a lot.
https://github.com/hdinsight/livy/commit/6c7f15b58d0fd6844021930b6d6087aaee8b0a9a

zjffdu:
@tc0312 I see you also have a cancelJob method, but don't see where it is used. Have you already implemented the cancellation ? And do you have a plan to merge that commit to upstream ?

tc0312:
@zjffdu 
No. I didn't touch that method. It's already there. I didn't change much besides proxying statement result requests to repl. I'm going to upstream it. But I'm upstreaming them one by one:
1. LIVY-211 Session recovery for batch sessions
2. LIVY-212 Session recovery for interactive sessions
3. LIVY-213 Statement result recovery for interactive sessions

tc0312:
Jerry has taken over this PR.

PR #197, opened 2016-10-05
LIVY-220. Workaround SPARK-17512 for Spark 2.

OferE:
The livy code takes the pyFiles variable and passes it as a --conf "spark.submit.pyFiles=<whatever>" as a command line to spark.
The spark.submit.pyFiles was only introduced in spark 2.0.X and is not availablw in 1.6.X
I tried to run it with spark 2.0.1 but it turns out that spark 2.0.X has also a bug(!!!) that ignores this option lolol....
This change make it being passed as --py-files instead...

This is critical for pyspark users as it solves a bug that makes livy useless for pyspark developers.

### DIFF
--- a/server/src/main/scala/com/cloudera/livy/utils/SparkProcessBuilder.scala
+++ b/server/src/main/scala/com/cloudera/livy/utils/SparkProcessBuilder.scala
@@ -178,8 +178,13 @@ class SparkProcessBuilder(livyConf: LivyConf) extends Logging {
     addOpt("--name", _name)
     addOpt("--class", _className)
     _conf.foreach { case (key, value) =>
-      arguments += "--conf"
-      arguments += f"$key=$value"
+      if (key == "spark.submit.pyFiles") {
+         arguments += "--py-files"
+         arguments += f"$value"
+      } else {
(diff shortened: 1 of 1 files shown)


jerryshao:
Should we also fix this issue in `InteractiveSession`? I think this issue also exists in interactive session.

OferE:
I didn't check it in InteractiveSession, since it was less critical to me.
It's easier to workaround this in interactiveSession since one can use addPyFile interactively.
(https://spark.apache.org/docs/0.8.1/api/pyspark/pyspark.context.SparkContext-class.html)

I agree that it is a good idea to check this in intereactiveSession and fix it there if needed too.

tc0312:
For interactive session, I don't think this applies because livy-repl is starting its own py4j interpreter. It's not started by org.apache.spark.deploy.PythonRunner.

This's a missing feature in interactive session.

ksakellis:
@OferE We are preparing to donate this project to the ASF and just to be extra careful, can you please fill out and send us the CCLA or ICLA (or both) as described here: https://github.com/cloudera/livy/wiki/Contributing-to-Livy

PR #199, opened 2016-10-10
Added an override for hadoop.scope to Hadoop dependencies in pom.xml.

tc0312:
- Default to compile.
- Users who want to exclude Hadoop jars in their Livy build can set it to provided.

### DIFF
--- a/server/pom.xml
+++ b/server/pom.xml
@@ -97,6 +97,7 @@
         <dependency>
             <groupId>org.apache.hadoop</groupId>
             <artifactId>hadoop-auth</artifactId>
+            <scope>${hadoop.scope}</scope>
         </dependency>
 
         <dependency>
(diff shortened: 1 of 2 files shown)


tc0312:
Please let me know if there are any concerns.

server/pom.xml
```diff
@@ -97,6 +97,7 @@
         <dependency>
             <groupId>org.apache.hadoop</groupId>
             <artifactId>hadoop-auth</artifactId>
+            <scope>${hadoop.scope}</scope>
```
jerryshao: Is changing `hadoop-auth` scope only enough, what about others like `hadoop-common`, `hadoo-client`, `hadoop-yarn-xxx`?

tc0312: They are transitive dependencies of hadoop-client.
I changed hadoop-client's scope below.

tc0312: Does that address your concern?

tc0312:
Will merge soon.

server/pom.xml
jerryshao: I see, thanks.

PR #200, opened 2016-10-10
LIVY-221. Default Livy name is overwriting spark.app.name.

linchan-ms:
Removed default value for name parameter for batch.


tc0312:
This will remove the default session name "Livy", which's acceptable to me.
If there's no objection, I will merge.

jerryshao:
LGTM.

PR #201, opened 2016-10-11
LIVY-219. Support scala-2.11 integration test for Livy IT

jerryshao:
This PR adds support for cross verification of scala-2.11 and 2.10 repl in integration test. To achieve this this PR has several changes:

1. Isolate Livy server classpath to avoid mixing into IT, so Livy server will always use scala 2.10.
2. Repl jars will be picked according to scala version.
3. Scala 2.11 integration test will be enabled by "-Dscala-2.11-it".
4. Changed travis script to run 2.11 IT speparately.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,7 @@ before_script:
 
 script:
   - mvn verify -e
+  - mvn verify -Dscala-2.11-it -e -DskipTests
 
 after_success:
   - bash <(curl -s https://codecov.io/bash)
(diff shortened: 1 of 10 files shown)


tc0312:
Figuring out a way to compile integration-test just with Scala-2.10.

tc0312:
I prototyped another apporach that builds integration-test just in Scala 2.10. So we don't have to duplicate things from livy-server.

The main idea is add 2 maven packages with Spark and its dependencies with a Scala version and write the maven classpath to a file. Integration test will then pick up the correct classpath from the file.
https://github.com/hdinsight/livy/tree/scala-2.11-it

tc0312:
See #211

PR #202, opened 2016-10-12
LIVY-225. Fix two issues running integration test with Spark 2.0

jerryshao:
1. Create a dummy jar to let Spark yarn#client correctly launch the application.
2. Since `SQLContext` is deprecated in Spark 2.0, so additional information will be printed out when creating, so here change the regex pattern.

### DIFF
--- a/integration-test/src/test/scala/com/cloudera/livy/test/InteractiveIT.scala
+++ b/integration-test/src/test/scala/com/cloudera/livy/test/InteractiveIT.scala
@@ -48,9 +48,9 @@ class InteractiveIT extends BaseIntegrationTestSuite with BeforeAndAfter {
     dumpLogOnFailure(sessionId) {
       matchResult("1+1", "res0: Int = 2")
       matchResult("""sc.getConf.get("spark.executor.instances")""", "res1: String = 1")
-      matchResult("sqlContext", startsWith("res2: org.apache.spark.sql.hive.HiveContext"))
       matchResult("val sql = new org.apache.spark.sql.SQLContext(sc)",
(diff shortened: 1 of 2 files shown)


integration-test/src/test/scala/com/cloudera/livy/test/InteractiveIT.scala
```diff
@@ -48,9 +48,9 @@ class InteractiveIT extends BaseIntegrationTestSuite with BeforeAndAfter {
     dumpLogOnFailure(sessionId) {
       matchResult("1+1", "res0: Int = 2")
       matchResult("""sc.getConf.get("spark.executor.instances")""", "res1: String = 1")
-      matchResult("sqlContext", startsWith("res2: org.apache.spark.sql.hive.HiveContext"))
       matchResult("val sql = new org.apache.spark.sql.SQLContext(sc)",
-        startsWith("sql: org.apache.spark.sql.SQLContext = org.apache.spark.sql.SQLContext"))
+        ".*" + Pattern.quote(
```
tc0312: I think we should test with SparkSession here if the test's running in Spark 2.0. If this's not trivial, feel free to create a JIRA and do it later.

zjffdu:
Do we need to update triavis.yml to run it again spark 2.0 ?

jerryshao:
Yes, I think we should enable Spark2 integration test also, but I think we should scala 2.11 integration test #211 first, then enable Spark2 test.

tc0312:
LGTM. Will resolve conflict on your behalf and merge.

PR #204, opened 2016-10-12
LIVY-224. Datanucleus jars are not detected for spark 2.x

zjffdu:
This PR is to detect datanucleus jars for spark 2.x where datanucleus is in different locations compared to spark 1.x.

Manually tested.

### DIFF
--- a/server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
+++ b/server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
@@ -342,7 +342,7 @@ class InteractiveSession(
     }
   }
 
-  private def datanucleusJars(livyConf: LivyConf, sparkMajorVersion:Int): Seq[String] = {
+  private def datanucleusJars(livyConf: LivyConf, sparkMajorVersion: Int): Seq[String] = {
     if (sys.env.getOrElse("LIVY_INTEGRATION_TEST", "false").toBoolean) {
       // datanucleus jars has already been in classpath in integration test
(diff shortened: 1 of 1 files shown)


zjffdu:
@jerryshao @tc0312 @vanzin  Please help review.

jerryshao:
LGTM.

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
```diff
@@ -342,18 +342,29 @@ class InteractiveSession(
...
+            new File(sparkHome, "lib")
+          } else {
+            new File(sparkHome, "lib_managed/jars")
+          }
+        case 2 =>
+          if (new File(sparkHome, "RELEASE").isFile) {
+            new File(sparkHome, "jars")
+          } else if (new File(sparkHome, "assembly/target/scala-2.11/jars").isDirectory) {
+            new File(sparkHome, "assembly/target/scala-2.11/jars")
+          } else {
```
tc0312: Can we do a check here?
} else if (new File(sparkHome, "assembly/target/scala-2.10/jars").isDirectory) {

If all if clauses are false, can we print a warning?

zjffdu: warning message will be printed here if datanucleus jars are not found, https://github.com/zjffdu/livy/blob/LIVY-224/server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala#L375

PR #205, opened 2016-10-14
LIVY-227. Include both repl-2.10 and repl-2.11 bundles in Livy assembly.

jerryshao:
With this change, Livy's assembly package will have two REPL bundles: `repl_2.10-jars` and `repl_2.11-jars`. Livy will pick the right jars based on:

1. If Scala version can be detected automatically by Livy.
2. If User specify Scala major version in `CreateInteractiveSession`.
3. If Livy could figure out default Scala version for Spark.

Please review and suggest, thanks a lot.

CC @tc0312 .

### DIFF
--- a/server/src/main/scala/com/cloudera/livy/LivyConf.scala
+++ b/server/src/main/scala/com/cloudera/livy/LivyConf.scala
@@ -41,6 +41,8 @@ object LivyConf {
   val SPARK_HOME = Entry("livy.server.spark-home", null)
   val LIVY_SPARK_MASTER = Entry("livy.spark.master", "local")
   val LIVY_SPARK_DEPLOY_MODE = Entry("livy.spark.deployMode", null)
+  val LIVY_SPARK_SCALA_VERSION = Entry("livy.spark.scalaVersion", null)
+  val LIVY_SPARK_VERSION = Entry("livy.spark.version", null)
 
   val SESSION_STAGING_DIR = Entry("livy.session.staging-dir", null)
   val FILE_UPLOAD_MAX_SIZE = Entry("livy.file.upload.max.size", 100L * 1024 * 1024)
--- a/server/src/main/scala/com/cloudera/livy/LivyConf.scala
+++ b/server/src/main/scala/com/cloudera/livy/LivyConf.scala
@@ -41,6 +41,11 @@ object LivyConf {
   val SPARK_HOME = Entry("livy.server.spark-home", null)
   val LIVY_SPARK_MASTER = Entry("livy.spark.master", "local")
   val LIVY_SPARK_DEPLOY_MODE = Entry("livy.spark.deployMode", null)
+
+  // Two configurations to specify Spark and related Scala version. These are internal
+  // configurations will be set by LivyServer and used in session creation. It is not required to
+  // set usually unless running with unofficial Spark + Scala versions
+  // (like Spark 2.0 + Scala 2.10, Spark 1.6 + Scala 2.11)
   val LIVY_SPARK_SCALA_VERSION = Entry("livy.spark.scalaVersion", null)
   val LIVY_SPARK_VERSION = Entry("livy.spark.version", null)
 
--- a/server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
+++ b/server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
@@ -64,7 +64,16 @@ class LivyServer extends Logging {
 
     // Make sure the `spark-submit` program exists, otherwise much of livy won't work.
     testSparkHome(livyConf)
-    testSparkSubmit(livyConf)
+
+    // Test spark-submit and get Spark Scala version accordingly.
+    val (sparkVersion, scalaVersion) = sparkSubmitVersion(livyConf)
+    testSparkVersion(sparkVersion)
+    // If Spark and Scala version is not set, set into livy configuration, this will be used by
+    // session creation.
+    livyConf.setIfMissing(LIVY_SPARK_VERSION.key,
+      formatSparkVersion(sparkVersion).productIterator.mkString("."))
+    livyConf.setIfMissing(LIVY_SPARK_SCALA_VERSION.key,
+      formatScalaVersion(scalaVersion, formatSparkVersion(sparkVersion)))
 
     testRecovery(livyConf)
 
--- a/server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
+++ b/server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
@@ -68,12 +68,31 @@ class LivyServer extends Logging {
     // Test spark-submit and get Spark Scala version accordingly.
     val (sparkVersion, scalaVersion) = sparkSubmitVersion(livyConf)
     testSparkVersion(sparkVersion)
-    // If Spark and Scala version is not set, set into livy configuration, this will be used by
+
+    // If Spark and Scala version is set manually, should verify if they're consistent with
+    // ones parsed from "spark-submit --version"
+    val formattedSparkVersion = formatSparkVersion(sparkVersion)
+    Option(livyConf.get(LIVY_SPARK_VERSION)).map(formatSparkVersion).foreach { version =>
+      require(formattedSparkVersion == version,
+        s"Configured Spark version $version is not equal to Spark version $formattedSparkVersion " +
+          "got from spark-submit -version")
+    }
+
+    if (!scalaVersion.isEmpty && Option(livyConf.get(LIVY_SPARK_SCALA_VERSION)).isDefined) {
+      Option(livyConf.get(LIVY_SPARK_SCALA_VERSION))
+        .map(s => formatScalaVersion(s, formattedSparkVersion))
+        .foreach { version =>
+          require(version == scalaVersion,
+            s"Configured Scala version $version is not equal to Scala version $scalaVersion got " +
+              "from spark-submit -version")
+        }
+    }
+
+    // Set formatted Spark and Scala version into livy configuration, this will be used by
     // session creation.
-    livyConf.setIfMissing(LIVY_SPARK_VERSION.key,
-      formatSparkVersion(sparkVersion).productIterator.mkString("."))
-    livyConf.setIfMissing(LIVY_SPARK_SCALA_VERSION.key,
-      formatScalaVersion(scalaVersion, formatSparkVersion(sparkVersion)))
+    livyConf.set(LIVY_SPARK_VERSION.key, formattedSparkVersion.productIterator.mkString("."))
+    livyConf.set(LIVY_SPARK_SCALA_VERSION.key,
+      formatScalaVersion(scalaVersion, formattedSparkVersion))
 
     testRecovery(livyConf)
 
--- a/server/src/main/scala/com/cloudera/livy/server/interactive/CreateInteractiveRequest.scala
+++ b/server/src/main/scala/com/cloudera/livy/server/interactive/CreateInteractiveRequest.scala
@@ -36,6 +36,5 @@ class CreateInteractiveRequest {
   var queue: Option[String] = None
   var name: Option[String] = None
   var conf: Map[String, String] = Map()
-  var scalaVersion: Option[String] = None
 
 }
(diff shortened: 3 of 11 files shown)


server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
```diff
@@ -121,12 +121,18 @@ class InteractiveSession(
     }
     builderProperties.put(RSCConf.Entry.SESSION_KIND.key, kind.toString)
 
-    mergeConfList(livyJars(livyConf), LivyConf.SPARK_JARS)
+    val (sparkVersionStr, scalaVersionStr) = LivySparkUtils.sparkSubmitVersion(livyConf)
```
tc0312: sparkSubmitVersion() will run every time a session is created.

jerryshao: Yes, the purpose here is to combine Spark/Scala with Session. Currently Spark/Scala is a global settings, we need to change them to be session based, that will support difference Sparks in one Livy.

server/src/main/scala/com/cloudera/livy/server/interactive/CreateInteractiveRequest.scala
```diff
@@ -36,5 +36,6 @@ class CreateInteractiveRequest {
   var queue: Option[String] = None
   var name: Option[String] = None
   var conf: Map[String, String] = Map()
+  var scalaVersion: Option[String] = None
```
jerryshao: The purpose is also to isolate Scala version per session based. Usually user don't need to set this, Livy itself will figure out.

server/src/main/scala/com/cloudera/livy/utils/LivySparkUtils.scala
```diff
@@ -89,27 +100,54 @@ object LivySparkUtils {
...
+
+  /**
+   * Return formatted Scala version, if it cannot be parsed from input version string, it will
+   * pick default Scala version related to Spark version.
+   *
+   * @param version Scala version String
+   * @param sparkVersion formatted Spark version.
+   * @return Tuple of Scala version, one is major version and the other is minor version.
+   */
+  def formatScalaVersion(version: String, sparkVersion: (Int, Int)): (Int, Int) = {
```
tc0312: I think we can just handle Scala version as a string and do simple string matching here. As we are not using the Scala version number to do comparsion. Also since we support only "2.10" and "2.11", there's no much combinations.

server/src/main/scala/com/cloudera/livy/server/interactive/CreateInteractiveRequest.scala
tc0312: I understand that. But I think we don't have a use case that requires different Scala version on a per session basis. To make this work, admins have to carefully setup CLASSPATH to include both spark jars from both Scala versions. Unless a lot of users are asking for it, I would avoid supporting it.

In the future we might want to support different Spark version on a per session basis. It's a big item. Mind if we tackle it separately, and keep Scala version a global LivyConf?

jerryshao: This will only be used for user who wants to use Spark 1.6 + Scala 2.11 or Spark 2.0 + Scala 2.10. Also Spark will figure out this Scala version in future, so usually user doesn't need to take care of this field.

I could remove this field, but it will lose the ability to use Spark with unofficial Scala buildings.

Also setting Scala version as a global conf will further limit the usage. For example if we set to scala 2.10, so only Spark 1.6 and 2.0 with Scala 2.10 can be worked. In that case why do we need to ship two different repl bundles, it may not be so useful.

jerryshao: Also for the classpath thing, I don't think we need to worry about it. SparkLaunch and spark-class will handle this, currently what Livy will inject into Spark's classpath is rsc and repl jars.

This field and per-session based Scala version check may not be useful currently, since SPARK_HOME is still a global setting. But as I mentioned this is one task for Livy to support different Sparks.

Also this fix actually will not break current usage and no behavior changes for user.

tc0312: I think we agree on the ultimate goal to support 2 Spark versions in runtime. However, we have different opinions on milestones. 
1. Support 2 Spark versions in a single assembly.
2. Support 2 Spark versions in runtime.

With milestone 1, we can release a single Livy assembly for different Scala and Spark version. This addresses the immediate need for my team.

Milestone 2 requires a lot of work. We need to think about how the configuration is laid-out (e.g. specify multiple SPARK_HOME). For Spark 1.6, how do we specify unofficial Scala binding per SPARK_HOME. How do users pick a specific Spark version with a specific Scala binding for batch, interactive and rsc. I don't think we have answers for these questions and I prefer to address milestone 1 first.

> I could remove this field, but it will lose the ability to use Spark with unofficial Scala buildings.
We can still support unofficial Scala buildings by changing it to a LivyConf.

tc0312: > This field and per-session based Scala version check may not be useful currently, since SPARK_HOME is still a global setting. But as I mentioned this is one task for Livy to support different Sparks.

Since it's not useful currently, I would prefer to not expose this in CreateInteractiveSession and remain a global flag.

tc0312: Thanks a lot Jerry.

tc0312:
LGTM, will merge if there's no other feedback.
@zjffdu @vanzin

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
```diff
@@ -121,13 +121,18 @@ class InteractiveSession(
     }
     builderProperties.put(RSCConf.Entry.SESSION_KIND.key, kind.toString)
 
-    mergeConfList(livyJars(livyConf), LivyConf.SPARK_JARS)
+    require(livyConf.get(LivyConf.LIVY_SPARK_VERSION) != null)
+    require(livyConf.get(LivyConf.LIVY_SPARK_SCALA_VERSION) != null)
+
+    val sparkVersion = LivySparkUtils.formatSparkVersion(livyConf.get(LivyConf.LIVY_SPARK_VERSION))
```
zjffdu: how about changing it to 
```
val (sparkMajorVersion, sparkMinorVersion) = ...
```
Otherwise the following 2 statements would make me think they return the same type. 
```
val sparkVersion = LivySparkUtils.formatSparkVersion(livyConf.get(LivyConf.LIVY_SPARK_VERSION))
val scalaVersion = livyConf.get(LivyConf.LIVY_SPARK_SCALA_VERSION)
```

server/src/main/scala/com/cloudera/livy/utils/LivySparkUtils.scala
```diff
@@ -27,6 +27,22 @@ import com.cloudera.livy.util.LineBufferedProcess
 
 object LivySparkUtils {
 
+  // For each Spark version we supported, we need to add this mapping relation in case Scala
+  // version cannot be detected from "spark-submit --version".
+  private val defaultSparkScalaVersion = Map(
+    // Spark 2.0 + Scala 2.11
+    (2, 0) -> "2.11",
```
zjffdu: does that mean we need to update this if we want to support spark 2.1 ?

server/src/main/scala/com/cloudera/livy/utils/LivySparkUtils.scala
```diff
@@ -112,4 +132,23 @@ object LivySparkUtils {
...
+
+  /**
+   * Return Scala binary version, if it cannot be parsed from input version string, it will
+   * pick default Scala version related to Spark version.
+   *
+   * @param version Scala binary version String
+   * @param sparkVersion formatted Spark version.
+   * @return Scala binary version String based on Spark version and livy conf.
+   */
+  def formatScalaVersion(version: String, sparkVersion: (Int, Int)): String = {
```
zjffdu: version -> scalaVersion, version is a little vague.

zjffdu: And how about change it to formatScalaVersion(scalaVersion:String, sparkVersion:String) ?
formatSparkVersion can be called inside this method.

server/src/test/scala/com/cloudera/livy/utils/LivySparkUtilsSuite.scala
```diff
@@ -41,13 +42,15 @@ class LivySparkUtilsSuite extends FunSuite with LivyBaseUnitTestSuite {
     testSparkVersion("1.6.0")
     testSparkVersion("1.6.1")
     testSparkVersion("1.6.2")
+    testSparkVersion("1.6")
```
zjffdu: add HDP version test as well ?

server/src/main/scala/com/cloudera/livy/utils/LivySparkUtils.scala
tc0312: Agree. I think we should remove the max version check. Livy can print a warning instead.

server/src/test/scala/com/cloudera/livy/utils/LivySparkUtilsSuite.scala
jerryshao: Added.

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
jerryshao: Done.

server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
```diff
@@ -64,7 +64,16 @@ class LivyServer extends Logging {
...
+
+    // Test spark-submit and get Spark Scala version accordingly.
+    val (sparkVersion, scalaVersion) = sparkSubmitVersion(livyConf)
+    testSparkVersion(sparkVersion)
+    // If Spark and Scala version is not set, set into livy configuration, this will be used by
+    // session creation.
+    livyConf.setIfMissing(LIVY_SPARK_VERSION.key,
+      formatSparkVersion(sparkVersion).productIterator.mkString("."))
+    livyConf.setIfMissing(LIVY_SPARK_SCALA_VERSION.key,
+      formatScalaVersion(scalaVersion, formatSparkVersion(sparkVersion)))
```
jerryshao: Done.

server/src/main/scala/com/cloudera/livy/LivyConf.scala
```diff
@@ -42,6 +42,13 @@ object LivyConf {
   val LIVY_SPARK_MASTER = Entry("livy.spark.master", "local")
   val LIVY_SPARK_DEPLOY_MODE = Entry("livy.spark.deployMode", null)
 
+  // Two configurations to specify Spark and related Scala version. These are internal
+  // configurations will be set by LivyServer and used in session creation. It is not required to
+  // set usually unless running with unofficial Spark + Scala versions
+  // (like Spark 2.0 + Scala 2.10, Spark 1.6 + Scala 2.11)
+  val LIVY_SPARK_SCALA_VERSION = Entry("livy.spark.scalaVersion", null)
+  val LIVY_SPARK_VERSION = Entry("livy.spark.version", null)
```
tc0312: I've given these configs more thought. I think we should remove livy.spark.version because we are always detecting the version from spark-submit. And we are using require() to make sure they match. I don't think this LivyConf is useful for customers.
Is this added to pass spark version and scala version to InteractiveSession?

If you are too busy to make the change, I can do it in #211.

tc0312: Ideally LivyConf should remain to be an immutable conf object. Also I think we are passing a lot of other objects (like SessionStore, LivyConf) from LivyServer to SessionServlets. We can extract them to a class plus the Scala and Spark version and pass them from LivyServer to Servlets.

Since we need to move forward with the Spark 2.0 IT change, let's deal with this later.

tc0312:
@zjffdu Does this look good to you? Will merge when you sign off.

zjffdu:
LGTM

PR #207, opened 2016-10-17
LIVY-228. Shade json4s dependencies and mitigate py4j difference

jerryshao:


tc0312:
LGTM.

jerryshao:
\cc @tc0312

repl/pom.xml
```diff
@@ -174,6 +211,15 @@
                         <goals>
                             <goal>copy-dependencies</goal>
                         </goals>
+                        <configuration>
+                            <excludeArtifactIds>
+                                json4s-ast_${scala.binary.version},
+                                json4s-core_${scala.binary.version},
+                                json4s-jackson_${scala.binary.version},
+                                paranamer,
+                                scalap
```
zjffdu: why paranamer, scalap is excluded here too ?

jerryshao: scalap will be assembled into shade jar, so don't need to copy into jars folder.

PR #208, opened 2016-10-18
LIVY-212. Implemented session recovery for interactive session. Only work with YARN.

tc0312:
livy-server can now recover interactive sessions after restart.
There are some special cases:
- If livy-server crashes while an interactive session is starting, there's a chance the session is unrecoverable, depending on timing (whether livy-repl has registered to livy-server).
- If livy-server is down longer than the value of server.idle_timeout (default: 10min), livy-repl will timeout and quit.

Note: All previous statements are lost after recovery. This will be fixed in a different commit.

### DIFF
--- a/integration-test/src/main/scala/com/cloudera/livy/test/framework/LivyRestClient.scala
+++ b/integration-test/src/main/scala/com/cloudera/livy/test/framework/LivyRestClient.scala
@@ -145,8 +145,10 @@ class LivyRestClient(val httpClient: AsyncHttpClient, val livyEndpoint: String) 
               val data = output("data").asInstanceOf[Map[String, Any]]
               Left(data("text/plain").asInstanceOf[String])
             case Some("error") => Right(mapper.convertValue(output, classOf[StatementError]))
-            case Some(status) => throw new Exception(s"Unknown statement $stmtId status: $status")
-            case None => throw new Exception(s"Unknown statement $stmtId output: $newStmt")
+            case Some(status) =>
+              throw new IllegalStateException(s"Unknown statement $stmtId status: $status")
+            case None =>
+              throw new IllegalStateException(s"Unknown statement $stmtId output: $newStmt")
           }
         }
       }
--- a/integration-test/src/main/scala/com/cloudera/livy/test/framework/LivyRestClient.scala
+++ b/integration-test/src/main/scala/com/cloudera/livy/test/framework/LivyRestClient.scala
@@ -99,7 +99,7 @@ class LivyRestClient(val httpClient: AsyncHttpClient, val livyEndpoint: String) 
       // Keeping the original timeout to avoid slowing down local development.
       eventually(timeout(t), interval(1 second)) {
         val s = snapshot().state
-        assert(strStates.contains(s), s"Session state $s doesn't equal one of $strStates")
+        assert(strStates.contains(s), s"Session $id state $s doesn't equal one of $strStates")
       }
     }
 
@@ -145,8 +145,8 @@ class LivyRestClient(val httpClient: AsyncHttpClient, val livyEndpoint: String) 
               val data = output("data").asInstanceOf[Map[String, Any]]
               Left(data("text/plain").asInstanceOf[String])
             case Some("error") => Right(mapper.convertValue(output, classOf[StatementError]))
-            case Some(status) => throw new Exception(s"Unknown statement status: $status")
-            case None => throw new Exception(s"Unknown statement output: $newStmt")
+            case Some(status) => throw new Exception(s"Unknown statement $stmtId status: $status")
+            case None => throw new Exception(s"Unknown statement $stmtId output: $newStmt")
           }
         }
       }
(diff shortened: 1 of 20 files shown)


javierluraschi:
This PR improves some statements execution by 10X, this is great and really needed in livy. For instance, a simple "1 + 1" statement post + polling went from 1.220s down to 0.117s after applying this patch.

@tc0312 aside form session recovery, would you mind pointing out the specific change that improved this significantly? Are there any other statement execution perf-improvements your can think of that are worth exploring?

zjffdu:
@tc0312 I restart the livy server,and the interactive session is still killed. Does the recovery needs any configuration ?

tc0312:
@zjffdu Yes. Example livy.conf:
```
livy.server.recovery.mode = recovery
livy.server.recovery.state-store = filesystem
livy.server.recovery.state-store.url = file:///tmp
```

server/src/main/scala/com/cloudera/livy/sessions/SessionManager.scala
```diff
@@ -24,77 +24,51 @@ import java.util.concurrent.atomic.AtomicInteger
...
 
-// TODO Replace SessionManager with this class when interactive sessions support recovery.
 class BatchSessionManager(
     livyConf: LivyConf,
     sessionStore: SessionStore,
     mockSessions: Option[Seq[BatchSession]] = None)
-  extends SessionManager[BatchSession](livyConf) {
+  extends SessionManager[BatchSession, BatchRecoveryMetadata] (
+    livyConf, BatchSession.recover(_, livyConf, sessionStore), sessionStore, "batch", mockSessions)
+  { }
```
jerryshao: this `{ }` may not be necessary.

tc0312:
@jerryshao 
> When server is done how can we talk with dangling session? From my understanding of Livy code, user needs to talk to Livy server, and LivyServer forwards the statements to driver.

Your understanding is correct. Session is unavailable until livy-server is restarted. This is by design.

> After server is restarted (trying to recover the session), does it need to reconnect to the RPC channel with repl driver? From the code I'm not sure if it is already addressed.
It's implemented [here](https://github.com/cloudera/livy/pull/208/files#diff-24336a07af4c97fea3e9e55bda022f5dR128)

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
```diff
@@ -30,54 +30,228 @@ import scala.collection.mutable
...
+  }
 
-  val kind = request.kind
+  def recover(
+      metadata: InteractiveRecoveryMetadata,
+      livyConf: LivyConf,
+      sessionStore: SessionStore,
+      mockApp: Option[SparkApp] = None,
+      mockClient: Option[RSCClient] = None): InteractiveSession = {
+    val client = mockClient orElse metadata.rscDriverUri.map { uri =>
```
jerryshao: `mockClient.orElse(...)`, better not to use postfix operation.

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
```diff
@@ -139,83 +313,96 @@ class InteractiveSession(
...
-      "spark.executor.instances" -> request.numExecutors.map(_.toString)
-    )
+    builderProperties
+  }
+}
 
-    userOpts.foreach { case (key, opt) =>
-      opt.foreach { value => builderProperties.put(key, value) }
-    }
+class InteractiveSession(
```
jerryshao: Can we simplify or use case class to group the constructor parameters of `InteractiveSession`, currently it is a bit more.

jerryshao:
The implementation LGTM.

integration-test/src/main/scala/com/cloudera/livy/test/framework/LivyRestClient.scala
```diff
@@ -145,8 +145,8 @@ class LivyRestClient(val httpClient: AsyncHttpClient, val livyEndpoint: String)
               val data = output("data").asInstanceOf[Map[String, Any]]
               Left(data("text/plain").asInstanceOf[String])
             case Some("error") => Right(mapper.convertValue(output, classOf[StatementError]))
-            case Some(status) => throw new Exception(s"Unknown statement status: $status")
-            case None => throw new Exception(s"Unknown statement output: $newStmt")
+            case Some(status) => throw new Exception(s"Unknown statement $stmtId status: $status")
```
tc0312: Changed them to IllegalStateException.

server/src/main/scala/com/cloudera/livy/sessions/SessionManager.scala
tc0312: Removed.

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
tc0312: I agree but mind if I do it in another PR? I want to keep Batch and Interactive code similar/symmetric. Since this's already a big PR, I don't want to make it even bigger.

tc0312:
@zjffdu @vanzin

PR #209, opened 2016-10-18
LIVY-229. Spark version check fails when it's used with HDP built Spark.

tc0312:
Fixed Spark version regex to support version like 2.0.0.2.5.1.0-56.


tc0312:
@zjffdu @jerryshao Mind if you two take a look?

jerryshao:
LGTM, I think we should add more test one this, like CDH, HDI (not sure if it is issue).

PR #210, opened 2016-10-19
LIVY-230. SparkSession is not avalible for PySparkInterperter and SparkRInterperter

zjffdu:
This PR make SparkSession avalible in PySparkInterpreter and SparkRInterpreter.

### DIFF
--- a/repl/src/main/resources/fake_shell.py
+++ b/repl/src/main/resources/fake_shell.py
@@ -514,7 +514,7 @@ def main():
     sys.stdout = UnicodeDecodingStringIO()
     sys.stderr = UnicodeDecodingStringIO()
 
-    spark_major_version = os.getenv("SPARK_MAJOR_VERSION")
+    spark_major_version = os.getenv("LIVY_SPARK_MAJOR_VERSION")
     try:
         listening_port = 0
         if os.environ.get("LIVY_TEST") != "true":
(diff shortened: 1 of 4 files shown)


tc0312:
LGTM. Do you mind adding a test for it?

jerryshao:
@linchan-ms adding Spark2 supports for `JobContextImpl` is still in consideration, it is not so straightforward compared to interpreter things.

repl/src/main/resources/fake_shell.py
```diff
@@ -523,6 +524,8 @@ def main():
             exec('from pyspark.sql import HiveContext', global_dict)
```
tc0312: I think this is imported for PySpark RSC support.
I think in the long run we should separate repl code and rsc code to 2 separate python files. And we can fix this when we add SparkSession support for RSC.

repl/src/main/scala/com/cloudera/livy/repl/SparkRInterpreter.scala
```diff
@@ -125,9 +126,10 @@ object SparkRInterpreter {
   }
 }
 
-class SparkRInterpreter(process: Process, backendInstance: Any, backendThread: Thread)
-  extends ProcessInterpreter(process)
-{
+class SparkRInterpreter(process: Process,
+                        backendInstance: Any,
```
zjffdu: Isn't this style more readable ?

tc0312: Personally I have no opinion on this. But existing code follows Spark's coding style. Let's be consistent and follow Spark's coding style.

zjffdu: Fixed

PR #211, opened 2016-10-20
LIVY-219. Support Scala-2.11 integration test for Livy IT.

tc0312:
- Added Maven submodules (minicluster-dependencies) to write Spark classpath of different Scala version to files for MiniCluster.
- Writing classpath to a file instead of Maven properties to maintain ability to run integration tests from IDEs.
- MiniCluster picks and reads the correct Spark classpath and pass it to MiniLivy via spark-default.conf.
- Dependency changes to make sure both repl is built when Maven runs integration-test.
- Verify "ensure Livy internal configurations are not exposed" in InteractiveIT instead of JobsApiIT to avoid job incompatibility issue between Scala 2.10 and 2.11.
- Changed Maven pom.xml to start integration tests in 2.10 and 2.11.
- MiniCluster sets env. var. to tell LivyServer what Spark Scala version it should use.
- Changed Travis to launch 2 builds for Spark 1.6 and Spark 2.0 with Scala 2.10 and 2.11.
- During development, to skip integration test for a Scala version, use `-DskipITs-2.10` or `-DskipITs-2.11`. `-DskipITs` will skip both.
- Fixed JaCoCo EOFException by disabling JaCoCo for `InteractiveIT.should kill RSCDriver if it doesn't respond to end session`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,8 @@ dist: trusty
 language: scala
 
 env:
-  - MVN_FLAG=
-  - MVN_FLAG=-Dminicluster-2.10.spark.version=2.0.1 -Dminicluster-2.11.spark.version=1.6.2
+  - MVN_FLAG=-Pspark-1.6
+  - MVN_FLAG=-Pspark-2.0
 
 jdk:
   - oraclejdk7
@@ -30,7 +30,7 @@ before_install:
   - pip3 install --user cloudpickle
 
 install:
-  - mvn install -Dskip -DskipTests -DskipITs -Dmaven.javadoc.skip=true -B -V
+  - mvn $MVN_FLAG install -Dskip -DskipTests -DskipITs -Dmaven.javadoc.skip=true -B -V
 
 before_script:
   - pip install --user requests pytest flaky flake8
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,10 @@ sudo: required
 dist: trusty
 language: scala
 
+env:
+  - MVN_FLAG=
+  - MVN_FLAG=-Dminicluster-2.10.spark.version=2.0.1 -Dminicluster-2.11.spark.version=1.6.2
+
 jdk:
   - oraclejdk7
 
@@ -33,7 +37,7 @@ before_script:
   - pip3 install --user requests pytest flaky
 
 script:
-  - mvn verify -e
+  - mvn $MVN_FLAG verify -e
 
 after_success:
   - bash <(curl -s https://codecov.io/bash)
--- a/integration-test/minicluster-dependencies/pom.xml
+++ b/integration-test/minicluster-dependencies/pom.xml
@@ -29,12 +29,11 @@
     <relativePath>../../scala/pom.xml</relativePath>
     <version>0.3.0-SNAPSHOT</version>
   </parent>
-  <artifactId>livy-integration-test-minicluster-dependencies-parent</artifactId>
+  <artifactId>minicluster-dependencies-parent</artifactId>
   <version>0.3.0-SNAPSHOT</version>
   <packaging>pom</packaging>
   <properties>
     <skipDeploy>true</skipDeploy>
-    <minicluster.spark.version></minicluster.spark.version>
   </properties>
   <dependencies>
     <dependency>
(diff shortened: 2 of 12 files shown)


server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
```diff
@@ -88,10 +88,13 @@ class LivyServer extends Logging {
...
     }
 
+    // TODO Create a new class to pass variables from LivyServer to sessions and remove these
+    // internal LivyConfs.
     // Set formatted Spark and Scala version into livy configuration, this will be used by
     // session creation.
     livyConf.set(LIVY_SPARK_VERSION.key, formattedSparkVersion.productIterator.mkString("."))
-    livyConf.set(LIVY_SPARK_SCALA_VERSION.key,
+    // Integration test sets spark scala version.
+    livyConf.setIfMissing(LIVY_SPARK_SCALA_VERSION.key,
```
jerryshao: This will introduce bug if user set `LIVY_SPARK_SCALA_VERSION` like `2.10.4` or `2.11.8`. We should format user's input to `major.minor` and reset to LivyConf to override it. Otherwise if user set "2.10.4", this will introduce issue in `InteractiveSession`.

jerryshao: I deliberately use `set` instead.

.travis.yml
```diff
@@ -2,6 +2,10 @@ sudo: required
 dist: trusty
 language: scala
 
+env:
+  - MVN_FLAG=
+  - MVN_FLAG=-Dminicluster-2.10.spark.version=2.0.1 -Dminicluster-2.11.spark.version=1.6.2
```
jerryshao: Why test 2.10 with Spark2 and 2.11 with Spark 1.6, should we test official version?

integration-test/minicluster-dependencies/pom.xml
```diff
@@ -0,0 +1,136 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>com.cloudera.livy</groupId>
+    <artifactId>multi-scala-project-root</artifactId>
+    <relativePath>../../scala/pom.xml</relativePath>
+    <version>0.3.0-SNAPSHOT</version>
+  </parent>
+  <artifactId>livy-integration-test-minicluster-dependencies-parent</artifactId>
```
jerryshao: Can we simplify the artificatId, it looks too long to me.

integration-test/pom.xml
```diff
@@ -64,6 +90,33 @@
 
     <dependency>
       <groupId>${project.groupId}</groupId>
+      <artifactId>livy-integration-test-minicluster-dependencies_2.10</artifactId>
```
jerryshao: Is it safe to put both 2.10 and 2.11 jar in the same classpath, from my understanding it is hard to guarantee the method we called is coming from 2.10 rather 2.11.

jerryshao: If we shift to Scala 2.11 for default supported version, how we can change to use 2.11 one.

integration-test/src/main/scala/com/cloudera/livy/test/framework/MiniCluster.scala
```diff
@@ -52,6 +54,13 @@ private class MiniClusterConfig(val config: Map[String, String]) {
 }
 
 sealed trait MiniClusterUtils extends ClusterUtils {
+  private val livySparkScalaVersionEnvVarName = "LIVY_SPARK_SCALA_VERSION"
+
+  protected def getSparkScalaVersion(): String = {
+    sys.env.get(livySparkScalaVersionEnvVarName).getOrElse {
```
jerryshao: `sys.env.getOrElse(xxx, xxx)`

integration-test/pom.xml
```diff
@@ -51,6 +53,30 @@
 
     <dependency>
       <groupId>${project.groupId}</groupId>
+      <artifactId>livy-repl_2.10</artifactId>
```
jerryshao: Why do we need to import repl jars, I think it is not necessary, `InteractiveSession` will pick and upload to Spark automatically if existed in repl/scala-2.xx/target/jars.

scala/pom.xml
```diff
@@ -32,6 +32,10 @@
     <relativePath>../pom.xml</relativePath>
   </parent>
 
+  <properties>
+    <minicluster.spark.version>100.1.1</minicluster.spark.version>
```
jerryshao: Is it necessary to write here again?

jerryshao:
Basically LGTM, I'm wondering if some artifact ids and properties can simplified, it looks too long and little puzzled to me.

.travis.yml
tc0312: Official version is tested by line 6.
The default for minicluster-2.10.spark.version is 1.6.2 and inicluster-2.11.spark.version is 2.0.1

integration-test/pom.xml
tc0312: This is for not importing repl jars to the classpath.This is a build dependency to make sure repl-2.10 and repl-2.11 projects are built before integration-test starts.

integration-test/pom.xml
tc0312: minicluster-dependencies projects are special. They don't produce a jar. Also, <exclusions> tags below removes all their dependencies. They add nothing to the classpath.
This's also a build dependency to make sure minicluster-dependencies submodules are processed before integration-test.

server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
tc0312: I changed it to sefIfMissing so integration test can pass Scala version to LivyServer. Let me change it a bit to address your concern.

integration-test/minicluster-dependencies/pom.xml
tc0312: Changed it to minicluster-dependencies-parent. Feel free to purpose better name.

integration-test/src/main/scala/com/cloudera/livy/test/framework/MiniCluster.scala
tc0312: Fixed.

PR #213, opened 2016-10-28
LIVY-225. Fix two issues running integration test with Spark 2.0.

tc0312:
1. Create a dummy jar to let Spark yarn#client correctly launch the application.
2. Since `SQLContext` is deprecated in Spark 2.0, so additional information will be printed out when creating. Changing the regex pattern.


tc0312:
#202 but with conflict resolved.

PR #214, opened 2016-10-28
LIVY-213. Implemented interactive statement results recovery.

tc0312:
- Removed statement results history cache in livy-server.
- Proxy statement results history requests to livy-repl.
- Modified RSC repl API:
  - ReplJobRequest to be synchronus and return the id of new statement.
  - GetReplJobResults to return all or some statements in sorted order. Also return current repl state.
  - Reuse Statement and StatementState classes in server, repl and rsc.
- livy-server polls only repl state but not statement results from repl. Repl's state is also updated on statement result requests.
- New StatementState "waiting" to show a stetement that's submitted to repl but hasn't been executed.
- Send statement results to server in JSON instead of JValue to avoid Kryo issue.


tc0312:
@jerryshao @zjffdu @vanzin Mind if you review my change?

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
```diff
@@ -351,6 +348,21 @@ class InteractiveSession(
...
+          // State is also updated when we get statement results from repl, not just here.
+          setSessionStateFromReplState(client.map(_.getReplState.get()))
+          Thread.sleep(1000)
+        }
+      } catch {
+        case _: InterruptedException =>
+      }
+    }
+  })
+
```
tc0312: No. But I want to do this in a separate change.
See // TODO Replace this with a Rpc call from repl to server.

repl/src/main/scala/com/cloudera/livy/repl/Session.scala
```diff
@@ -51,7 +54,9 @@ class Session(interpreter: Interpreter)
   private implicit val formats = DefaultFormats
 
   private var _state: SessionState = SessionState.NotStarted()
-  private var _history = IndexedSeq[Statement]()
+  private val _statements = mutable.Map[Int, Statement]()
```
jerryshao: I would suggest to use guava's cache instead of this `Map`, that will be easy for you to later support retention.

repl/src/main/scala/com/cloudera/livy/repl/Session.scala
```diff
@@ -70,29 +75,41 @@ class Session(interpreter: Interpreter)
...
+    _statements.synchronized {
+      _statements(statementId) = new Statement(statementId, StatementState.Waiting, null)
+    }
+    Future {
+      _statements.synchronized {
+        _statements(statementId) = new Statement(statementId, StatementState.Running, null)
+      }
+
+      val statement =
+        new Statement(statementId, StatementState.Available, executeCode(statementId, code))
```
jerryshao: `executeCode` should also be in synchronized block, otherwise interpreter will potentially has concurrency issue.

repl/src/test/scala/com/cloudera/livy/repl/BaseSessionSpec.scala
```diff
@@ -22,18 +22,28 @@ import scala.concurrent.Await
...
 import com.cloudera.livy.LivyBaseUnitTestSuite
+import com.cloudera.livy.rsc.driver.{Statement, StatementState}
 import com.cloudera.livy.sessions.SessionState
 
 abstract class BaseSessionSpec extends FlatSpec with Matchers with LivyBaseUnitTestSuite {
 
   implicit val formats = DefaultFormats
 
-  def withSession(testCode: Session => Any): Unit = {
+  protected def execute(session: Session, code: String): Statement = {
```
jerryshao: Can we change to `withSession(session: Session)(code: => String)`?

rsc/src/main/java/com/cloudera/livy/rsc/BaseProtocol.java
```diff
@@ -171,30 +172,30 @@ public RemoteDriverAddress() {
...
+      this.allResults = false;
+      this.from = from;
+      this.size = size;
     }
 
-    public GetReplJobResult() {
-      this(null);
+    public GetReplJobResults() {
+      this.allResults = true;
+      from = size = null;
```
jerryshao: I would avoid this chain assignment.

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
```diff
@@ -486,54 +510,28 @@ class InteractiveSession(
...
+  private def transition(newState: SessionState) = synchronized {
     // When a statement returns an error, the session should transit to error state.
     // If the session crashed because of the error, the session should instead go to dead state.
     // Since these 2 transitions are triggered by different threads, there's a race condition.
     // Make sure we won't transit from dead to error state.
-    if (!_state.isInstanceOf[SessionState.Dead] || !state.isInstanceOf[SessionState.Error]) {
-      debug(s"$this session state change from ${_state} to $state")
-      _state = state
+    val areSameStates = _state.getClass() == newState.getClass()
+    val transitFromInactiveToActive = !_state.isActive && newState.isActive
```
jerryshao: What's the usage of condition here, I don't understand?

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
jerryshao: I agree with @zjffdu , this will create one thread per session, also do we need to pull the repl state so frequently? Each session will send RPC requests per second, that's a big overhead if there's lots of session.

tc0312:
Sorry for the big PR. This was done way before I submit the last PR. I will submit new work in smaller incremental PR.

integration-test/src/test/scala/com/cloudera/livy/test/JobApiIT.scala
```diff
@@ -76,7 +76,8 @@ class JobApiIT extends BaseIntegrationTestSuite with BeforeAndAfterAll with Logg
       assert(list.total === 1)
       val tempSessionId = list.sessions(0).id
 
-      livyClient.connectSession(tempSessionId).verifySessionIdle()
+      val s = livyClient.connectSession(tempSessionId)
```
tc0312: I thought users shouldn't submit RSC jobs while the session's starting.
I will fix the code to re-enable this use case then revert this change.

repl/src/main/scala/com/cloudera/livy/repl/Session.scala
tc0312: Repl was using Guava and caused a lot of jar collision problems with user jars. We don't want to shade Guava either because it's pretty large.
Mind if we keep this PR solely for recovery and deal with this later?

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
tc0312: Yup. I don't disagree with that. See // TODO Replace this with a Rpc call from repl to server.
Replacing something timer based to Rpc usually introduces race conditions. It's better if I do it in a separate PR so you can review it separately.

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
tc0312: Which condition? areSameStates or transitFromInactiveToActive?

repl/src/main/scala/com/cloudera/livy/repl/Session.scala
tc0312: Fixed.

repl/src/test/scala/com/cloudera/livy/repl/BaseSessionSpec.scala
tc0312: Then wouldn't it create a new session every time when the test executes a statement?

rsc/src/main/java/com/cloudera/livy/rsc/BaseProtocol.java
tc0312: Fixed.

repl/src/main/scala/com/cloudera/livy/repl/Session.scala
jerryshao: Java `LinkedHashMap` is another one for you to try. It's fine here, we could improve later.

repl/src/test/scala/com/cloudera/livy/repl/BaseSessionSpec.scala
jerryshao: No, `session` argument is the one you already created. Changing to curry style makes it more scala-ish, you could further define `val b = withSession(session)` and `b(code)` to simplify.

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
jerryshao: the latter?

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
tc0312: Sorry I didn't pay attention to that because I'm gonna replace it with a Rpc callback. Will fix.

repl/src/test/scala/com/cloudera/livy/repl/BaseSessionSpec.scala
tc0312: code is a string of test code, in Scala/Python/R, which's sent to the interpreter class for testing.

jerryshao: You may misunderstood my point, my suggestion is to change to:

```
val statement = execute(session)("blahblah")

or

val execution = execute(session)
val statement = execution("blah")
```

Not semantic changes, only the style changes.

tc0312: I don't really think it makes a big difference but I will make the change.

jerryshao: Yes, there's no big difference, what I want is to keep the same style as original code `withXXX()` or `withXXX(xxx)(...)`. I don't strongly ask you to change, just a suggestion.

repl/src/main/scala/com/cloudera/livy/repl/ReplDriver.scala
```diff
@@ -73,19 +63,28 @@ class ReplDriver(conf: SparkConf, livyConf: RSCConf)
...
+  def handle(ctx: ChannelHandlerContext, msg: BaseProtocol.ReplJobRequest): Int = {
+    session.execute(msg.code)
   }
 
-  def handle(ctx: ChannelHandlerContext, msg: BaseProtocol.GetReplJobResult): String = {
-    val result = jobFutures.getOrElse(msg.id, null)
-    Option(result).map { r => compact(render(r)) }.orNull
+  /**
+    * Return statement results. Results are sorted by statement id.
+    */
```
jerryshao: Please change to the javadoc comment style:

```
/**
 *
 */
```

repl/src/main/scala/com/cloudera/livy/repl/Session.scala
```diff
@@ -70,29 +75,41 @@ class Session(interpreter: Interpreter)
 
   def state: SessionState = _state
 
-  def history: IndexedSeq[Statement] = _history
+  def statements: mutable.Map[Int, Statement] = _statements
+
+  def execute(code: String): Int = {
+    val statementId = newStatementId.getAndIncrement()
+    _statements.synchronized {
```
jerryshao: I'm wondering if we could use `ConcurrentHashMap`, then this synchronization block is not necessary, what do you think?

repl/src/main/scala/com/cloudera/livy/repl/ReplDriver.scala
```diff
@@ -73,19 +63,28 @@ class ReplDriver(conf: SparkConf, livyConf: RSCConf)
...
+  def handle(ctx: ChannelHandlerContext, msg: BaseProtocol.GetReplJobResults): ReplJobResults = {
+    val stmts = if (msg.allResults) {
+      session.statements.values.toArray
+    } else {
+      assert(msg.from != null)
+      assert(msg.size != null)
+      val until = msg.from + msg.size
+      session.statements.filterKeys(id => id >= msg.from && id < until).values.toArray
```
jerryshao: I think there're two potential issues here:

1. Since `statements` is mutable, here without synchronization, the returned result may be inconsistent.
2. Traverse through the who map will be slower if Livy is running for hours, it would be better to use other data structure like sorted map instead.

Also I'm not sure in which scenario do we require to bulk get the statement results?

repl/src/main/scala/com/cloudera/livy/repl/Session.scala
```diff
@@ -140,7 +157,7 @@ class Session(interpreter: Interpreter)
         (EVALUE -> e.getMessage) ~
         (TRACEBACK -> List())
     }
+
+    compact(render(resultInJson))
```
jerryshao: What is the advantage of render to json here in driver side?

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
tc0312: I increased the timeout to 30s. Ideally it should be a LivyConf. But this doesn't really matter since I will submit a following PR to remove this thread.

repl/src/main/scala/com/cloudera/livy/repl/Session.scala
tc0312: Realistically speaking, even with HA, no one's gonna connect thousands of livy-server to the same livy-repl. This sounds like a premature optimization.

repl/src/main/scala/com/cloudera/livy/repl/Session.scala
tc0312: We know for sure Kryo can serialize strings, but might not be able to serialize non POJOs.

repl/src/main/scala/com/cloudera/livy/repl/ReplDriver.scala
tc0312: Added synchronized around functions which will mutate `statements`.
> Also I'm not sure in which scenario do we require to bulk get the statement results?

`statements()` in InteractiveSession. Which's called by `get("/:id/statements")` in `InteractiveSessionServlet`.

jerryshao: How about the 2nd question?

tc0312:
Any feedback for iteration 2?

repl/src/main/scala/com/cloudera/livy/repl/ReplDriver.scala
```diff
@@ -73,19 +63,29 @@ class ReplDriver(conf: SparkConf, livyConf: RSCConf)
     }
   }
 
-  def handle(ctx: ChannelHandlerContext, msg: BaseProtocol.ReplJobRequest): Unit = {
-    Future {
-      jobFutures(msg.id) = session.execute(msg.code).result
-    }
+  def handle(ctx: ChannelHandlerContext, msg: BaseProtocol.ReplJobRequest): Int = synchronized {
```
jerryshao: I'm not sure here `synchronized` is required? From my understanding it is not necessary, since `session.execute` has already been synchronized.

repl/src/main/scala/com/cloudera/livy/repl/ReplDriver.scala
```diff
@@ -73,19 +63,29 @@ class ReplDriver(conf: SparkConf, livyConf: RSCConf)
...
   }
 
-  def handle(ctx: ChannelHandlerContext, msg: BaseProtocol.GetReplJobResult): String = {
-    val result = jobFutures.getOrElse(msg.id, null)
-    Option(result).map { r => compact(render(r)) }.orNull
+  /**
+   * Return statement results. Results are sorted by statement id.
+   */
+  def handle(ctx: ChannelHandlerContext, msg: BaseProtocol.GetReplJobResults): ReplJobResults =
+    synchronized {
```
jerryshao: Noted that this `synchronized` has a different locker compared to `session.synchronized`, this actually locks nothing.

repl/src/main/scala/com/cloudera/livy/repl/ReplDriver.scala
tc0312: Yes because `def handle(ctx: ChannelHandlerContext, msg: BaseProtocol.GetReplJobResults)` is locking this class too.

repl/src/main/scala/com/cloudera/livy/repl/ReplDriver.scala
tc0312: If we don't get rid of the lock in `def handle(ctx: ChannelHandlerContext, msg: BaseProtocol.ReplJobRequest)` then it will locks something.

jerryshao: Sorry I cannot get your point, can you please elaborate more?

tc0312: I think I have misunderstood your comment. Can you explain which line is doing session.synchronized?

repl/src/main/scala/com/cloudera/livy/repl/ReplDriver.scala
tc0312: Fixed but keeping session.synchronized here for defensive purpose.

PR #215, opened 2016-10-28
LIVY-236. Fix NPE issue in Job API when return value is Void

jerryshao:
to reproduce this issue:

```
val job = new Job[Void] {
      override def call(jc: JobContext): Void = {
        jc.sc().map(xxx).saveAsTextFile(xxx)
        null
      }
    }

    val result = waitFor(client2.submit(job))
```

We should also consider return value explicitly set as Void not to throw exception.


tc0312:
LGTM

PR #216, opened 2016-10-31
LIVY-219. Follow-up. Some follow-up works for Cross integration test

jerryshao:
1. Update the pom to restrict the scope and remove some unnecessary dependencies.
2. Rewrite `ScalaEcho` test with Java one. Because test-lib will be run under Scala 2.10 and 2.11 environments, while it is built against with Scala 2.10, to avoid potential problems, propose to rewrite with Java.

### DIFF
--- a/integration-test/pom.xml
+++ b/integration-test/pom.xml
@@ -167,31 +167,25 @@
     <dependency>
       <groupId>org.apache.spark</groupId>
       <artifactId>spark-core_${scala.binary.version}</artifactId>
+      <scope>test</scope>
     </dependency>
 
     <dependency>
       <groupId>org.apache.spark</groupId>
       <artifactId>spark-sql_${scala.binary.version}</artifactId>
+      <scope>test</scope>
     </dependency>
 
     <dependency>
       <groupId>org.apache.spark</groupId>
       <artifactId>spark-streaming_${scala.binary.version}</artifactId>
+      <scope>test</scope>
     </dependency>
 
     <dependency>
       <groupId>org.apache.spark</groupId>
       <artifactId>spark-hive_${scala.binary.version}</artifactId>
-    </dependency>
-
-    <dependency>
-      <groupId>org.apache.spark</groupId>
-      <artifactId>spark-repl_${scala.binary.version}</artifactId>
-    </dependency>
-
-    <dependency>
-      <groupId>org.apache.spark</groupId>
-      <artifactId>spark-yarn_${scala.binary.version}</artifactId>
+      <scope>test</scope>
     </dependency>
 
     <dependency>
--- a/integration-test/src/test/scala/com/cloudera/livy/test/JobApiIT.scala
+++ b/integration-test/src/test/scala/com/cloudera/livy/test/JobApiIT.scala
@@ -22,11 +22,11 @@ import java.io.{File, InputStream}
 import java.net.URI
 import java.nio.charset.StandardCharsets.UTF_8
 import java.nio.file.{Files, StandardCopyOption}
-import java.util.concurrent.{Future => JFuture, TimeUnit}
+import java.util.concurrent.{TimeUnit, Future => JFuture}
 import javax.servlet.http.HttpServletResponse
 
 import scala.collection.JavaConverters._
-import scala.util.Try
+import scala.util.{Properties, Try}
 
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.fasterxml.jackson.module.scala.DefaultScalaModule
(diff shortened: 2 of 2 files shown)


jerryshao:
\cc @tc0312 please help to review, thanks a lot.

integration-test/pom.xml
```diff
@@ -167,31 +167,25 @@
     <dependency>
       <groupId>org.apache.spark</groupId>
       <artifactId>spark-core_${scala.binary.version}</artifactId>
+      <scope>test</scope>
```
tc0312: I think these dependencies are used by code under src, their scope shoudln't be test.

jerryshao: Can you point out which code requires spark-core, AFAIK the only requirement is JobAPI tests.

integration-test/src/test/scala/com/cloudera/livy/test/JobApiIT.scala
```diff
@@ -157,16 +157,17 @@ class JobApiIT extends BaseIntegrationTestSuite with BeforeAndAfterAll with Logg
     assert(result === "hello")
   }
 
-  test("run scala jobs") {
+  test("run echo jobs") {
```
tc0312: Does this test case provide additional test coverage over  `run spark job` and `run spark sql job`.

integration-test/pom.xml
tc0312: Oops you are right. Those ITs are not supposed to be under `test/`.
There's a pending work item to generate a livy-intergration test jar to run Livy IT without Maven.

integration-test/src/test/scala/com/cloudera/livy/test/JobApiIT.scala
jerryshao: Actually it is not necessary to test Scala, because we don't use livy-scala-api, also as I mentioned, involve Scala code in test-lib will potentially introduce issue.

tc0312: You are right. Let me rephrase your words a bit. There are 2 problems:

1. IT doesn't use `livy-scala-api`.
2. test-lib is not built with Scala 2.10 and 2.11.

I think the correct fix here is changing `livy-test-lib` to build with Scala 2.10 and 2.11.

jerryshao: That really increase the pom modules, I thought your suggestion before (build two versions), but just for one test may not be so necessary. Also I'm not sure what actually tested by using Scala code + Java API client.

tc0312: Do you think it's a good idea to merge `minicluster-dependencies` with `test-lib`?

jerryshao: @tc0312 One thing I may ignored, since IT is built against scala 2.10 and which will add test-lib as dependency, so even if we built two versions of test-lib, we could still use test-lib_2.10, unless we change IT to build against 2.11. So I think it is not necessary to build two versions of test-lib.

tc0312: test-lib is picked up by code but not by classpath [Code](https://github.com/cloudera/livy/blob/master/integration-test/src/main/scala/com/cloudera/livy/test/framework/MiniCluster.scala#L227).
We can change it to use sparkScalaVersion to solve this problem.

jerryshao: JobAPIIT explicitly depends on test-lib for some jobs in the compile time , it is not enough to only pick in the runtime .

tc0312: You are right. I missed that.
Besides ScalaEcho, all other test jobs are Java classes. I think we can load them from test-lib-2.10.jar.
For ScalaEcho test, I don't have a solution yet... Let me sleep on it and get back to you.

jerryshao:
@tc0312 , any further thoughts on this? From my understanding, unless `integration-test` code can be built against 2.11, otherwise we don't have better solution to test scala JobAPI in 2.11 environment.

tc0312:
I don't have a good solution for the problem. Let's test `scala-api` in just Scala 2.10 for now.

jerryshao:
Thanks @tc0312 , so regarding this issue, since we don't have better solution to test different scala version code for jobApiIT, I would suggest to fix like this way to avoid potential issues when testing different with scalas. What do you think?

tc0312:
I think may be it's better to keep ScalaEcho in Scala because SQLGetTweets is in Java.
I think we can cancel the test when LIVY_SPARK_SCALA_VERSION is set to 2.11. This way we can keep current test coverage and avoid potential Scala incompatible issues.

jerryshao:
Thanks @tc0312 , I'm OK with it. Let me update the code.

jerryshao:
@tc0312 would you please take chance to review again, thanks a lot.

PR #217, opened 2016-11-01
LIVY-233. RFC. SparkSession support for Job API

jerryshao:
Current proposal is to use generic type for SparkSession, the way to use it like:

```java
    JobHandle<String> handler = client.submit(new Job<String>() {
      @Override
      public String call(JobContext jc) throws Exception {
        SparkSession session = jc.sparksession();
        return session.version();
      }
    });
```

And it is similar for Scala API. If `SparkSession` is not supported, it will throw exception instead.

Please suggest, greatly appreciate your comments.


jerryshao:
@zjffdu @tc0312 greatly appreciate your inputs, thanks.

tc0312:
LGTM as the first iteration. If users ask for a easier way to access sparksession, we can provide a shim later that provide strong typed SparkSession object in a separate jar.

api/src/main/java/com/cloudera/livy/JobContext.java
```diff
@@ -63,4 +63,8 @@
    */
   File getLocalTmpDir();
 
+  /**
+   * Returns SparkSession if it existed, otherwise throws Exception.
+   */
+  <E> E sparksession() throws Exception;
```
tc0312: Nit: sparkSession

tc0312:
@zjffdu @ksakellis @vanzin

tc0312:
LGTM. Can we add integration test?

jerryshao:
It may not be easy to add a test for Spark 2.0 specific things, since no module is currently built against Spark 2.0.

tc0312:
I think we should make Spark 2.0 specific test possible.

tc0312:
Hi Jerry, I prototyped the build rule for adding Spark 2 specific test:
https://github.com/hdinsight/livy/commit/a91bf012e1aa8aed11b950ca78065aed2096ba90
I also added a example test for interactive session. I think it's not difficult to add Spark 2 specific test.

jerryshao:
Thanks @tc0312 , I checked your implementation, it may not difficult to add support of Spark2 test for interactive session, but I'm not sure if your change also apply to Job API, because Job API Strong typed, which means we should have Spark2 dependencies explicitly in the test environment in compile time.

tc0312:
> which means we should have Spark2 dependencies explicitly in the test environment in compile time.
You are right but it's already true when we build Livy with profile `spark-2.0`. Profile `spark-2.0` sets `spark.version` to 2.0.

My change adds test files to source path according to Spark's version. Tests under `spark-2.0/` directory are built only when Livy's building with Spark 2.0.  They can safely reference to Spark 2.0 classes and will not break Spark 1.6 build.

jerryshao:
@tc0312 thanks a lot for your suggestion. I changed the code to add Spark2 Job API integration test for this PR. I think with this patch merged, later on we could add more interactive and batch test for Spark2 specifically.

spark2-test-lib/pom.xml
```diff
@@ -0,0 +1,87 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
tc0312: I think we can do this without adding a new package. We can add this to test-lib and re-use the Spark version depending build rule. What do you think?

jerryshao: Agree. Done.

test-lib/pom.xml
```diff
@@ -88,4 +88,40 @@
     </plugins>
   </build>
 
+  <profiles>
```
tc0312: Instead of relying on profile, mind if you follow what [integration-test is doing](https://github.com/cloudera/livy/pull/217/files#diff-652e444a472b79b385022ec7dd221790R213)?

jerryshao: done.

tc0312:
LGTM. Merging if there's no further comment.

PR #218, opened 2016-11-02
LIVY-237. Fixed JaCoCo is randomly throwing EOFException on Travis.

tc0312:
- Disabled JaCoCo for `InteractiveIT.should kill RSCDriver if it doesn't respond to end session`. This test doesn't shutdown RSCDriver's JVM gracefully. JaCoCo might generate a corrupted exec dump and cause EOFException.


tc0312:
@jerryshao Thanks!

PR #219, opened 2016-11-02
Minor. Fix build error and add some missing dependencies for report aggregat…

jerryshao:
…ion.

### DIFF
--- a/coverage/pom.xml
+++ b/coverage/pom.xml
@@ -53,13 +53,25 @@
 
     <dependency>
       <groupId>${project.groupId}</groupId>
-      <artifactId>livy-core_${scala.binary.version}</artifactId>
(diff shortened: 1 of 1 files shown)


tc0312:
Sorry I missed these in my change.
It isn't clear in the PR. What build error does this fix?

coverage/pom.xml
```diff
@@ -77,7 +89,13 @@
 
     <dependency>
       <groupId>${project.groupId}</groupId>
-      <artifactId>livy-scala-api</artifactId>
```
jerryshao: This is the build error, since currently livy-scala-api module is not existed, so it will be failed. It occurs in our in house build, but I'm not why it is not happened in travis build. @tc0312

PR #220, opened 2016-11-03
LIVY-239: Moving the logic to generate session IDs from Session Manager to SessionStore

meisam:
Summary of changes:

- Removes the `AtomicCounter` that is used in `SessionManager` to generate session IDs.
- Generates the session IDs in `SessionStore` instead of SessionManager
- Writes the newly generated session ID to the store each time `nextSessionId` is called.

Task-url: https://issues.cloudera.org/browse/LIVY-239


server/src/main/scala/com/cloudera/livy/server/recovery/SessionStore.scala
```diff
@@ -69,14 +65,17 @@ class SessionStore(
...
    *
    * @throws Exception If SessionManagerState stored is corrupted, it throws an error.
    */
-  def getNextSessionId(sessionType: String): Int = {
-    store.get[SessionManagerState](sessionManagerPath(sessionType))
+  def getNextSessionId(sessionType: String): Int = synchronized {
+    val nextSessionId = store.get[SessionManagerState](sessionManagerPath(sessionType))
       .map(_.nextSessionId).getOrElse(0)
+    store.set(sessionManagerPath(sessionType), SessionManagerState(nextSessionId + 1))
+    nextSessionId
```
jerryshao: IIUC, the `nextSessionId` you get is the previous one persisted and you didn't increase the number when returned.

Also each time when you get the session id, you will fetch it from persisted storage, I think it is not necessary and will bring in extra overhead.

tc0312:
Is this done for implementing HA?

meisam:
@tc0312 Yes. This change is needed to implement multimode HA.

server/src/main/scala/com/cloudera/livy/server/recovery/SessionStore.scala
meisam: @jerryshao The scala doc comments are a bit misleading. The value stored in session store is the next unused session ID.
Performance wise, this change calls the session store as many times as the previous code. The only difference is that previously `SessionManager` was calling `SessionStore.set`, but now the `SessionStore` calls it. Please take a look at lines 81 to 84 in https://github.com/cloudera/livy/blob/5e8474e32d81d794d2d7eca7cfa7c81c9be689aa/server/src/main/scala/com/cloudera/livy/sessions/SessionManager.scala#L81

server/src/main/scala/com/cloudera/livy/server/recovery/SessionStore.scala
```diff
@@ -69,14 +65,17 @@ class SessionStore(
 
   /**
    * Return the next unused session id with specified session type.
-   * If checks the SessionManagerState stored and returns the next free session id.
+   * It checks the SessionManagerState stored and returns the next free session id.
    * If no SessionManagerState is stored, it returns 0.
+    * It saves the new session ID to the session store.
```
tc0312: Nit: Extra space here.

server/src/main/scala/com/cloudera/livy/server/recovery/SessionStore.scala
tc0312: Sorry about the poor doc. My bad. Would you mind fixing it?
I think fixing the doc and rename `val nextSessionId = store.get` to lastUsedSessionId?

server/src/main/scala/com/cloudera/livy/server/recovery/SessionStore.scala
```diff
@@ -69,14 +65,17 @@ class SessionStore(
...
    *
    * @throws Exception If SessionManagerState stored is corrupted, it throws an error.
    */
-  def getNextSessionId(sessionType: String): Int = {
-    store.get[SessionManagerState](sessionManagerPath(sessionType))
+  def getNextSessionId(sessionType: String): Int = synchronized {
+    val nextSessionId = store.get[SessionManagerState](sessionManagerPath(sessionType))
```
tc0312: Reviewing this with HA in mind. I think it's possible that another livy-server instance updates SessionManagerState in between the `store.get()` and `store.set()` call.
I think we should add locking/optimistic locking + retry here to avoid servers stepping on each other. Or better, we change the interface of StateStore to make use of Sequence Nodes in ZooKeeper.

server/src/main/scala/com/cloudera/livy/sessions/SessionManager.scala
```diff
@@ -78,11 +77,8 @@ class SessionManager[S <: Session, R <: RecoveryMetadata : ClassTag](
   mockSessions.getOrElse(recover()).foreach(register)
   new GarbageCollector().start()
 
-  def nextId(): Int = synchronized {
-    val id = idCounter.getAndIncrement()
-    sessionStore.saveNextSessionId(sessionType, idCounter.get())
-    id
-  }
+  // sessionStore.getNextSessionId is guaranteed to return  atomic and returns unique IDs.
```
tc0312: Nit: Extra space between return & atomic.

server/src/main/scala/com/cloudera/livy/server/recovery/SessionStore.scala
meisam: I have to correct myself. Prior to this commit, there was one read from `SessionStore` on recovery or on server start. There was also one write per `nextId()` call. Now we have one read and one write to `SessionStore` per `nextId()` call.

server/src/main/scala/com/cloudera/livy/server/recovery/SessionStore.scala
meisam: That's right! But for multi-node HA, I am overriding this method in ZooKeeperStore and I am using a distributed atomic long recipe (https://curator.apache.org/curator-recipes/distributed-atomic-long.html) to guarantee that generated session IDs are unique and are generated and stored atomically. But for single node HA synchronizing on the session store instance would suffice.

tc0312: With your ZooKeeperStore change, who's doing nextSessionId + 1?

server/src/main/scala/com/cloudera/livy/server/recovery/SessionStore.scala
tc0312: I think this's necessary for HA consistency.
Can we come up with some smart way to not pay this price for recovery mode, but just for HA mode?

server/src/main/scala/com/cloudera/livy/server/recovery/SessionStore.scala
meisam: `SessionStore` calls into the distributed Atomic recipe from Apache curator.
This is the how I am doing it:

```scala
  val distributedSessionId = new DistributedAtomicLong(curatorClient, zkPath, retryPolicy)

  def nextBatchSessionId: Int = {
    recursiveTry(distributedSessionId, MAX_RETRY) match {
      case Some(sessionId) =>
        sessionId.toInt
      case None =>
        val msg: String = "Failed to get the next session id from Zookeeper"
        logger.warn(msg)
        throw new IOException(msg)
    }
  }

  @tailrec
  private def recursiveTry(distributedLong: DistributedLong, retryCount: Int): Option[Long] = {
    val updatedValue = distributedLong.increment
    updatedValue.succeeded match {
      case _ if retryCount <= 0 =>
        None
      case true if retryCount > 0 =>
        Option(updatedValue.preValue())
      case _ =>
        recursiveTry(distributedLong, retryCount - 1)
    }
  }
```

tc0312: Are these codes in SessionStore or a subclass of SessionStore?

server/src/main/scala/com/cloudera/livy/server/recovery/SessionStore.scala
meisam: This is my initial thoughts on it.
We can refactor the logic to generate sessionIDs into a separate trait/class and have two implementation: one for single-node, and one for multi-node. The single node mode can simply use a local atomic value, but the multi-node mode should use a more sophisticated implementation to guarantee uniqueness and consistency across all nodes.

server/src/main/scala/com/cloudera/livy/server/recovery/SessionStore.scala
meisam: The above code is in `ZookeeperStore`.

server/src/main/scala/com/cloudera/livy/server/recovery/SessionStore.scala
tc0312: I think we are thinking similarly but have 1 difference. You are adding the session id generation to SessionStore while I'm thinking about adding it to StateStore.

I would prefer to keep HA logic out of SessionStore and move these logic into ZookeeperStateStore. Having a single code path in SessionStore makes isolating HA related changes easier.

server/src/main/scala/com/cloudera/livy/server/recovery/SessionStore.scala
tc0312: Do you mean `ZooKeeperStateStore`?

tc0312: Did you add a new method in StateStore to generate unique id?

meisam: Yes I did. But as you said in your comment, moving the logic out of `SessionStore` makes a lot more sense.

> I would prefer to keep HA logic out of SessionStore and move these logic into ZookeeperStateStore. Having a single code path in SessionStore makes isolating HA related changes easier.

jerryshao:
Can we have a whole plan of Livy HA before doing this small refactoring work? My concern is that currently we don't have a decided plan of Livy HA, if finally the plan is changed or updated, current changes may be obsolete.

meisam:
@jerryshao  There is a JIRA ticket open for it: https://issues.cloudera.org/browse/LIVY-231
Desing doc: https://github.com/meisam/livy/wiki/Multi-node-HA-for-batch-sessions
preliminary pull request for LIVY-231: https://github.com/cloudera/livy/pull/222

PR #221, opened 2016-11-04
LIVY-232. Remove Max Spark version cap in LivyServer

jerryshao:
With this change, if Spark version detected from `spark-submit` is greater than the specified one, Livy will log warning instead of throwing the exception. Also change the Scala version check with the same way.


tc0312:
LGTM

PR #224, opened 2016-11-07
LIVY-240. Aggressively clean the obsolete session metadata

jerryshao:
Current `SessionManager` will only clean the session in a timely triggered GC mechanism, and the default timeout is 1hr, which means the session metadata will be in memory for 1hr, this is OK without recovery enabled.

But with session recovery enabled, stored session data will also only be cleaned out after timeout, this will unnecessarily prolong the metadata persisted in the external storage. If Livy server is failed in this time period, though session is already inactive, recovered Livy server will recover and mark as inactive again, and the TTL will again set to 1hr. To the extreme, if Livy server always fails within this time period, the persisted metadata will never get a chance to clean up.

So here propose to aggressively clean the metadata. For those obsolete sessions, no matter the TTL is reached or not, metadata should be cleaned up once gc is triggered .

### DIFF
--- a/server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
+++ b/server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
@@ -346,10 +346,8 @@ class InteractiveSession(
   private val operationCounter = new AtomicLong(0)
   private var rscDriverUri: Option[URI] = None
   private var sessionLog: IndexedSeq[String] = IndexedSeq.empty
-  private val sessionSaveLock = new Object()
 
   _appId = appIdHint
-  sessionStore.save(RECOVERY_SESSION_TYPE, recoveryMetadata)
 
   private val app = mockApp.orElse {
     if (livyConf.isRunningOnYarn()) {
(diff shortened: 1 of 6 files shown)


meisam:
@jerryshao What would happen to applications that run for days/months without any interaction with Livy? E.g. if a user submits a steaming application as a Livy batch session.

jerryshao:
@meisam Sorry I'm not very sure about your question. If a user submitted a streaming application through batch session, then Livy will launch the application for this user and poll the status of this application is running in yarn mode. The session will be alive until the application is killed or finished.

tc0312:
Jeff created a [LIVY-168](https://issues.cloudera.org/browse/LIVY-168) a while back to remove GC for batch.

jerryshao:
Thanks @tc0312 for pointing out this JIRA. I think GC for batch is still necessary, but could do when session is succeed or failure, no need to wait for session timeout. If there's no GC mechanism and user didn't delete the session explicitly, this batch session will be lingered.

The import issue is that batch session doesn't record activities, so it will be killed after timeout even it is still running (like streaming application), so we should not use activities to track the batch session. What do you think?

tc0312:
After reading your comment, I feel that we should do GC the same way for both batch and interactive sessions.
GC should delete sessions (dead/succeed/killed) terminated a while ago. To workaround the recovery problem, we should persist the termination time to the state store.
However, for running(batch) idle, busy, error(interactive), they should never be GC-ed.

jerryshao:
Close this in favor of #277 .

PR #226, opened 2016-11-09
LIVY-241. Return HTTP 400 Bad Request on bad user input.

linchan-ms:
LIVY-241.


tc0312:
Mind if you add these to [here](https://github.com/cloudera/livy/blob/master/server/src/main/scala/com/cloudera/livy/server/JsonServlet.scala#L59) instead?

Please also add unit test to JsonServletSpec.

PR #227, opened 2016-11-10
LIVY-242. Shorten the time to get RSC driver URI

jerryshao:
Currently RSC driver URI is gotten when `PingJob` is finished, which means only when SparkContext and other Spark entry point are fully started and job are finished we could get the RSC driver URI. This unnecessarily prolong the intermediate state of recovery metadata, if LivyServer is failed at this time period, it will never talk the RSC application again.

So we should shorten the time to get RSC driver URI ASAP to mitigate the intermediate state of recovery metadata, that will make session recovery more robust.

### DIFF
--- a/server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
+++ b/server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
@@ -371,8 +371,6 @@ class InteractiveSession(
   } else {
     val uriFuture = future { client.get.getServerUri.get() }
 
-    info(s"current time: ${System.currentTimeMillis()}")
-
     uriFuture onSuccess { case url =>
       rscDriverUri = Option(url)
       sessionSaveLock.synchronized {
@@ -382,8 +380,6 @@ class InteractiveSession(
     uriFuture onFailure { case e =>
       throw new IllegalStateException("Fail to get rsc uri", e)
     }
-
-    info(s"current time: ${System.currentTimeMillis()}")
 
     // Send a dummy job that will return once the client is ready to be used, and set the
     // state to "idle" at that point.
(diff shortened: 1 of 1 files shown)


server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
```diff
@@ -369,6 +369,18 @@ class InteractiveSession(
...
+    val uriFuture = future { client.get.getServerUri.get() }
+
+    uriFuture onSuccess { case url =>
+      rscDriverUri = Option(url)
+      sessionSaveLock.synchronized {
+        sessionStore.save(RECOVERY_SESSION_TYPE, recoveryMetadata)
+      }
+    }
+    uriFuture onFailure { case e =>
+      throw new IllegalStateException("Fail to get rsc uri", e)
```
tc0312: Instead of throwing an exception here, I think we can print a warning. If the RSC Driver is indeed dead, PingJob will fail and its listener will stop the session properly.

tc0312:
Good fix. Thanks!

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
jerryshao: Sure, I will change it.

tc0312:
LGTM.

meisam:
I am getting a build error after this PR merge
```
[error] livy/server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala:384: not found: value future
[error]     val uriFuture = future { client.get.getServerUri.get() }
[error]                     ^
[error] one error found
[error] Compile failed at Nov 11, 2016 10:51:54 AM [1.009s]
```

jerryshao:
@meisam it is weird travis didn't complain this error, also my local build is fine. Can you please make a clean build again?

meisam:
@jerryshao It is already fixed on the master branch. One of the imports was omitted during merge.

PR #228, opened 2016-11-10
LIVY-243. ZooKeeperStateStore ignores settings for RetryPolicy

meisam:
Use the retry times value the sleep time value specified in the config
to create the retry policy for the curator client.

Also elevates `retryPolicy` from a local val into a field that can be
used in other places in the same class.

Task-url: https://issues.cloudera.org/browse/LIVY-243

### DIFF
--- a/server/src/test/scala/com/cloudera/livy/server/recovery/ZooKeeperStateStoreSpec.scala
+++ b/server/src/test/scala/com/cloudera/livy/server/recovery/ZooKeeperStateStoreSpec.scala
@@ -96,6 +96,16 @@ class ZooKeeperStateStoreSpec extends FunSpec with LivyBaseUnitTestSuite {
       }
     }
 
+    it("get should retrieve retry policy configs") {
+      conf.set(com.cloudera.livy.server.recovery.ZooKeeperStateStore.ZK_RETRY_CONF, "11,77")
+        withMock { f =>
+        mockExistsBuilder(f.curatorClient, true)
+
+        f.stateStore.retryPolicy should not be null
+        f.stateStore.retryPolicy.getN shouldBe 11
+      }
+    }
+
     it("get should retrieve data from curatorClient") {
       withMock { f =>
         mockExistsBuilder(f.curatorClient, true)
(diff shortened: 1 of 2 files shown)


tc0312:
My bad. I didn't write a unit test for it. Thanks so much for the fix.
Would you mind writing that unit test? I apologize for my double standard.

meisam:
I added the test case.

server/src/test/scala/com/cloudera/livy/server/recovery/ZooKeeperStateStoreSpec.scala
```diff
@@ -96,6 +96,21 @@ class ZooKeeperStateStoreSpec extends FunSpec with LivyBaseUnitTestSuite {
...
     }
 
+    it("get should retrieve retry policy configs") {
+      withMock { f =>
+        mockExistsBuilder(f.curatorClient, true)
+
+        val getDataBuilder = mock[GetDataBuilder]
+
+        assert(f.stateStore.retryPolicy != null)
+        assert(f.stateStore.retryPolicy.getN === 5) // default value is 5 retries with 100ms sleeps
```
tc0312: Mind if you change the retry value in `conf` to a value different from the default at the top of the test, then verify against it here?

server/src/test/scala/com/cloudera/livy/server/recovery/ZooKeeperStateStoreSpec.scala
```diff
@@ -96,6 +96,21 @@ class ZooKeeperStateStoreSpec extends FunSpec with LivyBaseUnitTestSuite {
       }
     }
 
+    it("get should retrieve retry policy configs") {
+      withMock { f =>
+        mockExistsBuilder(f.curatorClient, true)
+
+        val getDataBuilder = mock[GetDataBuilder]
+
+        assert(f.stateStore.retryPolicy != null)
```
tc0312: Instead of assert, mind if you use `should`? When the check fails, `should` provides clear error messages.

`f.stateStore.retryPolicy should not be null`

server/src/test/scala/com/cloudera/livy/server/recovery/ZooKeeperStateStoreSpec.scala
tc0312: Also, please use `should`
f.stateStore.retryPolicy.getN shouldBe <...>

PR #230, opened 2016-11-14
LIVY-215. Support statement canceling for repl session

jerryshao:
This work is based on @zjffdu 's initial proposal, with some code simplification to Support statement cancel for Spark interpreter.

Please review @tc0312 , thanks.

### DIFF
--- a/repl/src/main/scala/com/cloudera/livy/repl/ProcessInterpreter.scala
+++ b/repl/src/main/scala/com/cloudera/livy/repl/ProcessInterpreter.scala
@@ -50,10 +50,7 @@ abstract class ProcessInterpreter(process: Process)
   override def start(): SparkContext = {
     waitUntilReady()
 
-    // At this point there should be an already active SparkContext that can be retrieved
-    // using SparkContext.getOrCreate. But we don't really support running "pre-compiled"
-    // jobs against pyspark or sparkr, so just return null here.
-    null
+    SparkContext.getOrCreate()
   }
 
   override def execute(code: String): Interpreter.ExecuteResponse = {
--- a/repl/src/main/scala/com/cloudera/livy/repl/ProcessInterpreter.scala
+++ b/repl/src/main/scala/com/cloudera/livy/repl/ProcessInterpreter.scala
@@ -28,7 +28,6 @@ import org.apache.spark.SparkContext
 import org.json4s.JValue
 
 import com.cloudera.livy.{Logging, Utils}
-import com.cloudera.livy.repl.Interpreter.ExecuteResponse
 
 private sealed trait Request
 private case class ExecuteRequest(code: String, promise: Promise[JValue]) extends Request
@@ -48,12 +47,16 @@ abstract class ProcessInterpreter(process: Process)
   protected[this] val stdin = new PrintWriter(process.getOutputStream)
   protected[this] val stdout = new BufferedReader(new InputStreamReader(process.getInputStream), 1)
 
-  override def internalStart(): SparkContext = {
+  override def start(): SparkContext = {
     waitUntilReady()
-    SparkContext.getOrCreate
+
+    // At this point there should be an already active SparkContext that can be retrieved
+    // using SparkContext.getOrCreate. But we don't really support running "pre-compiled"
+    // jobs against pyspark or sparkr, so just return null here.
+    null
   }
 
-  override def internalExecute(code: String): ExecuteResponse = {
+  override def execute(code: String): Interpreter.ExecuteResponse = {
     try {
       sendExecuteRequest(code)
     } catch {
--- a/repl/src/main/scala/com/cloudera/livy/repl/ProcessInterpreter.scala
+++ b/repl/src/main/scala/com/cloudera/livy/repl/ProcessInterpreter.scala
@@ -28,6 +28,7 @@ import org.apache.spark.SparkContext
 import org.json4s.JValue
 
 import com.cloudera.livy.{Logging, Utils}
+import com.cloudera.livy.repl.Interpreter.ExecuteResponse
 
 private sealed trait Request
 private case class ExecuteRequest(code: String, promise: Promise[JValue]) extends Request
@@ -47,15 +48,12 @@ abstract class ProcessInterpreter(process: Process)
   protected[this] val stdin = new PrintWriter(process.getOutputStream)
   protected[this] val stdout = new BufferedReader(new InputStreamReader(process.getInputStream), 1)
 
-  override def start(): SparkContext = {
+  override def internalStart(): SparkContext = {
     waitUntilReady()
-
-    // At this point there should be an already active SparkContext that can be retrieved
-    // using SparkContext.getOrCreate. But we don't really support running "pre-compiled"
-    // jobs against pyspark or sparkr, so just return null here.
-    null
+    SparkContext.getOrCreate
   }
-  override def execute(code: String): Interpreter.ExecuteResponse = {
+
+  override def internalExecute(code: String): ExecuteResponse = {
     try {
       sendExecuteRequest(code)
     } catch {
--- a/repl/src/main/scala/com/cloudera/livy/repl/Session.scala
+++ b/repl/src/main/scala/com/cloudera/livy/repl/Session.scala
@@ -44,14 +44,16 @@ object Session {
   val TRACEBACK = "traceback"
 }
 
-class Session(interpreter: Interpreter, stateChangedCallback: SessionState => Unit = { _ => } )
-  extends Logging
-{
+class Session(interpreter: Interpreter, stateChangedCallback: SessionState => Unit = { _ => })
+  extends Logging {
   import Session._
 
   private implicit val executor = ExecutionContext.fromExecutorService(
     Executors.newSingleThreadExecutor())
+
   private implicit val formats = DefaultFormats
+
+  @volatile private var _sc: Option[SparkContext] = None
 
   private var _state: SessionState = SessionState.NotStarted()
   private val _statements = TrieMap[Int, Statement]()
@@ -67,9 +69,12 @@ class Session(interpreter: Interpreter, stateChangedCallback: SessionState => Un
       changeState(SessionState.Idle())
       sc
     }
+
     future.onFailure { case _ =>
       changeState(SessionState.Error())
     }
+    future.onSuccess { case sc => _sc = Option(sc) }
+
     future
   }
 
@@ -91,8 +96,17 @@ class Session(interpreter: Interpreter, stateChangedCallback: SessionState => Un
     statementId
   }
 
-  def cancel(statementId: String): Unit = {
-    interpreter.cancel(statementId)
+  def cancel(statementId: Int): Unit = {
+    val stmt = synchronized {
+      _statements.remove(statementId)
+    }
+
+    stmt.foreach { st =>
+      info(s"Statement $statementId is canceled")
+      if (st.state == StatementState.Running) {
+        _sc.foreach(_.cancelJobGroup(statementId.toString))
+      }
+    }
   }
 
   def close(): Unit = {
--- a/repl/src/main/scala/com/cloudera/livy/repl/Session.scala
+++ b/repl/src/main/scala/com/cloudera/livy/repl/Session.scala
@@ -88,14 +88,17 @@ class Session(interpreter: Interpreter, stateChangedCallback: SessionState => Un
     Future {
       _statements(statementId).state.compareAndSet(StatementState.Waiting, StatementState.Running)
 
-      val executeResult = if (_statements(statementId).state.get() != StatementState.Cancelled) {
-        executeWithJobGroup(statementId, code)
+      setJobGroup(statementId)
+      val executeResult = if (_statements(statementId).state.get() != StatementState.Cancelling) {
+        executeCode(statementId, code)
       } else {
         null
       }
 
       _statements(statementId).output = executeResult
       _statements(statementId).state.compareAndSet(StatementState.Running, StatementState.Available)
+      _statements(statementId).state.compareAndSet(
+        StatementState.Cancelling, StatementState.Cancelled)
     }
 
     statementId
@@ -103,12 +106,14 @@ class Session(interpreter: Interpreter, stateChangedCallback: SessionState => Un
 
   def cancel(statementId: Int): Unit = {
     if (_statements.contains(statementId)) {
-      val oldStmtState = _statements(statementId).state.getAndSet(StatementState.Cancelled)
+      val oldStmtState = _statements(statementId).state.getAndSet(StatementState.Cancelling)
 
       info(s"Statement $statementId is canceled")
 
       if (oldStmtState == StatementState.Running) {
         _sc.foreach(_.cancelJobGroup(statementId.toString))
+      } else if (oldStmtState == StatementState.Available) {
+        _statements(statementId).state.set(StatementState.Cancelled)
       }
     }
   }
@@ -191,23 +196,20 @@ class Session(interpreter: Interpreter, stateChangedCallback: SessionState => Un
     compact(render(resultInJson))
   }
 
-  private def executeWithJobGroup(executionCount: Int, code: String): String = {
+  private def setJobGroup(statementId: Int): String = {
     val cmd = Kind(interpreter.kind) match {
       case Spark() | PySpark() | PySpark3() =>
-        s"""sc.setJobGroup($executionCount, "Job group for statement $executionCount")"""
+        s"""sc.setJobGroup("$statementId", "Job group for statement $statementId")"""
       case SparkR() =>
         interpreter.asInstanceOf[SparkRInterpreter].sparkMajorVersion match {
           case "1" =>
-            s"""setJobGroup(sc, $executionCount, "Job group for statement $executionCount", """ +
+            s"""setJobGroup(sc, "$statementId", "Job group for statement $statementId", """ +
               "FALSE)"
           case "2" =>
-            s"""setJobGroup($executionCount, "Job group for statement $executionCount", FALSE)"""
+            s"""setJobGroup("$statementId", "Job group for statement $statementId", FALSE)"""
         }
     }
     // Set the job group
-    executeCode(executionCount, cmd)
-
-    // execute the statement
-    executeCode(executionCount, code)
+    executeCode(statementId, cmd)
   }
 }
--- a/repl/src/main/scala/com/cloudera/livy/repl/Session.scala
+++ b/repl/src/main/scala/com/cloudera/livy/repl/Session.scala
@@ -18,11 +18,13 @@
 
 package com.cloudera.livy.repl
 
-import java.util.concurrent.{Executors, TimeUnit}
+import java.util.concurrent.Executors
 import java.util.concurrent.atomic.AtomicInteger
 
 import scala.collection.concurrent.TrieMap
 import scala.concurrent.{ExecutionContext, Future}
+import scala.concurrent.duration._
+
 import org.apache.spark.SparkContext
 import org.json4s.jackson.JsonMethods.{compact, render}
 import org.json4s.DefaultFormats
@@ -111,33 +113,41 @@ class Session(
   }
 
   def cancel(statementId: Int): Unit = {
-    if (_statements.contains(statementId)) {
-      val oldStmtState = _statements(statementId).state.getAndSet(StatementState.Cancelling)
+    if (!_statements.contains(statementId)) {
+      return
+    }
 
-      info(s"Statement $statementId is cancelling")
+    if (_statements(statementId).state.get() == StatementState.Available ||
+      _statements(statementId).state.get() == StatementState.Cancelled ||
+      _statements(statementId).state.get() == StatementState.Cancelling) {
+      return
+    } else {
+      _statements(statementId).state.getAndSet(StatementState.Cancelling)
+    }
+
+    info(s"Cancelling statement $statementId...")
 
       Future {
-        if (oldStmtState == StatementState.Running) {
-          val currTime = System.currentTimeMillis()
-          while (_statements(statementId).state.get() != StatementState.Cancelled &&
-            System.currentTimeMillis() - currTime <= TimeUnit.SECONDS.toMillis(30)) {
+        val deadline = livyConf.getTimeAsMs(RSCConf.Entry.JOB_CANCEL_TIMEOUT).millis.fromNow
+        while (_statements(statementId).state.get() == StatementState.Cancelling) {
+          if (deadline.isOverdue()) {
+            info(s"Failed to cancel statement $statementId.")
+            _statements(statementId).state.compareAndSet(
+              StatementState.Cancelling, StatementState.Cancelled)
+          } else {
             _sc.foreach(_.cancelJobGroup(statementId.toString))
-            Thread.sleep(livyConf.getTimeAsMs(RSCConf.Entry.JOB_CANCEL_TRIGGER_INTERVAL))
           }
-        } else if (oldStmtState == StatementState.Available) {
-          _statements(statementId).state.set(StatementState.Cancelled)
+          Thread.sleep(livyConf.getTimeAsMs(RSCConf.Entry.JOB_CANCEL_TRIGGER_INTERVAL))
+        }
+        if (_statements(statementId).state.get() == StatementState.Cancelled) {
+          info(s"Statement $statementId cancelled.")
         }
       }(cancelExecutor)
-    }
   }
 
   def close(): Unit = {
     executor.shutdown()
     interpreter.close()
-  }
-
-  def clearStatements(): Unit = {
-    _statements.clear()
   }
 
   private def changeState(newState: SessionState): Unit = {
--- a/repl/src/main/scala/com/cloudera/livy/repl/Session.scala
+++ b/repl/src/main/scala/com/cloudera/livy/repl/Session.scala
@@ -70,11 +70,8 @@ class Session(interpreter: Interpreter, stateChangedCallback: SessionState => Un
       sc
     }
 
-    future.onFailure { case _ =>
-      changeState(SessionState.Error())
-    }
+    future.onFailure { case _ => changeState(SessionState.Error()) }
     future.onSuccess { case sc => _sc = Option(sc).orElse(Some(SparkContext.getOrCreate())) }
-
     future
   }
 
@@ -195,16 +192,22 @@ class Session(interpreter: Interpreter, stateChangedCallback: SessionState => Un
   }
 
   private def executeWithJobGroup(executionCount: Int, code: String): String = {
-    Kind(interpreter.kind) match {
+    val cmd = Kind(interpreter.kind) match {
       case Spark() | PySpark() | PySpark3() =>
-        val cmd = """sc.setJobGroup(%d, "Job group for statement %d")"""
-          .format(executionCount, executionCount)
-        executeCode(executionCount, cmd)
+        s"""sc.setJobGroup($executionCount, "Job group for statement $executionCount")"""
       case SparkR() =>
-        // SparkR doesn't support statement cancelling, so doesn't need to set job group
-      case _ =>
-        // Unknown interpreter type.
+        interpreter.asInstanceOf[SparkRInterpreter].sparkMajorVersion match {
+          case "1" =>
+            s"""setJobGroup(sc, $executionCount, "Job group for statement $executionCount", """ +
+              "FALSE)"
+          case "2" =>
+            s"""setJobGroup($executionCount, "Job group for statement $executionCount", FALSE)"""
+        }
     }
+    // Set the job group
+    executeCode(executionCount, cmd)
+
+    // execute the statement
     executeCode(executionCount, code)
   }
 }
--- a/repl/src/main/scala/com/cloudera/livy/repl/Session.scala
+++ b/repl/src/main/scala/com/cloudera/livy/repl/Session.scala
@@ -86,9 +86,9 @@ class Session(interpreter: Interpreter, stateChangedCallback: SessionState => Un
     _statements(statementId) = new Statement(statementId, StatementState.Waiting, null)
 
     Future {
+      setJobGroup(statementId)
       _statements(statementId).state.compareAndSet(StatementState.Waiting, StatementState.Running)
 
-      setJobGroup(statementId)
       val executeResult = if (_statements(statementId).state.get() != StatementState.Cancelling) {
         executeCode(statementId, code)
       } else {
@@ -108,13 +108,20 @@ class Session(interpreter: Interpreter, stateChangedCallback: SessionState => Un
     if (_statements.contains(statementId)) {
       val oldStmtState = _statements(statementId).state.getAndSet(StatementState.Cancelling)
 
-      info(s"Statement $statementId is canceled")
+      info(s"Statement $statementId is cancelling")
 
-      if (oldStmtState == StatementState.Running) {
-        _sc.foreach(_.cancelJobGroup(statementId.toString))
-      } else if (oldStmtState == StatementState.Available) {
-        _statements(statementId).state.set(StatementState.Cancelled)
-      }
+      val singleThreadExecutor = ExecutionContext.fromExecutorService(
+        Executors.newSingleThreadExecutor())
+      Future {
+        if (oldStmtState == StatementState.Running) {
+          while (_statements(statementId).state.get() != StatementState.Cancelled) {
+            _sc.foreach(_.cancelJobGroup(statementId.toString))
+            Thread.sleep(100)
+          }
+        } else if (oldStmtState == StatementState.Available) {
+          _statements(statementId).state.set(StatementState.Cancelled)
+        }
+      }(singleThreadExecutor)
     }
   }
 
@@ -198,7 +205,11 @@ class Session(interpreter: Interpreter, stateChangedCallback: SessionState => Un
 
   private def setJobGroup(statementId: Int): String = {
     val cmd = Kind(interpreter.kind) match {
-      case Spark() | PySpark() | PySpark3() =>
+      case Spark() =>
+        // A dummy value to avoid automatic value binding in scala REPL.
+        s"""val _livyJobGroup$statementId = sc.setJobGroup("$statementId",""" +
+          s""""Job group for statement $statementId")"""
+      case PySpark() | PySpark3() =>
         s"""sc.setJobGroup("$statementId", "Job group for statement $statementId")"""
       case SparkR() =>
         interpreter.asInstanceOf[SparkRInterpreter].sparkMajorVersion match {
--- a/repl/src/main/scala/com/cloudera/livy/repl/Session.scala
+++ b/repl/src/main/scala/com/cloudera/livy/repl/Session.scala
@@ -53,12 +53,13 @@ class Session(interpreter: Interpreter, stateChangedCallback: SessionState => Un
 
   private implicit val formats = DefaultFormats
 
-  @volatile private var _sc: Option[SparkContext] = None
+  @volatile private[repl] var _sc: Option[SparkContext] = None
 
   private var _state: SessionState = SessionState.NotStarted()
   private val _statements = TrieMap[Int, Statement]()
 
   private val newStatementId = new AtomicInteger(0)
+  private val lock = new Object()
 
   stateChangedCallback(_state)
 
@@ -73,7 +74,7 @@ class Session(interpreter: Interpreter, stateChangedCallback: SessionState => Un
     future.onFailure { case _ =>
       changeState(SessionState.Error())
     }
-    future.onSuccess { case sc => _sc = Option(sc) }
+    future.onSuccess { case sc => _sc = Option(sc).orElse(Some(SparkContext.getOrCreate())) }
 
     future
   }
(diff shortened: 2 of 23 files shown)


repl/src/main/scala/com/cloudera/livy/repl/Session.scala
```diff
@@ -82,21 +88,48 @@ class Session(interpreter: Interpreter)
...
+          _statements(statementId) = statement
+        }
       }
     }
     statementId
   }
 
+  def cancel(statementId: Int): Unit = {
+    val stmt = synchronized {
+      _statements.remove(statementId)
```
tc0312: This will cancel executed statement.
Also it leaves holes in statement history. Instead of removing the statement, I think we should change its state to "cancelled". This's more explicit to users that a statement was submitted then cancelled.

Also, we should be careful of race conditions like cancelling a statement being executing but is succeeding. To avoid locking the whole statement map `statements`, I think we could change `Statement` to a mutable object, and when we change its state, we lock it then call its setter:
```
_statements(statementId).synchronized { 
  _statements(statementId).setState(...) 
}
```
instead of 
```_statements(statementId) = new Statement(statementId, ..., null)```.

Comment (jerryshao) [inline repl/src/main/scala/com/cloudera/livy/repl/Session.scala]: I see, I will still keep it in memory while changing the state.

**Comment by jerryshao** [comment]
Currently it is not possible to cancel non-spark statement, even for Spark only job's related statements can be cancelled. This is a limitation of statement cancelling.

**Comment by jerryshao** [comment]
Looks like this PR still requires change after [LIVY-244](https://issues.cloudera.org/projects/LIVY/issues/LIVY-244?filter=allopenissues). I will modify it after LIVY-244 is merged.

**Comment by tc0312** [comment]
Thank you so much Jerry.

**Comment by jerryshao** [comment]
@zjffdu please help to review this PR, thanks.

File: repl/src/main/scala/com/cloudera/livy/repl/ProcessInterpreter.scala [hunk 2]
```diff
@@ -55,6 +55,7 @@ abstract class ProcessInterpreter(process: Process)
     // jobs against pyspark or sparkr, so just return null here.
     null
   }
+
```
zjffdu: non-necessary change.

zjffdu: And I think cancel may not work for PySpark & SparkR, because ProcessInterpreter.start return null.

zjffdu: `getOrCreate` should return SparkContext which is created in the pyspark/spark initialization script.

zjffdu: And it's better to add test for pyspark & sparkr as well.

zjffdu:
I verified the latest change, but cancel in pyspark seems doesn't work. Only the statement state is changed to canceled, but the real job in spark is not canceled. I think the setJobGroup needs to be called in python rather than in scala. Because as the doc of spark, setJobGroup affect all the jobs started by this thread.

zjffdu:
If the changes is too much to pyspark & sparkr, we could do it in another PR and also update the title of PR

jerryshao:
Thanks @zjffdu , does PySpark and SparkR provides equivalent APIs to cancel JobGroup?

Also the problem I guess due to setting job group in another thread and Spark uses thread local properties to handle this group id.

A possible solution is to insert "setJobGroup" statement before every statement, rather than directly invoke the `SparkContext` API.

zjffdu:
Yes, that why I suggest to call setJobGroup in python script. PySpark do have such api, I haven't check sparkr yet.

jerryshao:
One potential problem of inserting “setJobGroup” statement before each statement is that when submitted statement is not complete, inserting "setJobGroup" statement will break the code: for example:

````
//statement A:

sc.parallelize(1 to 100).map
// comment

// set job group
sc.setJobGroup(...)

// statement B:
{ xxx }
```

repl/src/main/scala/com/cloudera/livy/repl/Session.scala
```diff
@@ -82,21 +88,57 @@ class Session(interpreter: Interpreter, stateChangedCallback: SessionState => Un
   def execute(code: String): Int = {
     val statementId = newStatementId.getAndIncrement()
     _statements(statementId) = new Statement(statementId, StatementState.Waiting, null)
+
     Future {
-      _statements(statementId) = new Statement(statementId, StatementState.Running, null)
+      def stmtState = _statements(statementId).state
+
+      if (stmtState != StatementState.Cancelled) {
+        lock.synchronized {
```
tc0312: We can get rid of this global lock if we make `Statement` mutable and lock the `Statement` being executed/cancelled.

tc0312: We can get rid of this table lock if we make `Statement` mutable and lock the `Statement` being executed/cancelled.

repl/src/main/scala/com/cloudera/livy/repl/Session.scala
```diff
@@ -82,21 +88,57 @@ class Session(interpreter: Interpreter, stateChangedCallback: SessionState => Un
   def execute(code: String): Int = {
     val statementId = newStatementId.getAndIncrement()
     _statements(statementId) = new Statement(statementId, StatementState.Waiting, null)
+
     Future {
-      _statements(statementId) = new Statement(statementId, StatementState.Running, null)
+      def stmtState = _statements(statementId).state
+
+      if (stmtState != StatementState.Cancelled) {
```
tc0312: Instead of having multiple `if (stmtState != StatementState.Cancelled)` blocks, we can merge them into a single one.

repl/src/main/scala/com/cloudera/livy/repl/Session.scala
```diff
@@ -82,21 +88,57 @@ class Session(interpreter: Interpreter, stateChangedCallback: SessionState => Un
...
+              new Statement(statementId, StatementState.Available, executeResult)
+          }
+        }
+      }
     }
+
     statementId
   }
 
+  def cancel(statementId: Int): Unit = {
```
tc0312: What's the API contract of the `cancel()`?
Should `cancel()` block until the statement is actually finished/cancelled? Or should `cancel()` return as soon as the statement is marked as `cancelling`?
For the former, this call should block.
For the latter, we should change the Statement state to `cancelling` to distinguish `cancelling` statements and `cancelled` statements. Current definition of `cancelled` is ambiguous.

repl/src/main/scala/com/cloudera/livy/repl/Session.scala
```diff
@@ -82,21 +88,57 @@ class Session(interpreter: Interpreter, stateChangedCallback: SessionState => Un
...
     Future {
-      _statements(statementId) = new Statement(statementId, StatementState.Running, null)
+      def stmtState = _statements(statementId).state
+
+      if (stmtState != StatementState.Cancelled) {
+        lock.synchronized {
+          if (stmtState != StatementState.Cancelled) {
+            _statements(statementId) = new Statement(statementId, StatementState.Running, null)
+            _sc.foreach(
+              _.setJobGroup(statementId.toString, s"Job group for statement $statementId"))
```
tc0312: Per Jeff's comment, instead of calling `setJobGroup()` here, I think we need to move it to the implementations of interpreter to support statement cancellation in non Scala languages.

If you agree, I think we can limit the scope of this PR for Scala support only.

zjffdu:
@jerryshao How about constructing a statement of `sc.setJobGroup` to prepend the user code before feeding to underlying scala/pyspark/sparkr interpreter, I think this would involve less change.

repl/src/main/scala/com/cloudera/livy/repl/Session.scala
jerryshao: It is hard to distinguish `cancelling` and `cancelled`:

1. `sc.cancelJobGroup` is non-blocking, it will return immediately.
2. `sc.cancelJobGroup` is asynchronous, Job will be cancelled in future after this is called.
3. Caller doesn't know when job is cancelled, so it doesn't know when to change the statement to `Cancelled`.

It is hard to set `cancelling` and `cancelled` in Livy side.

repl/src/main/scala/com/cloudera/livy/repl/Session.scala
jerryshao: This bunch of code is really ugly, I cannot figure out a lock free and elegant way to check the state while cancel the job.

jerryshao: Figure out a way to use `AtomicReference` to simplify the codes, will update soon.

repl/src/main/scala/com/cloudera/livy/repl/SparkRInterpreter.scala
```diff
@@ -133,7 +133,7 @@ class SparkRInterpreter(process: Process, backendInstance: Any, backendThread: T
   implicit val formats = DefaultFormats
 
   private[this] var executionCount = 0
-  override def kind: String = "sparkR"
+  override def kind: String = "sparkr"
```
zjffdu: Why to change the kind name ? This may cause incompatibility with previous version.

jerryshao: `sparkR` is not correct, please see:

```
object Kind {

  def apply(kind: String): Kind = kind match {
    case "spark" | "scala" => Spark()
    case "pyspark" | "python" => PySpark()
    case "pyspark3" | "python3" => PySpark3()
    case "sparkr" | "r" => SparkR()
    case other => throw new IllegalArgumentException(s"Invalid kind: $other")
  }

}
```

repl/src/main/scala/com/cloudera/livy/repl/Session.scala
```diff
@@ -210,4 +193,18 @@ class Session(interpreter: Interpreter, stateChangedCallback: SessionState => Un
...
   }
+
+  private def executeWithJobGroup(executionCount: Int, code: String): String = {
+    Kind(interpreter.kind) match {
+      case Spark() | PySpark() | PySpark3() =>
+        val cmd = """sc.setJobGroup(%d, "Job group for statement %d")"""
+          .format(executionCount, executionCount)
+        executeCode(executionCount, cmd)
+      case SparkR() =>
+        // SparkR doesn't support statement cancelling, so doesn't need to set job group
```
zjffdu: You can find `setJobGroup` here. http://spark.apache.org/docs/latest/api/R/index.html
And SparkContext.cancel can be called anywhere, not necessary in the same thread. So we can call SparkContext.cancel in scala side.

jerryshao: Not sure how to write R code, is below sufficient?

```
setJobGroup("myJobGroup", "My job group description", TRUE)
```

zjffdu: This is correct for spark2, but for spark1 we need to add SparkContext. 
Quoted from http://spark.apache.org/docs/latest/sparkr.html
```
The sc SparkContext parameter is no longer required for these functions: setJobGroup, clearJobGroup, cancelJobGroup
```

repl/src/main/scala/com/cloudera/livy/repl/Session.scala
tc0312: When the job is actually cancelled, the thread which's executing the cancelled Spark call will get a SparkException("Job x cancelled part of cancelled job group y").

jerryshao: How and when to catch this exception, in which thread? The job will be cancelled immediately or later on or never indefinitely, depends on the scheduler and pipeline. I'm inclining to not fix this unless user really need to distinguish `cancelling` or `cancelled`.

tc0312: The interpreter thread [right here](https://github.com/cloudera/livy/pull/230/files#diff-726504ec4098560b7f98fdec97b165e7R211)?

jerryshao: Does this also apply to python and R interpreter?

tc0312: Why doesn't?

jerryshao: > the thread which's executing the cancelled Spark 
>The interpreter thread right here?

From my understanding of the code, python and R interpreter uses another process to execute code, so is it still still possible to catch the exception?

tc0312: *It isn't really necessary to catch the exception.* But when that statement's finished, and the statement's state is cancelling, change it to cancelled. Else, change it to available.

repl/src/main/scala/com/cloudera/livy/repl/Session.scala
```diff
@@ -168,4 +190,24 @@ class Session(interpreter: Interpreter, stateChangedCallback: SessionState => Un
 
     compact(render(resultInJson))
   }
+
+  private def executeWithJobGroup(executionCount: Int, code: String): String = {
+    val cmd = Kind(interpreter.kind) match {
+      case Spark() | PySpark() | PySpark3() =>
+        s"""sc.setJobGroup($executionCount, "Job group for statement $executionCount")"""
```
zjffdu: miss quotation mark around `$executionCount`, this make the spark job not canceled

repl/src/main/scala/com/cloudera/livy/repl/Session.scala
```diff
@@ -168,4 +190,24 @@ class Session(interpreter: Interpreter, stateChangedCallback: SessionState => Un
...
+      case SparkR() =>
+        interpreter.asInstanceOf[SparkRInterpreter].sparkMajorVersion match {
+          case "1" =>
+            s"""setJobGroup(sc, $executionCount, "Job group for statement $executionCount", """ +
```
zjffdu: ditto.

zjffdu:
@jerryshao The spark job is actually canceled. I left some comments inline. 
Besides, I think you need to verify the evalue in statement execution response. 

This is what I get if the statement is canceld successfully.

```
{
  "total_statements": 1,
  "statements": [
    {
      "id": 0,
      "state": "cancelled",
      "output": {
        "status": "error",
        "execution_count": 0,
        "ename": "Error",
        "evalue": "org.apache.spark.SparkException: Job 0 cancelled part of cancelled job group 0",
        "traceback": [
          "  at org.apache.spark.scheduler.DAGScheduler.org$apache$spark$scheduler$DAGScheduler$$failJobAndIndependentStages(DAGScheduler.scala:1450)",
          "\n  at org.apache.spark.scheduler.DAGScheduler.handleJobCancellation(DAGScheduler.scala:1389)",
          "\n  at org.apache.spark.scheduler.DAGScheduler$$anonfun$handleJobGroupCancelled$1.apply$mcVI$sp(DAGScheduler.scala:795)",
          "\n  at org.apache.spark.scheduler.DAGScheduler$$anonfun$handleJobGroupCancelled$1.apply(DAGScheduler.scala:795)",
          "\n  at org.apache.spark.scheduler.DAGScheduler$$anonfun$handleJobGroupCancelled$1.apply(DAGScheduler.scala:795)",
          "\n  at scala.collection.mutable.HashSet.foreach(HashSet.scala:78)",
          "\n  at org.apache.spark.scheduler.DAGScheduler.handleJobGroupCancelled(DAGScheduler.scala:795)",
          "\n  at org.apache.spark.scheduler.DAGSchedulerEventProcessLoop.doOnReceive(DAGScheduler.scala:1638)",
          "\n  at org.apache.spark.scheduler.DAGSchedulerEventProcessLoop.onReceive(DAGScheduler.scala:1618)",
          "\n  at org.apache.spark.scheduler.DAGSchedulerEventProcessLoop.onReceive(DAGScheduler.scala:1607)",
          "\n  at org.apache.spark.util.EventLoop$$anon$1.run(EventLoop.scala:48)",
          "\n  at org.apache.spark.scheduler.DAGScheduler.runJob(DAGScheduler.scala:632)",
          "\n  at org.apache.spark.SparkContext.runJob(SparkContext.scala:1871)",
          "\n  at org.apache.spark.SparkContext.runJob(SparkContext.scala:1934)",
          "\n  at org.apache.spark.rdd.RDD$$anonfun$fold$1.apply(RDD.scala:1046)",
          "\n  at org.apache.spark.rdd.RDDOperationScope$.withScope(RDDOperationScope.scala:151)",
          "\n  at org.apache.spark.rdd.RDDOperationScope$.withScope(RDDOperationScope.scala:112)",
          "\n  at org.apache.spark.rdd.RDD.withScope(RDD.scala:358)",
          "\n  at org.apache.spark.rdd.RDD.fold(RDD.scala:1040)",
          "\n  at org.apache.spark.rdd.DoubleRDDFunctions$$anonfun$sum$1.apply$mcD$sp(DoubleRDDFunctions.scala:34)",
          "\n  at org.apache.spark.rdd.DoubleRDDFunctions$$anonfun$sum$1.apply(DoubleRDDFunctions.scala:34)",
          "\n  at org.apache.spark.rdd.DoubleRDDFunctions$$anonfun$sum$1.apply(DoubleRDDFunctions.scala:34)",
          "\n  at org.apache.spark.rdd.RDDOperationScope$.withScope(RDDOperationScope.scala:151)",
          "\n  at org.apache.spark.rdd.RDDOperationScope$.withScope(RDDOperationScope.scala:112)",
          "\n  at org.apache.spark.rdd.RDD.withScope(RDD.scala:358)",
          "\n  at org.apache.spark.rdd.DoubleRDDFunctions.sum(DoubleRDDFunctions.scala:33)",
          "\n  ... 48 elided"
        ]
      }
    }
  ]
}
```

jerryshao:
@zjffdu , can you please check R code, I'm not familiar with that part.

repl/src/main/scala/com/cloudera/livy/repl/Session.scala
```diff
@@ -168,4 +190,24 @@ class Session(interpreter: Interpreter, stateChangedCallback: SessionState => Un
 
     compact(render(resultInJson))
   }
+
+  private def executeWithJobGroup(executionCount: Int, code: String): String = {
```
zjffdu: executionCount -> statementId

repl/src/main/scala/com/cloudera/livy/repl/Session.scala
```diff
@@ -67,9 +69,9 @@ class Session(interpreter: Interpreter, stateChangedCallback: SessionState => Un
       changeState(SessionState.Idle())
       sc
     }
-    future.onFailure { case _ =>
-      changeState(SessionState.Error())
-    }
+
+    future.onFailure { case _ => changeState(SessionState.Error()) }
+    future.onSuccess { case sc => _sc = Option(sc).orElse(Some(SparkContext.getOrCreate())) }
```
zjffdu: I think we can call SparkContext.getOrCreate() here, https://github.com/cloudera/livy/blob/master/repl/src/main/scala/com/cloudera/livy/repl/ProcessInterpreter.scala#L56, so that we can always get active SparkContext here. Otherwise something wrong must happens.

repl/src/main/scala/com/cloudera/livy/repl/SparkRInterpreter.scala
```diff
@@ -126,14 +126,17 @@ object SparkRInterpreter {
   }
 }
 
-class SparkRInterpreter(process: Process, backendInstance: Any, backendThread: Thread,
-  sparkMajorVersion: String) extends ProcessInterpreter(process) {
+class SparkRInterpreter(process: Process,
+    backendInstance: Any,
+    backendThread: Thread,
+    val sparkMajorVersion: String)
```
zjffdu: val is not necessary

repl/src/test/scala/com/cloudera/livy/repl/SparkSessionSpec.scala
```diff
@@ -194,4 +200,22 @@ class SparkSessionSpec extends BaseSessionSpec {
...
+    // Sleep a while to make sure job is running
+    Thread.sleep(1000)
+    session.cancel(stmtId)
+
+    eventually(timeout(30 seconds), interval(100 millis)) {
+      assert(session.statements(stmtId).state.get() == StatementState.Cancelled)
+      session.statements(stmtId).output should include (
+        "Job 0 cancelled part of cancelled job group 0")
+    }
+  }
```
zjffdu: There's another case I can think of is that we have 2 statements, one is in RUNNING, and another is WAITING. and we cancel the second statement. After the first statement is finished, we should be able to cancel the second statement successfully.

rsc/src/main/java/com/cloudera/livy/rsc/driver/Statement.java
```diff
@@ -17,17 +17,19 @@
...
+
 import com.fasterxml.jackson.annotation.JsonRawValue;
 
 public class Statement {
   public final Integer id;
-  public final StatementState state;
+  public final AtomicReference<StatementState> state;
   @JsonRawValue
-  public final String output;
+  public String output;
```
zjffdu: Why to remove final ?

zjffdu:
@jerryshao overall it lgtm. I verified it spark, pyspark and sparkr.  Just left a few minor comments inline.

rsc/src/main/java/com/cloudera/livy/rsc/driver/Statement.java
jerryshao: Because we need to update the output later [here](https://github.com/cloudera/livy/pull/230/files#diff-726504ec4098560b7f98fdec97b165e7R98).

repl/src/test/scala/com/cloudera/livy/repl/SparkSessionSpec.scala
jerryshao: Sure, I will add this scenario.

repl/src/main/scala/com/cloudera/livy/repl/SparkRInterpreter.scala
jerryshao: this is exposed to be used by `session` to differentiate SparkR `setJobGroup` API, do we have any other way to get spark major version?

rsc/src/main/java/com/cloudera/livy/rsc/driver/Statement.java
zjffdu: I see

repl/src/main/scala/com/cloudera/livy/repl/Session.scala
```diff
@@ -168,4 +195,21 @@ class Session(interpreter: Interpreter, stateChangedCallback: SessionState => Un
 
     compact(render(resultInJson))
   }
+
+  private def setJobGroup(statementId: Int): String = {
+    val cmd = Kind(interpreter.kind) match {
+      case Spark() | PySpark() | PySpark3() =>
+        s"""sc.setJobGroup("$statementId", "Job group for statement $statementId")"""
```
tc0312: Can we do
```
val _ = sc.setJobGroup("$statementId", "Job group for statement $statementId")
```
to keep the id in res<id> unchanged?

repl/src/main/scala/com/cloudera/livy/repl/Session.scala
```diff
@@ -82,21 +84,46 @@ class Session(interpreter: Interpreter, stateChangedCallback: SessionState => Un
...
     _statements(statementId) = new Statement(statementId, StatementState.Waiting, null)
+
     Future {
-      _statements(statementId) = new Statement(statementId, StatementState.Running, null)
+      _statements(statementId).state.compareAndSet(StatementState.Waiting, StatementState.Running)
 
-      _statements(statementId) =
-        new Statement(statementId, StatementState.Available, executeCode(statementId, code))
+      setJobGroup(statementId)
+      val executeResult = if (_statements(statementId).state.get() != StatementState.Cancelling) {
```
tc0312: There's a race condition here. Consider the following execution orders:
Interpreter execution thread:
Session.execute() [line 92](https://github.com/cloudera/livy/pull/230/files#diff-726504ec4098560b7f98fdec97b165e7R92) state is StatementState.Running, will execute line 93.

Cancelling thread:
Session.cancel() [line 109](https://github.com/cloudera/livy/pull/230/files#diff-726504ec4098560b7f98fdec97b165e7R109)
oldStmtState =  StatementState.Runnig
Session.cancel() calls `_sc.foreach(_.cancelJobGroup(statementId.toString))` [line 113-114](https://github.com/cloudera/livy/pull/230/files#diff-726504ec4098560b7f98fdec97b165e7R114)

Interpreter execution thread:
Session.execute() calls executeCode() [line 93](https://github.com/cloudera/livy/pull/230/files#diff-726504ec4098560b7f98fdec97b165e7R93)

Because cancelJobGroup() is called before executeCode() is executed, calls to Spark will be executed.

tc0312: Another problem is, if user code is catching SparkException between calls, 2nd and further calls to Spark are not cancelled.
```
try {
  sc... # Call 1
} catch {
   case _ =>
}
try {
  sc... # Call 2
} catch {
   case _ =>
}
```
I think it's fine to not fix this but it should be documented clearly because it's hard to spot.

tc0312: There's a solution to both issues. cancel() should call sc.cancelJobGroup() in a loop until Statement's state changes to available or cancelled.

repl/src/test/scala/com/cloudera/livy/repl/SparkSessionSpec.scala
```diff
@@ -194,4 +200,39 @@ class SparkSessionSpec extends BaseSessionSpec {
...
+
+  it should "cancel spark jobs" in withSession { session =>
+    val stmtId = session.execute(
+      """sc.parallelize(0 to 10).map { i => Thread.sleep(10000); i + 1 }.collect""".stripMargin)
+    eventually(timeout(30 seconds), interval(100 millis)) {
+      assert(session.statements(stmtId).state.get() == StatementState.Running)
+    }
+
+    // Sleep a while to make sure job is running
+    Thread.sleep(2000)
```
tc0312: This hides the race condition I commented above. Please remove.

repl/src/main/scala/com/cloudera/livy/repl/Session.scala
```diff
@@ -168,4 +195,21 @@ class Session(interpreter: Interpreter, stateChangedCallback: SessionState => Un
 
     compact(render(resultInJson))
   }
+
+  private def setJobGroup(statementId: Int): String = {
```
jerryshao: That will change all the related interpreter codes including `Interpreter` interface, because we need to pass in `statementId`, I would suggest to change it later on, I really don't want to change so many codes.

repl/src/main/scala/com/cloudera/livy/repl/Session.scala
tc0312: Yes. It's called ignored variable. I tested it with scala repl 2.10.5 before suggesting it.
I'm surprised it doesn't work on Scala 2.11. But since Scala 2.11 doesn't provide res<id> variables, I think we can call `sc.setJobGroup("$statementId", "Job group for statement $statementId")` in Scala 2.11. It should be possible if we extract setJobGroup() to interpreters.

repl/src/main/scala/com/cloudera/livy/repl/Session.scala
tc0312: I think it's a bug we should really fix. If we hit this in a simple unit test, users will definitely hit it in production.

repl/src/main/scala/com/cloudera/livy/repl/Session.scala
jerryshao: > But since Scala 2.11 doesn't provide res variables, I think we can call sc.setJobGroup("$statementId", "Job group for statement $statementId") in Scala 2.11.

Can you explain more, I cannot get your point.

tc0312: Scala 2.11's repl  doesn't bind results to result variables like `res0` `res1` hence we don't need to use ignored variable.

jerryshao: From my test scala 2.11 repl will still occupy one placeholder:

```
scala> sc.setJobGroup("dfsa", "fds")

scala> 1
res1: Int = 1
```

As you mentioned 2nd statement "1" should be "res0", but it is "res1" as I tested.

tc0312: This's weird. I tested this with Livy w/ Spark 2.0:
screen shot 2016-12-14 at 1 04 38 am


```
     {
      "id": 5,
      "state": "available",
      "output": {
        "status": "ok",
        "execution_count": 5,
        "data": {
          "text/plain": "u'2.0.0.2.5.2.1-1'"
        }
      }
    },
    {
      "id": 6,
      "state": "available",
      "output": {
        "status": "ok",
        "execution_count": 6,
        "data": {
          "text/plain": "1"
        }
      }
    },
    {
      "id": 7,
      "state": "available",
      "output": {
        "status": "error",
        "execution_count": 7,
        "ename": "NameError",
        "evalue": "name 'res0' is not defined",
        "traceback": [
          "Traceback (most recent call last):\n",
          "NameError: name 'res0' is not defined\n"
        ]
      }
    }
```

jerryshao: Yes, currently I'm using a unique name to handle this.

repl/src/main/scala/com/cloudera/livy/repl/Session.scala
```diff
@@ -82,21 +84,53 @@ class Session(interpreter: Interpreter, stateChangedCallback: SessionState => Un
...
     statementId
   }
 
+  def cancel(statementId: Int): Unit = {
+    if (_statements.contains(statementId)) {
+      val oldStmtState = _statements(statementId).state.getAndSet(StatementState.Cancelling)
+
+      info(s"Statement $statementId is cancelling")
+
+      val singleThreadExecutor = ExecutionContext.fromExecutorService(
```
tc0312: Each cancel request is allocating a new thread pool. It's essentially unbounded. It's fine to fix it later. I think we can refactor this code to use a single thread with a priority queue.

repl/src/main/scala/com/cloudera/livy/repl/Session.scala
```diff
@@ -82,21 +84,53 @@ class Session(interpreter: Interpreter, stateChangedCallback: SessionState => Un
...
+        Executors.newSingleThreadExecutor())
+      Future {
+        if (oldStmtState == StatementState.Running) {
+          while (_statements(statementId).state.get() != StatementState.Cancelled) {
```
tc0312: It should timeout because it's not always possible to cancel a statement.

repl/src/main/scala/com/cloudera/livy/repl/Session.scala
```diff
@@ -82,21 +84,53 @@ class Session(interpreter: Interpreter, stateChangedCallback: SessionState => Un
...
+            _sc.foreach(_.cancelJobGroup(statementId.toString))
+            Thread.sleep(100)
```
tc0312: Please make it a RSCConf.

repl/src/main/scala/com/cloudera/livy/repl/Session.scala
jerryshao: From my understanding, `cancelling` state definitely changes to `cancelled`, so the loop will not be infinite.

tc0312: If the statement being cancelled doesn't make any calls to Spark, we can't cancel it.

repl/src/main/scala/com/cloudera/livy/repl/Session.scala
```diff
@@ -110,18 +116,18 @@ class Session(interpreter: Interpreter, stateChangedCallback: SessionState => Un
 
       info(s"Statement $statementId is cancelling")
 
-      val singleThreadExecutor = ExecutionContext.fromExecutorService(
-        Executors.newSingleThreadExecutor())
       Future {
         if (oldStmtState == StatementState.Running) {
```
tc0312: After moving the cancel logic into a future, I think it makes more sense to use the latest statement state instead of `oldStmtState`.

repl/src/main/scala/com/cloudera/livy/repl/Session.scala
```diff
@@ -110,18 +116,18 @@ class Session(interpreter: Interpreter, stateChangedCallback: SessionState => Un
...
-          while (_statements(statementId).state.get() != StatementState.Cancelled) {
+          val currTime = System.currentTimeMillis()
+          while (_statements(statementId).state.get() != StatementState.Cancelled &&
+            System.currentTimeMillis() - currTime <= TimeUnit.SECONDS.toMillis(30)) {
             _sc.foreach(_.cancelJobGroup(statementId.toString))
-            Thread.sleep(100)
+            Thread.sleep(livyConf.getTimeAsMs(RSCConf.Entry.JOB_CANCEL_TRIGGER_INTERVAL))
           }
         } else if (oldStmtState == StatementState.Available) {
           _statements(statementId).state.set(StatementState.Cancelled)
```
tc0312: execute() will set the statement state to cancelled. We can remove this clause.

repl/src/main/scala/com/cloudera/livy/repl/Session.scala
```diff
@@ -82,21 +90,53 @@ class Session(interpreter: Interpreter, stateChangedCallback: SessionState => Un
...
+      _statements(statementId).state.compareAndSet(
+        StatementState.Cancelling, StatementState.Cancelled)
     }
+
     statementId
   }
 
+  def cancel(statementId: Int): Unit = {
+    if (_statements.contains(statementId)) {
+      val oldStmtState = _statements(statementId).state.getAndSet(StatementState.Cancelling)
```
tc0312: _statements(statementId).state.set(StatementState.Cancelling)

repl/src/main/scala/com/cloudera/livy/repl/Session.scala
jerryshao: There's one possibility that when cancel is issued, the statement execution is already finished. In this cancel `execute` will not change the state of statement, so we should deliberately set it to `cancelled` here.

repl/src/main/scala/com/cloudera/livy/repl/Session.scala
```diff
@@ -82,21 +90,53 @@ class Session(interpreter: Interpreter, stateChangedCallback: SessionState => Un
...
+
+      info(s"Statement $statementId is cancelling")
+
+      Future {
```
tc0312:
```
import scala.concurrent.duration._

Future {
  Deadline d = livyConf.getTimeAsMs(RSCConf.Entry.JOB_CANCEL_TIMEOUT).milliseconds.fromNow
  while (_statements(statementId).state.get() == StatementState.Cancelling) {
    if (d.isOverdue()) {
      info(s"Failed to cancel statement $statementId.")
      _statements(statementId).state.compareAndSet(StatementState.Cancelling, StatementState.Running)
    } else {
      _sc.foreach(_.cancelJobGroup(statementId.toString))
      Thread.sleep(livyConf.getTimeAsMs(RSCConf.Entry.JOB_CANCEL_TRIGGER_INTERVAL))
    }
  }
  if (_statements(statementId).state.get() == StatementState.Cancelled) {
    info("Statement $statementId cancelled.")
  }
}(cancelExecutor)
```

repl/src/main/scala/com/cloudera/livy/repl/Session.scala
```diff
@@ -82,21 +90,53 @@ class Session(interpreter: Interpreter, stateChangedCallback: SessionState => Un
...
+    }
+  }
+
   def close(): Unit = {
     executor.shutdown()
     interpreter.close()
   }
 
-  def clearStatements(): Unit = synchronized {
+  def clearStatements(): Unit = {
```
tc0312: `cancel()` doesn't really work if `_statements.clear()` is called while it's running.
We can fix this with a lock or just remove `clearStatements()` since no one is using it.

jerryshao: This `clearStatements` will also has contention with `execute()`. I'm inclining to remove it right now. If we later on supports this feature, we could add back it.

repl/src/main/scala/com/cloudera/livy/repl/Session.scala
jerryshao: Thanks @tc0312 , I like this change.

repl/src/main/scala/com/cloudera/livy/repl/Session.scala
```diff
@@ -111,35 +113,43 @@ class Session(
   }
 
   def cancel(statementId: Int): Unit = {
-    if (_statements.contains(statementId)) {
-      val oldStmtState = _statements(statementId).state.getAndSet(StatementState.Cancelling)
+    if (!_statements.contains(statementId)) {
+      return
+    }
+
+    if (_statements(statementId).state.get() == StatementState.Available ||
```
jerryshao:
```
_statements(statementId).state.compareAndSet(StatementState.Waiting, StatementState.Cancelled)
  _statements(statementId).state.compareAndSet(StatementState.Running, StatementState.Cancelling)
```

Is it necessary, I think the state will definitely be this two kind.

jerryshao: Sorry I misunderstood it.

tc0312:
LGTM.
I see that you removed "InteractiveSessionSpec.should cancel the submitted statement" test. Instead, could you modify it to test with a mock RSCClient, and verify that it calls cancel()? It's definitely fine to do it in a separate PR.

@zjffdu Do you have more comments?

tc0312:
@jerryshao Mind if you create the JIRAs for further improvement of this PR? I can share some of the work. Thanks.

jerryshao:
@tc0312  I think this [unit test](https://github.com/cloudera/livy/pull/230/files#diff-16f32ef1d9d8412e072090611abd5d2cR132) will verify the scenario you mentioned.

tc0312:
@jerryshao 
I was thinking more about testing InteractiveSession.cancelStatement() is calling client.get.cancelReplCode().
It's mostly to avoid regression. It's not a big deal. We can add it later.

jerryshao:
@zjffdu , do you have more comments?

jerryshao:
Create a patch to track the potential improvements we could do (https://issues.cloudera.org/browse/LIVY-278).

zjffdu:
@jerryshao @tc0312 LGTM. I will merge it if no more comments after CI passed.

PR #231, opened 2016-11-15
LIVY-244. Replace livy-server repl state polling thread with RPC.

tc0312:


repl/src/main/scala/com/cloudera/livy/repl/ReplDriver.scala
```diff
@@ -70,22 +73,16 @@ class ReplDriver(conf: SparkConf, livyConf: RSCConf)
...
-      val state = session.state.toString
-      new ReplJobResults(stmts.sortBy(_.id), state)
+  def handle(ctx: ChannelHandlerContext, msg: BaseProtocol.GetReplJobResults): ReplJobResults = {
+    val statements = if (msg.allResults) {
+      session.statements.values.toArray
+    } else {
+      assert(msg.from != null)
+      assert(msg.size != null)
+      val until = msg.from + msg.size
+      session.statements.filterKeys(id => id >= msg.from && id < until).values.toArray
```
jerryshao: From my understanding in most of the cases querying one statement is enough, compared to the previous code if we only want to get one statement, here we still requires to scan the map, this might be a regression IIUC.

 Can we separate the scenario for querying one statement, using another protocol to handle this issue?

repl/src/main/scala/com/cloudera/livy/repl/Session.scala
```diff
@@ -106,20 +100,34 @@ class Session(interpreter: Interpreter)
...
+  private def changeState(newState: SessionState): Unit = {
+    synchronized {
+      _state = newState
+    }
+    stateChangedCallback(newState)
+  }
+
   private def executeCode(executionCount: Int, code: String): String = synchronized {
-    _state = SessionState.Busy()
+    changeState(SessionState.Busy())
```
jerryshao: Can we simplify the Session state, do we really need to differentiate the state of session in such a fine-grained way?

State like "running", "busy" or "idle" are all means session is running correctly, do we really need to differentiate? I don't see any code specifically requires such state like "busy".

repl/src/main/scala/com/cloudera/livy/repl/Session.scala
```diff
@@ -106,20 +100,34 @@ class Session(interpreter: Interpreter)
...
+
+    def transitToIdle() = {
+      val executingLastStatement = executionCount == newStatementId.intValue() - 1
+      if (_statements.isEmpty || executingLastStatement) {
```
jerryshao: IMHO this is a bit tricky to handle session state jitter, for example, if statement is sent to ReplDriver in every second, and each statement is finished in 1 second. Then you will still meet frequent session state change.

From my understanding, it would be better to simplify the session state, that will simplify the maintenance overhead and enhance the scalability.

rsc/src/main/java/com/cloudera/livy/rsc/RSCClient.java
```diff
@@ -61,6 +61,7 @@
 
   private ContextInfo contextInfo;
   private volatile boolean isAlive;
+  private String replState;
```
jerryshao: Please add `volatile`.

rsc/src/main/java/com/cloudera/livy/rsc/RSCClient.java
```diff
@@ -382,5 +386,10 @@ private void handle(ChannelHandlerContext ctx, JobStarted msg) {
         LOG.warn("Received event for unknown job {}", msg.id);
       }
     }
+
+    private void handle(ChannelHandlerContext ctx, ReplState msg) {
+      LOG.debug("Received repl state for {}", msg.state);
```
jerryshao: Be better to change to trace log.

rsc/src/main/java/com/cloudera/livy/rsc/driver/RSCDriver.java
```diff
@@ -172,6 +172,11 @@ public RpcDispatcher onNewClient(Rpc client) {
         registerClient(client);
         return RSCDriver.this;
       }
+
+      @Override
+      public void onSaslComplete(Rpc client) {
```
jerryshao: Can you please explain the change here? What is the relation to sasl?

tc0312: When a new RSCClient connects to RSCDriver, repl should tells server its state. I added this callback to serve this purpose.
It's similar to [onNewClient()](https://github.com/cloudera/livy/pull/231/files/6f72649bfd69abb846f2e93f2e0440278b05829e#diff-fc62b1e9be342e38eeeae8a5f551aa78R271). However, RpcDispatcher isn't set when onNewClient() is called. Rpc.call() can't be called until RpcDispatcher is set. I added the new callback to workaround this issue.
It's called [here].(https://github.com/cloudera/livy/pull/231/files/6f72649bfd69abb846f2e93f2e0440278b05829e#diff-fc62b1e9be342e38eeeae8a5f551aa78R271)
It's called after Sasl authentication is completed.

repl/src/main/scala/com/cloudera/livy/repl/ReplDriver.scala
tc0312: Added a special case for this without adding a new message in protocol.

rsc/src/main/java/com/cloudera/livy/rsc/RSCClient.java
tc0312: Fixed.

rsc/src/main/java/com/cloudera/livy/rsc/RSCClient.java
tc0312: Fixed.

core/src/main/scala/com/cloudera/livy/sessions/SessionState.scala
```diff
@@ -25,6 +25,22 @@ sealed trait SessionState {
 
 object SessionState {
 
+  def parseString(s: String): SessionState = {
```
tc0312: Fixed.

repl/src/main/scala/com/cloudera/livy/repl/Session.scala
jerryshao: Frankly say, if I'm a user of Livy interpreter, I will not rely on "busy" or "idle" state, what I want to differentiate is "running" or not. I don't think Zeppelin really depends on that, not sure about Jupyter. Also as I mentioned before, such fine-grained status will increase the state management overhead for user (If user really want to differentiate busy state).

jerryshao: Also how to deal with the "busy" state for user is a question. What user get will always be later than the state updated in server side, so user may get a chance to get "idle" state and do some actions, but the state is changed to "busy" in server side, and vice versa. So in such case state showing to user and related actions by user cannot really prove the actual state and action for user.

tc0312: I don't disagree with you. But I'm hesitate to remove this now. I think we should keep the busy state. Start a conversation in livy-user@ mailing list and ask our users what's their takes on this.

repl/src/main/scala/com/cloudera/livy/repl/Session.scala
```diff
@@ -106,20 +100,34 @@ class Session(interpreter: Interpreter)
     _statements.clear()
   }
 
+  private def changeState(newState: SessionState): Unit = {
+    synchronized {
```
jerryshao: Do we need to check whether previous state is the same as new state, so we don't need to update.

PR #234, opened 2016-11-21
LIVY-251. YARN session leaked if PingJob times out

linchan-ms:
Fixed the issue and added test

### DIFF
--- a/rsc/src/main/java/com/cloudera/livy/rsc/RSCConf.java
+++ b/rsc/src/main/java/com/cloudera/livy/rsc/RSCConf.java
@@ -71,7 +71,8 @@ public class RSCConf extends ClientConf<RSCConf> {
     SASL_MECHANISMS("rpc.sasl.mechanisms", "DIGEST-MD5"),
     SASL_QOP("rpc.sasl.qop", null),
 
-    TEST_STUCK_END_SESSION("test.do_not_use.stuck_end_session", false);
+    TEST_STUCK_END_SESSION("test.do_not_use.stuck_end_session", false),
+    TEST_PING_JOB_ERROR("test.do_not_use.ping_job_error", false);
 
     private final String key;
     private final Object dflt;
--- a/rsc/src/main/java/com/cloudera/livy/rsc/RSCConf.java
+++ b/rsc/src/main/java/com/cloudera/livy/rsc/RSCConf.java
@@ -72,7 +72,7 @@ public class RSCConf extends ClientConf<RSCConf> {
     SASL_QOP("rpc.sasl.qop", null),
 
     TEST_STUCK_END_SESSION("test.do_not_use.stuck_end_session", false),
-    TEST_PING_JOB_ERROR("test.do_not_use.ping_job_error", false);
+    TEST_STUCK_START_SESSION("test.do_not_use.stuck_start_session", false);
 
     private final String key;
     private final Object dflt;
--- a/rsc/src/main/java/com/cloudera/livy/rsc/RSCConf.java
+++ b/rsc/src/main/java/com/cloudera/livy/rsc/RSCConf.java
@@ -72,7 +72,7 @@ public class RSCConf extends ClientConf<RSCConf> {
     SASL_QOP("rpc.sasl.qop", null),
 
     TEST_STUCK_END_SESSION("test.do_not_use.stuck_end_session", false),
-    TEST_STUCK_START_SESSION("test.do_not_use.stuck_start_session", false);
+    TEST_STUCK_START_DRIVER("test.do_not_use.stuck_start_driver", false);
 
     private final String key;
     private final Object dflt;
--- a/rsc/src/main/java/com/cloudera/livy/rsc/driver/RSCDriver.java
+++ b/rsc/src/main/java/com/cloudera/livy/rsc/driver/RSCDriver.java
@@ -163,13 +163,16 @@ public class RSCDriver extends BaseProtocol {
     // on the cluster, it would be tricky to solve that problem in a generic way.
     livyConf.set(RPC_SERVER_ADDRESS, null);
 
-    if (livyConf.getBoolean(TEST_STUCK_START_SESSION)) {
+    if (livyConf.getBoolean(TEST_STUCK_START_DRIVER)) {
       // Test flag is turned on so we will just infinite loop here. It should cause
       // timeout and we should still see yarn application being cleaned up.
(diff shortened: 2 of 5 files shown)


tc0312:
LGTM.

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
```diff
@@ -71,11 +72,18 @@ object InteractiveSession extends Logging {
       mockClient: Option[RSCClient] = None): InteractiveSession = {
     val appTag = s"livy-session-$id-${Random.alphanumeric.take(8).mkString}"
 
+    val livyConfWithTestConf = if (request.conf.contains(RSCConf.Entry.TEST_PING_JOB_ERROR.key())) {
+      new LivyConf().setAll(livyConf).set(RSCConf.Entry.TEST_PING_JOB_ERROR.key(), "true")
```
jerryshao: What is the purpose of using the conf, I'm not sure as a user how to use it, what it stands for?

jerryshao:
Taking a look at your description on JIRA, is it a good approach to kill the application is resource is not enough?

1. PingJob can be removed after session state pushing mechanism is merged.
2. What is the proper solution to handle the session if resource is not enough to start application, can we just return "starting" state to user and let user to decide whether to kill or still wait?

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
tc0312: I think this's a test configuration. It's forcing `PingJob` to fail by throwing an exception. When `PingJob` fails `livy-server` changes the session state to dead and trigger his fix.

jerryshao: So when should user set this configuration?

tc0312:
> PingJob can be removed after session state pushing mechanism is merged.

If we remove PingJob, I think it's still necessary to listen to `Promise` `driverRpc` in `RSCClient` for connection error. We can move the logic there.

> What is the proper solution to handle the session if resource is not enough to start application, can we just return "starting" state to user and let user to decide whether to kill or still wait?

This behaviour is ideal for batch but might not fit interactive because users wants to use interactive ASAP.

jerryshao:
I still don't know setting this configuration could solve the problem mentioned in JIRA? So user have to set this every time? I really don't know how to use it by user?

jerryshao:
Checked again, found that the configuration is only for test, IIUC the real fix is only 3 lines of code:

```
          app.foreach { a =>
             info(s"Failed to ping RSC driver for session $id. Killing application.")
             a.kill()
```

It really confuses me.

linchan-ms:
Thanks Jerry. Yes the fix is only those 3 lines. The other change to product code is simply to facilitate testing. I tried to name the flag with test_do_not_use and put the word test in all the affected variables. Also the changes are in 2 different commits. Do you think it will help to make it even clearer if I add more comments in the PingJob code?

jerryshao:
>If we remove PingJob, I think it's still necessary to listen to Promise driverRpc in RSCClient for connection error. We can move the logic there.

Yes, that would be another approach. But for me current fix is OK. Ideally we could remove this `PingJob` and consolidate the codes.

For the test thing, I think current test cannot really reflect the issue we met. We deliberately fail the PingJob to simulate resource not enough issue, it is a little misleading. Also this failure will call `stopSession` to stop App, we cannot tell it is your code or `stopSession` really stops the app, especially in normal situation like what you simulated for test.

So IMHO I would suggest we don't require to add a test here, since this test cannot really reflect the issue you met. Or if we could simulate the issue we met, that would be better.

linchan-ms:
> For the test thing, I think current test cannot really reflect the issue we met. We deliberately fail the PingJob to simulate resource not enough issue, it is a little misleading. Also this failure will call  stopSession  to stop App, we cannot tell it is your code or  stopSession  really stops the app in normal situation like what you simulated for test.

I think the test does do something. If we don't have the code fix, the test will fail. Essentially the test is a little more general that the specific case I described above. Basically, we are mimicing a case where the pingJob failed (in real case it could timeout, our testhook just made it fail a little faster).

jerryshao:
@linchan-ms What I try to deliver is that the failure (leakage) is not due to timeout of PingJob or failure of PingJob, it is due to timeout of RPC and passed this failure to PingJob and get caught by `InteractiveSession`. This issue will still be exposed even if there's no PingJob.

In you test case, you're trying to fail the PingJob which actually is not the same issue as I mentioned above. So if you could have a test really reflect the problem of RPC timeout, that would be better IMO.

linchan-ms:
As suggested, moved the test hook to RSC startup code. Will just loop there to cause a pingJob timeout which will mimic the real failure scenario.

rsc/src/main/java/com/cloudera/livy/rsc/RSCConf.java
```diff
@@ -72,7 +72,7 @@
     SASL_QOP("rpc.sasl.qop", null),
 
     TEST_STUCK_END_SESSION("test.do_not_use.stuck_end_session", false),
-    TEST_PING_JOB_ERROR("test.do_not_use.ping_job_error", false);
+    TEST_STUCK_START_SESSION("test.do_not_use.stuck_start_session", false);
```
tc0312: It's blocking the entire driver to start. Mind if we rename it to TEST_STUCK_START_DRIVER?
LGTM otherwise.

rsc/src/main/java/com/cloudera/livy/rsc/driver/RSCDriver.java
```diff
@@ -163,6 +163,16 @@ private void initializeServer() throws Exception {
     // on the cluster, it would be tricky to solve that problem in a generic way.
     livyConf.set(RPC_SERVER_ADDRESS, null);
 
+    if (livyConf.getBoolean(TEST_STUCK_START_SESSION)) {
+      // Test flag is turned on so we will just infinite loop here. It should cause
+      // timeout and we should still see yarn application being cleaned up.
+      LOG.info("Infinite looping as test flag TEST_STUCK_START_SESSION is turned on.");
+      while(true)
+      {
```
jerryshao: Nit: move `{` to above line.

jerryshao:
LGTM, just some small issues.

PR #235, opened 2016-11-22
LIVY-250. Change livy.repl.jars to work with different scala interpreter

jerryshao:
Current `livy.repl.jars` cannot automatically pick jars according to different scala interpreter. So here propose one way to address this issue. 

User could list all the dependencies to `livy.repl.jars`, including `livy-repl_2.10` and `livy-repl_2.11`, Livy automatically pick the right jars according to Spark's scala version and scala jar convention.

\cc @tc0312 .

### DIFF
--- a/server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
+++ b/server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
@@ -155,9 +155,9 @@ object InteractiveSession extends Logging {
       Option(livyConf.get(LIVY_REPL_JARS)).map { jars =>
         val regex = """[\w-]+_(\d\.\d\d).*\.jar""".r
         jars.split(",").filter { name => new Path(name).getName match {
-              // Filter out unmatched scala jars
+            // Filter out unmatched scala jars
             case regex(ver) => ver == scalaVersion
-              // Keep all the java jars end with ".jar"
+            // Keep all the java jars end with ".jar"
             case _ => name.endsWith(".jar")
           }
         }.toList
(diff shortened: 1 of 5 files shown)


server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
```diff
@@ -149,7 +152,27 @@ object InteractiveSession extends Logging {
...
     def livyJars(livyConf: LivyConf, scalaVersion: String): List[String] = {
-      Option(livyConf.get(LIVY_REPL_JARS)).map(_.split(",").toList).getOrElse {
+      Option(livyConf.get(LIVY_REPL_JARS)).map { jars =>
+        val regex = """[\w-]+_(\d\.\d\d).*\.jar""".r
+        jars.split(",").flatMap { n =>
+          val trimmed = n.trim
+          if (!trimmed.startsWith("local")) {
+            // expand the glob path for file and hdfs scheme
+            val path = new Path(toURI(trimmed))
+            val pathFs = FileSystem.get(path.toUri, livyConf.hadoopConf)
```
jerryshao: I don't think the latency will be huge by adding this API. 

1. `FileSystem` is locally cached, each time when you call `FileSystem.get` it will not create a new instance, just fetch a cached one.
2. For the latency of `Path.globStatus`, local hdfs client will cache the metadata of NN, so it may not trigger a RPC for each call. And to the worse, only one RPC call is enough. The latency is not so huge.
3. Compare to the time of allocating containers and starting Spark applications on YARN, the time spent on this part is very very small. Usually it will take about 30+ seconds to fully start a Spark application on YARN (in our test), here this RPC call may only take less than 1 second.

tc0312: Do you mind adding a simple check to bypass glob expansion if glob patterns are absent?
Or we could also remove glob expansion entirely?

jerryshao:
Unit test failure due to #251 .

jerryshao:
@tc0312 please help to review, I think this bug should be fixed before 0.3 is out. Thanks a lot.

PR #236, opened 2016-11-24
LIVY-253. Simplify interactive session state management 

jerryshao:
With LIVY-244, InteractiveSession bring in two session state, one is from yarn and another is from repl session, this bring the complexity of session state management compared to previous code. So here propose to simplify the state management code and unify these two states.


server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
```diff
@@ -387,14 +390,14 @@ class InteractiveSession(
       override def onJobFailed(job: JobHandle[Void], cause: Throwable): Unit = errorOut()
 
       override def onJobSucceeded(job: JobHandle[Void], result: Void): Unit = {
-        transition(SessionState.Running())
+        transition(SessionState.Idle())
```
tc0312: The repl might be in busy state. We shouldn't assume it's idle.

tc0312: The repl might be in busy state if it's a recovered session. We shouldn't assume it's idle.

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
```diff
@@ -511,24 +504,24 @@ class InteractiveSession(
...
-    require(serverSideState.isActive, "Session isn't active.")
+    require(_state.isActive, "Session isn't active.")
     require(client.isDefined, "Session is active but client hasn't been created.")
   }
 
   private def ensureRunning(): Unit = synchronized {
-    serverSideState match {
-      case SessionState.Running() =>
+    _state match {
+      case SessionState.Running() | SessionState.Idle() | SessionState.Busy() => Unit
```
tc0312: With your change, `running` isn't a legal repl state. We can remove it.

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
```diff
@@ -376,6 +376,9 @@ class InteractiveSession(
     }
     uriFuture onFailure { case e => warn("Fail to get rsc uri", e) }
 
+    // Register this class to RSCClient as a session state listener
+    client.get.registerStateListener(this)
```
tc0312: `livy-server` won't get notifications from repl state changes happened before registerStateListener()'s called. `livy-server` will not show the current repl state until repl changes its state once after connection.

jerryshao: I think normally this `register` should be invoked earlier than `RSCClient` connecting to remote driver. This is a very fast non-block operation compared to network connection operations in `RSCClient`.

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
jerryshao: Originally it is `Idle` state, your modification of LIVY-244 changed it to `Running` state. I didn't change the semantics for session recovery unless originally it is wrong.

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
jerryshao: Yes, actually in interactive session running is not used, only idle and busy and honored.

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
tc0312: LIVY-244 changes the **server state** to Running state so server will query repl for the actual state. The actual state could be idle or busy. Your code does change the semantics.

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
tc0312: But it's a race condition that could happen. I wouldn't introduce a race condition to simplify something.
Originally I took a similar approach too and moved away from it after spotting this race condition.

tc0312:
Let's find a way to simply this without introducing race condition :).

jerryshao:
OK, let me find out another way to address it.

PR #238, opened 2016-11-29
LIVY-256. Add Gzip support for Http rest response

jerryshao:
Add a configuration to identify whether to enable this feature or not. Manually verified in local environment.

### DIFF
--- a/client-http/src/main/java/com/cloudera/livy/client/http/HttpConf.java
+++ b/client-http/src/main/java/com/cloudera/livy/client/http/HttpConf.java
@@ -25,12 +25,14 @@ import com.cloudera.livy.client.common.ClientConf;
 class HttpConf extends ClientConf<HttpConf> {
 
   static enum Entry implements ConfEntry {
-    CONNETION_TIMEOUT("connection.timeout", "10s"),
+    CONNECTION_TIMEOUT("connection.timeout", "10s"),
     CONNECTION_IDLE_TIMEOUT("connection.idle.timeout", "10m"),
     SOCKET_TIMEOUT("connection.socket.timeout", "5m"),
 
     JOB_INITIAL_POLL_INTERVAL("job.initial_poll_interval", "100ms"),
-    JOB_MAX_POLL_INTERVAL("job.max_poll_interval", "5s");
+    JOB_MAX_POLL_INTERVAL("job.max_poll_interval", "5s"),
+
+    CONTENT_COMPRESS_ENABLE("content.compress.enable", true);
 
     private final String key;
     private final Object dflt;
--- a/client-http/src/main/java/com/cloudera/livy/client/http/LivyConnection.java
+++ b/client-http/src/main/java/com/cloudera/livy/client/http/LivyConnection.java
@@ -71,7 +71,7 @@ class LivyConnection {
     RequestConfig reqConfig = new RequestConfig() {
       @Override
       public int getConnectTimeout() {
-        return (int) config.getTimeAsMs(CONNETION_TIMEOUT);
+        return (int) config.getTimeAsMs(CONNECTION_TIMEOUT);
       }
 
       @Override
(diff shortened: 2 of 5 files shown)


server/src/main/scala/com/cloudera/livy/LivyConf.scala
```diff
@@ -109,6 +109,10 @@ object LivyConf {
   // How often Livy polls YARN to refresh YARN app state.
   val YARN_POLL_INTERVAL = Entry("livy.server.yarn.poll-interval", "5s")
 
+  // Http compression is enabled or not, if true the json data transmitted with be
+  // compressed with GZIP.
+  val COMPRESS_SUPPORTED = Entry("livy.server.content_compress.enabled", false)
```
tc0312: I think `GZipSupport` looks at `Accept-Header`. What do you think about enabling `GZipSupport` all the time? Then we can remove this conf and keep things simple.

jerryshao: From my understanding, several http clients like python's requester and apache http client as well as Chrome web browser by default enable gzip in their http request's accept-header, so if server response with gzipped content, then client will try to unzip them, if not then client will treat as plain data.

So unless we explicitly disable gzip in the client side, this compression will always be on, I'm not sure if it is expected.

This actually can be controlled either in server side or client side, I don't have a strong inclination, just looking for suggestion.

tc0312: Does  python's requester and apache http client handle unzipping transparently?

jerryshao: Yes, I think it is the basic functionality of http client.

tc0312: I think it's safe to enable it by default. It would be great if others can chime in.

client-http/src/main/java/com/cloudera/livy/client/http/HttpConf.java
```diff
@@ -25,12 +25,14 @@
...
-    CONNETION_TIMEOUT("connection.timeout", "10s"),
+    CONNECTION_TIMEOUT("connection.timeout", "10s"),
     CONNECTION_IDLE_TIMEOUT("connection.idle.timeout", "10m"),
     SOCKET_TIMEOUT("connection.socket.timeout", "5m"),
 
     JOB_INITIAL_POLL_INTERVAL("job.initial_poll_interval", "100ms"),
-    JOB_MAX_POLL_INTERVAL("job.max_poll_interval", "5s");
+    JOB_MAX_POLL_INTERVAL("job.max_poll_interval", "5s"),
+
+    CONTENT_COMPRESS_ENABLE("content.compress.enable", true);
```
tc0312: Unused. Please remove.

client-http/src/main/java/com/cloudera/livy/client/http/LivyConnection.java
```diff
@@ -83,6 +83,11 @@ public int getSocketTimeout() {
       public boolean isAuthenticationEnabled() {
         return true;
       }
+
+      @Override
+      public boolean isContentCompressionEnabled() {
```
tc0312: Unused. Please remove.

client-http/src/main/java/com/cloudera/livy/client/http/HttpConf.java
jerryshao: Why this is unused, I cannot get it.

tc0312: What are the potential use cases for isContentCompressionEnabled()?

jerryshao: This is just a configuration to disable gzip from client side, by default it is true, if user want to turn off gzip compression (Accept-Encoding), that could set this to false through Job API.

jerryshao: Also this works with `public boolean isContentCompressionEnabled()` to decide whether gzip compression should be turned off.

PR #239, opened 2016-11-29
LIVY-257. access log is not available

zjffdu:
Straightforward bugfix for access log file.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -71,6 +71,8 @@
     <test.redirectToFile>true</test.redirectToFile>
     <execution.root>${user.dir}</execution.root>
     <spark.home>${execution.root}/dev/spark</spark.home>
+    <!--  used for testing, NCSARequestLog use it for access log  -->
+    <livy.log.dir>${basedir}/target</livy.log.dir>
 
     <!-- Set this to "true" to skip R tests. -->
     <skipRTests>false</skipRTests>
@@ -693,6 +695,7 @@
             <environmentVariables>
               <LIVY_TEST>true</LIVY_TEST>
               <SPARK_HOME>${spark.home}</SPARK_HOME>
+              <LIVY_LOG_DIR>${livy.log.dir}</LIVY_LOG_DIR>
             </environmentVariables>
             <systemProperties>
               <java.awt.headless>true</java.awt.headless>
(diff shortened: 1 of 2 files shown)


jerryshao:
LGTM, the test failure should also be fixed.

pom.xml
```diff
@@ -693,6 +695,7 @@
             <environmentVariables>
               <LIVY_TEST>true</LIVY_TEST>
               <SPARK_HOME>${spark.home}</SPARK_HOME>
+              <LIVY_LOG_DIR>${livy.log.dir}</LIVY_LOG_DIR>
```
jerryshao: Will this change the log output dir for integration test, previously it is under `miniLivyMain`.

zjffdu: It is still under `miniLivyMain`, but the access log is under `target`

jerryshao: I see, thanks.

pom.xml
```diff
@@ -71,6 +71,8 @@
     <test.redirectToFile>true</test.redirectToFile>
     <execution.root>${user.dir}</execution.root>
     <spark.home>${execution.root}/dev/spark</spark.home>
+    <!--  used for testing, NCSARequestLog use it for access log  -->
+    <livy.log.dir>${basedir}/target</livy.log.dir>
```
tc0312: Nit: Mind if we keep it under MiniLivyMain for integration test?

server/src/main/scala/com/cloudera/livy/server/WebServer.scala
```diff
@@ -74,12 +72,14 @@ class WebServer(livyConf: LivyConf, var host: String, var port: Int) extends Log
   val handlers = new HandlerCollection
   handlers.addHandler(context)
 
-//  configure the access log
+  //  configure the access log
```
tc0312: Nit: 
```
// Configure the access log
```

pom.xml
zjffdu: The reason I put it in parent is that `HttpClientSpec` also use WebServer.

tc0312: I see.

server/src/main/scala/com/cloudera/livy/server/WebServer.scala
```diff
@@ -74,12 +72,14 @@ class WebServer(livyConf: LivyConf, var host: String, var port: Int) extends Log
...
   val requestLogHandler = new RequestLogHandler
-  val requestLog = new NCSARequestLog("/jetty-yyyy_mm_dd.request.log")
+  val requestLog = new NCSARequestLog(sys.env.getOrElse("LIVY_LOG_DIR",
+    sys.env("LIVY_HOME") + "/logs") + "/jetty-yyyy_mm_dd.request.log")
```
tc0312: Nit: name it "yyyy_mm_dd.request.log" instead of "jetty-yyyy_mm_dd.request.log"

tc0312:
LGTM.

PR #242, opened 2016-12-03
LIVY-259. Name of interactive sessions are not override-able.

linchan-ms:
Allow application name to be set from "name" parameter and "conf"s


jerryshao:
LGTM.

PR #244, opened 2016-12-06
LIVY-264. Fixed PySpark %table magic returning wrong header.

purechoc:
%table returns wrong header if input is a list of rows.
```
from pyspark.sql.types import Row
x = [Row(age=1, name=u'a'), Row(age=2, name=u'b'), Row(age=3, name=u'c')]
%table x
```
```
{
  ...
        "data": {
          "application/vnd.livy.table.v1+json": {
            "headers": [
              {
                "type": "INT_TYPE",
                "name": "0"
              },
              {
                "type": "STRING_TYPE",
                "name": "1"
              }
  ...
```
Notice that the header names are wrong.
Workaround the issue by converting row to dict.


tc0312:
Mind if you add test?

purechoc:
@tc0312 
there is any way to import pyspark library in unitest?

tc0312:
You can do it in integration test:
https://github.com/cloudera/livy/blob/master/integration-test/src/test/scala/com/cloudera/livy/test/InteractiveIT.scala#L66

purechoc:
ok. i'll try make uniitest

purechoc:
add integration test complete.

integration-test/src/test/scala/com/cloudera/livy/test/InteractiveIT.scala
```diff
@@ -69,7 +69,10 @@ class InteractiveIT extends BaseIntegrationTestSuite {
       s.run("sqlContext").verifyResult(startsWith("<pyspark.sql.context.HiveContext"))
       s.run("sc.parallelize(range(100)).map(lambda x: x * 2).reduce(lambda x, y: x + y)")
         .verifyResult("9900")
-
+      s.run("from pyspark.sql.types import Row").verifyResult(".*")
```
tc0312: verifyResult(".*") is a bit too board. Mind if you restrict it a bit?

repl/src/main/resources/fake_shell.py
```diff
@@ -403,6 +406,9 @@ def magic_table(name):
         cols = []
         data.append(cols)
 
+        if os.environ.get("LIVY_TEST") != "true" and isinstance(row, pyspark.sql.types.Row):
```
tc0312: Mind if you explain what's `os.environ.get("LIVY_TEST") != "true"` for?

purechoc: when os.environ.get("LIVY_TEST") == "true"
fake_shell do not import spark library (can not find pyspark.sql.types.Row)

integration-test/src/test/scala/com/cloudera/livy/test/InteractiveIT.scala
purechoc: got it. i'll fix it

repl/src/main/resources/fake_shell.py
tc0312: I think we can do the type checking without importing the class by doing string comparsion on row.__class__.__name__.

An alternative is, we can just check does asDict() present in the row object. If it does, then call it.

integration-test/src/test/scala/com/cloudera/livy/test/InteractiveIT.scala
```diff
@@ -69,9 +69,9 @@ class InteractiveIT extends BaseIntegrationTestSuite {
       s.run("sqlContext").verifyResult(startsWith("<pyspark.sql.context.HiveContext"))
       s.run("sc.parallelize(range(100)).map(lambda x: x * 2).reduce(lambda x, y: x + y)")
         .verifyResult("9900")
-      s.run("from pyspark.sql.types import Row").verifyResult(".*")
+      s.run("from pyspark.sql.types import Row").verifyResult("")
       s.run("x = [Row(age=1, name=u'a'), Row(age=2, name=u'b'), Row(age=3, name=u'c')]")
```
tc0312: Sorry mind if you explain what you are fixing?
I ran these code without your fix but %table magic is still working.

repl/src/main/resources/fake_shell.py
purechoc: 'string comparison' looks better. I will apply it

tc0312:
Merging once test passes.

PR #245, opened 2016-12-07
LIVY-252. HTTP auth support for Scala LivyClient.

tc0312:


tc0312:
@jerryshao @zjffdu Please review my change. Thanks.

jerryshao:
Does Livy's embedded jetty server already support user/passwd auth?

tc0312:
In our use case, authentication is added by the security gateway.
I consider HTTP authentication a separate feature request.

jerryshao:
>I consider HTTP authentication a separate feature request.

Can you describe more about it?

tc0312:
Sorry I think I misunderstood you. Do you mean adding simple user/password auth to livy-server?

jerryshao:
Yes, that's what I was thinking, from my understanding only auth in client side may not be enough.

tc0312:
I'm considering it as a separate feature request. The intention of this PR is to add user/password auth to LivyClient because some of our users are putting Livy behind a security gateway like Knox.

I agree we should add other authentication methods to livy-server. It's a complicated matters on its own. What authentication should we add? Should we use something like Shiro? I think it's better do it separately.

jerryshao:
I see, thanks.

>I agree we should add other authentication methods to livy-server. It's a complicated matters on its own. 

I'm not saying we should add other auths. My point is that since you add a client side auth, My thinking is that it is necessary to a corresponding one in the server side also. Since you mentioned this will be handled by other service, then I'm fine with it now.

Also for the feature completeness, should we also add such feature for python client?

tc0312:
Agree. Created https://issues.cloudera.org/browse/LIVY-277 for Python.

client-http/src/main/java/com/cloudera/livy/client/http/LivyConnection.java
```diff
@@ -90,24 +91,40 @@ public boolean isContentCompressionEnabled() {
...
+      String[] userInfo = uri.getUserInfo().split(":");
+      if (userInfo.length != 2) {
+        throw new IllegalArgumentException("Malformed user info in the url.");
       }
-
-      @Override
-      public Principal getUserPrincipal() {
-        return null;
+      try {
+        String username = URLDecoder.decode(userInfo[0], "UTF-8");
```
jerryshao: Minor: is it better to use `UTF_8.name` instead of string. In Spark such usage of string all changed to `UTF_8` object.

client-http/src/main/java/com/cloudera/livy/client/http/LivyConnection.java
```diff
@@ -90,24 +91,40 @@ public boolean isContentCompressionEnabled() {
...
     };
 
-    Credentials dummyCredentials = new Credentials() {
-      @Override
-      public String getPassword() {
-        return null;
+    Credentials credentials;
+    // If user info is specified in the url, pass them to the CredentialsProvider.
+    if (uri.getUserInfo() != null) {
+      String[] userInfo = uri.getUserInfo().split(":");
```
jerryshao: Do you enforce user to have both username and password in URL. As I know only username is acceptable in URL format. Should we also consider it?

jerryshao:
LGTM, just some minor comments.

jerryshao:
LGTM.

PR #246, opened 2016-12-07
LIVY-265. spark.matser can not be overridden

zjffdu:
Straightforward change to allow spark.master to be overridden when it is excluded in blacklist.


tc0312:
spark.master should never be overridden because all sessions in a Livy server must have the same master. This's a requirement from recovery and HA.

What's the purpose of this change?

zjffdu:
I see, because it can been overridden in 0.2. Then I will close this.

PR #247, opened 2016-12-07
LIVY-165. Added session heartbeat to help notebooks to prevent leakage.

tc0312:
Notebook applications like Jupyter might crash while livy is running. No one will clean up its corresponding Livy session and will be leaked.
Heartbeat is added to address this. To keep a session alive, the notebook application must continously make GET requests to the interactive session. If no GET request is made within the heartbeat interval, livy-server will delete the session regardless to its state (busy, idle).

Heartbeat is per session and is controlled by session property "heartbeatTimeoutInSecond". Its default is 0 and it means heartbeat is disabled. To enable heartbeat, please create the session with non-zero "heartbeatTimeoutInSecond" in the create request.


zjffdu:
Maybe we can use this heartbeat to return the job progress info to frontend.

tc0312:
I will merge this if there's no further comment.

PR #248, opened 2016-12-08
LIVY-245. Sharing variables across Scala jobs in RSC.

tc0312:


tc0312:
@jerryshao

jerryshao:
@tc0312 Can we hold on this feature request until after 0.3 release if it not so urgent?

Can we propose several different API design choice (no need to implement), so we could balance the pros and cons of each choice, the current design looks quite simple and straightforward, but we may expose too many things to user, what SJS did could be another choice (hide the details and expose user only the named variable). Let's list all the possible choice as we could think.

tc0312:
I was following this in SJS: https://github.com/spark-jobserver/spark-jobserver#named-objects

> Let's list all the possible choice as we could think.
Sure. Mind if you describe what you have in mind?

jerryshao:
I was wondering SJS may have sophisticated mechanism about this, but looks quite simple.

Another choice in my mind is to not expose data structure to user, only get or updated by name, like:

```
<E> E getSharedObject(String name)

<E> E updateSharedObject(String name, E object)
```

This maybe better for us to control the details of the implementation, or change the underlying data structure later on.

Also I think we should take care some other things:

1. Who is responsible for variable cleaning or removing. User may forget it and SC will probably be OOM. Besides Spark itself uses weak reference to handle cached RDD, so can we use the similar mechanism.
2. How to do user isolation for shared data. For example user A may cache a variable, should it be accessed by user B?

tc0312:
> Who is responsible for variable cleaning or removing. User may forget it and SC will probably be OOM. Besides Spark itself uses weak reference to handle cached RDD, so can we use the similar mechanism.

It's user's responsibility.

> How to do user isolation for shared data. For example user A may cache a variable, should it be accessed by user B?

In the current design, each session is owned by a user and is not shareable.

PR #250, opened 2016-12-09
LIVY-270. %table shouldn't treat None as its own type.

purechoc:
```
x = [{"name":"a", "value":None}, {"name":"b", "value":2}]
%table x
```
return error `table rows have different types`
if input value is list, tuple. dict type, need to be allow None value.

### DIFF
--- a/repl/src/main/resources/fake_shell.py
+++ b/repl/src/main/resources/fake_shell.py
@@ -426,11 +426,14 @@ def magic_table(name):
                 }
                 headers[name] = header
             else:
-                # Reject columns that have a different type.
-                if header['type'] != col_type:
-                    exc_type = Exception
-                    exc_value = 'table rows have different types'
-                    return execute_reply_error(exc_type, exc_value, None)
+                # Reject columns that have a different type. (allow none value)
+                if col_type is not "NULL_TYPE" and header['type'] != col_type:
+                    if header['type'] is "NULL_TYPE":
+                        header['type'] = col_type
+                    else:
+                        exc_type = Exception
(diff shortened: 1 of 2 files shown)


purechoc:
meet R test bug.

repl/src/main/resources/fake_shell.py
```diff
@@ -426,11 +426,14 @@ def magic_table(name):
                 }
                 headers[name] = header
             else:
-                # Reject columns that have a different type.
-                if header['type'] != col_type:
-                    exc_type = Exception
-                    exc_value = 'table rows have different types'
-                    return execute_reply_error(exc_type, exc_value, None)
+                # Reject columns that have a different type. (allow none value)
+                if col_type is not "NULL_TYPE" and header['type'] != col_type:
```
tc0312: These comparisons are using `is` but not `=`. Will these checks ever pass?

purechoc: change "is not" -> "!=" 
this is exception case : if header column type is not null, and column type different type to header.

if i'm wrong understand your question. comment again plaese.

purechoc: i think, i had a mistake.  i'll fix that condition soon.

tc0312:
Mind if you add a test case for
```
x = [{"a":None, "b":None}, {"a":"2", "b":2}]
```
I've a feeling it will not work.

purechoc:
@tc0312 
thank to comment. 
added test and that case is success.  it's work.

PR #251, opened 2016-12-09
LIVY-261. Fixed SparkRInterpreter failed to include error message randomly.

tc0312:
- Redirect stderr to stdout to avoid synchronization between stdout and stderr.


zjffdu:
lgtm

PR #252, opened 2016-12-12
LIVY-271. Fixed repl session to return traceback as a JSON list in error cases.

zjffdu:
traceback could be json list or json object which is misleading. So this PR would not include traceback when it is null. 

Tested manually.

### DIFF
--- a/repl/src/main/scala/com/cloudera/livy/repl/Session.scala
+++ b/repl/src/main/scala/com/cloudera/livy/repl/Session.scala
@@ -132,7 +132,8 @@ class Session(interpreter: Interpreter, stateChangedCallback: SessionState => Un
           (STATUS -> ERROR) ~
           (EXECUTION_COUNT -> executionCount) ~
           (ENAME -> "Error") ~
-          (EVALUE -> "incomplete statement")
+          (EVALUE -> "incomplete statement") ~
+          (TRACEBACK -> Seq.empty[String])
 
         case Interpreter.ExecuteError(ename, evalue, traceback) =>
           transitToIdle()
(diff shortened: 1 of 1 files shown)


zjffdu:
@tc0312 @jerryshao Please help review

jerryshao:
LGTM.

tc0312:
Traceback should always be a JSON List.
Mind if you provide a repro when it returns a JSON Object? Let's fix that bug instead of this.

zjffdu:
If you send code "sc.", you will get the following output. 
```
{
  "total_statements": 1,
  "statements": [
    {
      "id": 0,
      "state": "available",
      "output": {
        "status": "error",
        "execution_count": 0,
        "ename": "Error",
        "evalue": "incomplete statement",
        "traceback": {}
      }
    }
  ]
}
```

After this PR, we could get
```
    {
      "id": 1,
      "state": "available",
      "output": {
        "status": "error",
        "execution_count": 1,
        "ename": "Error",
        "evalue": "incomplete statement"
      }
    }
```

tc0312:
Can we fix that to return `[]` instead? This PR will break Jupyter integration.

zjffdu:
@tc0312 I just pushed a new commit to return [] for empty traceback.
```
{
  "total_statements": 1,
  "statements": [
    {
      "id": 0,
      "state": "available",
      "output": {
        "status": "error",
        "execution_count": 0,
        "ename": "Error",
        "evalue": "incomplete statement",
        "traceback": []
      }
    }
  ]
}
```

tc0312:
Thank you. Will merge once test passes.

PR #253, opened 2016-12-12
LIVY-267. Honor fixed port number for PRC server

jerryshao:
Current RPC server doesn't honor port number settings, it will pick a random port number, which will lead to some issues as described in [LIVY-267](https://issues.cloudera.org/projects/LIVY/issues/LIVY-267?filter=allopenissues). So here propose to fix it:

1. Separate configurations for two RPC server, one is launcher, another is rsc driver's rpc server.
2. Remove `livy.rsc.rpc.server.address` since current it sets null deliberately, so removing it for simplicity.
3. Add `livy.rsc.rpc.server.port` configuration to fix the port number of rsc driver RPC server.

### DIFF
--- a/rsc/src/main/java/com/cloudera/livy/rsc/driver/RSCDriver.java
+++ b/rsc/src/main/java/com/cloudera/livy/rsc/driver/RSCDriver.java
@@ -172,7 +172,7 @@ public class RSCDriver extends BaseProtocol {
 
     // Bring up the RpcServer an register the secret provided by the Livy server as a client.
     LOG.info("Starting RPC server...");
-    this.server = new RpcServer(livyConf, false);
+    this.server = new RpcServer(livyConf, null, livyConf.getInt(RPC_SERVER_PORT));
     server.registerClient(clientId, secret, new RpcServer.ClientCallback() {
       @Override
       public RpcDispatcher onNewClient(Rpc client) {
--- a/rsc/src/main/java/com/cloudera/livy/rsc/driver/RSCDriver.java
+++ b/rsc/src/main/java/com/cloudera/livy/rsc/driver/RSCDriver.java
@@ -172,7 +172,7 @@ public class RSCDriver extends BaseProtocol {
 
     // Bring up the RpcServer an register the secret provided by the Livy server as a client.
     LOG.info("Starting RPC server...");
-    this.server = new RpcServer(livyConf, null, livyConf.getInt(RPC_SERVER_PORT));
+    this.server = new RpcServer(livyConf, livyConf.getInt(RPC_SERVER_PORT));
     server.registerClient(clientId, secret, new RpcServer.ClientCallback() {
       @Override
       public RpcDispatcher onNewClient(Rpc client) {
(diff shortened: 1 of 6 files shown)


zjffdu:
@jerryshao You need to rebase it.

rsc/src/main/java/com/cloudera/livy/rsc/rpc/RpcServer.java
```diff
@@ -67,11 +68,35 @@
...
     this.config = lconf;
     this.group = new NioEventLoopGroup(
         this.config.getInt(RPC_MAX_THREADS),
         Utils.newDaemonThreadFactory("RPC-Handler-%d"));
+
+    final InetSocketAddress socketAddress;
+    if (isLauncher) {
+      if (lconf.get(LAUNCHER_ADDRESS) != null && lconf.getInt(LAUNCHER_PORT) > 0) {
+        socketAddress =
+          new InetSocketAddress(lconf.get(LAUNCHER_ADDRESS), lconf.getInt(LAUNCHER_PORT));
```
jerryshao: I don't change the logic, `LAUNCHER_PORT` still used for driver to connect back to livy server.

rsc/src/main/java/com/cloudera/livy/rsc/rpc/RpcServer.java
```diff
@@ -67,11 +68,35 @@
   private final ConcurrentMap<String, ClientInfo> pendingClients;
   private final RSCConf config;
 
-  public RpcServer(RSCConf lconf) throws IOException, InterruptedException {
+  @VisibleForTesting
+  RpcServer(RSCConf lconf) throws IOException, InterruptedException {
+    this(lconf, false);
+  }
+
+  public RpcServer(RSCConf lconf, boolean isLauncher) throws IOException, InterruptedException {
```
tc0312: The code can be simplified a lot if we pass `int port` to the constructor instead of `boolean isLauncher`.

rsc/src/main/java/com/cloudera/livy/rsc/rpc/RpcServer.java
```diff
@@ -67,11 +68,25 @@
   private final ConcurrentMap<String, ClientInfo> pendingClients;
   private final RSCConf config;
 
-  public RpcServer(RSCConf lconf) throws IOException, InterruptedException {
+  @VisibleForTesting
+  RpcServer(RSCConf lconf) throws IOException, InterruptedException {
+    this(lconf, null, 0);
+  }
+
+  public RpcServer(RSCConf lconf, String address, int port)
```
tc0312: Who's using this constructor?

rsc/src/main/java/com/cloudera/livy/rsc/driver/RSCDriver.java
```diff
@@ -178,7 +172,7 @@ private void initializeServer() throws Exception {
 
     // Bring up the RpcServer an register the secret provided by the Livy server as a client.
     LOG.info("Starting RPC server...");
-    this.server = new RpcServer(livyConf);
+    this.server = new RpcServer(livyConf, null, livyConf.getInt(RPC_SERVER_PORT));
```
tc0312: We can define a constructor for this use case and for testing, if there's any:
```
public RpcServer(RSCConf lconf, int port)
```

jerryshao:
I agreed that "livy.rsc.launcher.address" and "livy.rsc.launcher.port" are more like a system level configuration, should be configured in livy.conf rather then request conf.

For Launcher RPC server, it is only used for rsc to find out LivyServer, I think it is OK to use one specified port to bind. Also shared rpc server (with reference counting) should be enough.

tc0312:
Would you mind if we close this for now and re-open it later?

jerryshao:
Sure.

PR #255, opened 2016-12-15
LIVY-269. Fixed SparkRInterpreter hangs forever on executing `x[`.

tc0312:


tc0312:
I will merge this if there's no further comment.

PR #257, opened 2016-12-16
Minor. Fix flaky test for interactive session state

jerryshao:
due to the latency of session state update, we may get `Busy` instead of `Idle`, please see [here](https://travis-ci.org/cloudera/livy/jobs/184456364). So here using `eventually` to fix it.


tc0312:
LGTM. What was the last test failure?

PR #258, opened 2016-12-19
LIVY-279. Bump supported Spark version to 2.1.x

jerryshao:
Spark 2.1.0 will be released soon, we should also bump the supported version to 2.1.x.


zjffdu:
LGTM

tc0312:
Should release 0.3 include this?

jerryshao:
Yes, it would be better to. HDP 2.6 will adopt Spark 2.1, and we will need to support this version in Livy.

zjffdu:
@tc0312 Spark 2.1 is compatible with spark 2.0. I will merge it if no more discussion.

PR #259, opened 2016-12-19
LIVY-280. Livy session leakage issue when app submission timeout

zjffdu:
* Add a new daemon thread to kill leaked livy session regularly.  
* Change the default look-up timeout to 60s.

### DIFF
--- a/conf/livy.conf
+++ b/conf/livy.conf
@@ -76,9 +76,17 @@
 
 # If Livy can't find the yarn app within this time, consider it lost.
 # livy.server.yarn.app-lookup-timeout = 30s
+# When the cluster is busy, we may fail to launch yarn app in 30s, then it would cause session
+# leakage, so we need to check session leakage.
+# How long to check livy session leakage
+# livy.server.yarn.app-leakage.check_timeout = 600s
+# how often to check livy session leakage
+# livy.server.yarn.app-leakage.check_interval = 60s
 
 # How often Livy polls YARN to refresh YARN app state.
 # livy.server.yarn.poll-interval = 1s
 #
 # Days to keep Livy server request logs.
 # livy.server.request-log-retain.days = 5
+
+
--- a/conf/livy.conf
+++ b/conf/livy.conf
@@ -75,7 +75,7 @@
 # livy.server.recovery.state-store.url =
 
 # If Livy can't find the yarn app within this time, consider it lost.
-# livy.server.yarn.app-lookup-timeout = 30s
+# livy.server.yarn.app-lookup-timeout = 60s
 # When the cluster is busy, we may fail to launch yarn app in 30s, then it would cause session
 # leakage, so we need to check session leakage.
 # How long to check livy session leakage
@@ -88,5 +88,3 @@
 #
 # Days to keep Livy server request logs.
 # livy.server.request-log-retain.days = 5
-
-
--- a/conf/livy.conf
+++ b/conf/livy.conf
@@ -76,8 +76,8 @@
 
 # If Livy can't find the yarn app within this time, consider it lost.
 # livy.server.yarn.app-lookup-timeout = 60s
-# When the cluster is busy, we may fail to launch yarn app in 30s, then it would cause session
-# leakage, so we need to check session leakage.
+# When the cluster is busy, we may fail to launch yarn app in app-lookup-timeout, then it would
+# cause session leakage, so we need to check session leakage.
 # How long to check livy session leakage
 # livy.server.yarn.app-leakage.check_timeout = 600s
 # how often to check livy session leakage
--- a/server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
+++ b/server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
@@ -90,6 +90,7 @@ class LivyServer extends Logging {
 
     // Initialize YarnClient ASAP to save time.
     if (livyConf.isRunningOnYarn()) {
+      SparkYarnApp.init(livyConf)
       Future { SparkYarnApp.yarnClient }
     }
 
--- a/server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
+++ b/server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
@@ -41,8 +41,9 @@ object SparkYarnApp extends Logging {
   def init(livyConf: LivyConf): Unit = {
     sessionLeakageCheckInterval = livyConf.getTimeAsMs(LivyConf.YARN_APP_LEAKAGE_CHECK_INTERVAL)
     sessionLeakageCheckTimeout = livyConf.getTimeAsMs(LivyConf.YARN_APP_LEAKAGE_CHECK_TIMEOUT)
-    leakagedAppGCThread.setDaemon(true)
-    leakagedAppGCThread.start()
+    leakedAppsGCThread.setDaemon(true)
+    leakedAppsGCThread.setName("LeakedAppsGCThread")
+    leakedAppsGCThread.start()
   }
 
   // YarnClient is thread safe. Create once, share it across threads.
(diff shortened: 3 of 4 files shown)


jerryshao:
Did it fixe by this [commit](https://github.com/cloudera/livy/commit/e82780b96affc4d1cf8a1d08d41ac98844c2af62)?

zjffdu:
No. LIVY-251 resolve the session leakage when yarn app is launched but stuck in ACCEPTED state. While this PR fix the issue when even AM is not started due to resource limitation.

jerryshao:
I think LIVY-251 can also kill the stuck application in which AM cannot be started due to resource limitation. I tried before.

zjffdu:
I don't see the yarn app is killed. You can set `livy.server.yarn.app-lookup-timeout` to `5s` to reproduce the issue.

zjffdu:
@tc0312 @linchan-ms Please help review.

tc0312:
From our testing, LIVY-251, works also when AM is not started. Do you have a reliable way to repro the leakage you are seeing? May be there's another edge case.

zjffdu:
@tc0312 I can reproduce this issue even with LIVY-251. You can adjust livy.server.yarn.app-lookup-timeout and livy.server.yarn.poll-interval to make the app submission fails. And I don't see LIVY-251 handle the situation I mentioned here. 

This is the error what I see and the yarn app is still running.

```
{
  "from": 0,
  "total": 1,
  "sessions": [
    {
      "id": 0,
      "appId": null,
      "owner": null,
      "proxyUser": null,
      "state": "dead",
      "kind": "spark",
      "appInfo": {
        "driverLogUrl": null,
        "sparkUiUrl": null
      },
      "log": [
        "java.lang.Exception: No YARN application is tagged with livy-session-0-zyl0mqvy.",
        "com.cloudera.livy.utils.SparkYarnApp.com$cloudera$livy$utils$SparkYarnApp$$getAppIdFromTag(SparkYarnApp.scala:131) com.cloudera.livy.utils.SparkYarnApp$$anonfun$1$$anonfun$4.apply(SparkYarnApp.scala:195) com.cloudera.livy.utils.SparkYarnApp$$anonfun$1$$anonfun$4.apply(SparkYarnApp.scala:192) scala.Option.getOrElse(Option.scala:120) com.cloudera.livy.utils.SparkYarnApp$$anonfun$1.apply$mcV$sp(SparkYarnApp.scala:192) com.cloudera.livy.Utils$$anon$1.run(Utils.scala:95)"
      ]
    }
  ]
}
```

jerryshao:
@zjffdu , I think I could reproduce your issue, this is a different issue compared to LIVY-251. In your case the problem is mainly due to the timeout of this API `getAppIdFromTag`, looks like spark hasn't yet request application id from RM when this API was timed-out.

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
```diff
@@ -50,6 +58,47 @@ object SparkYarnApp extends Logging {
...
+            val entry = iter.next()
+            yarnClient.getApplications(appType).asScala
+              .find(_.getApplicationTags.contains(entry.getKey))
+              .foreach(e => {
+                info(s"Kill leakage app ${e.getApplicationId}")
+                yarnClient.killApplication(e.getApplicationId)
+                iter.remove()
+                isRemoved = true
+              })
+            if (!isRemoved) {
```
jerryshao: When will we go into this branch?

zjffdu: It happens when AM still doesn't start.

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
```diff
@@ -50,6 +58,47 @@ object SparkYarnApp extends Logging {
 
   private def getYarnPollInterval(livyConf: LivyConf): FiniteDuration =
     livyConf.getTimeAsMs(LivyConf.YARN_POLL_INTERVAL) milliseconds
+
+  private val appType = Set("SPARK").asJava
+
+  private val leakagedAppTags = new java.util.concurrent.ConcurrentHashMap[String, Long]()
```
tc0312: leakagedAppTags is not persisted across restart. We should persist it to the state store.

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
```diff
@@ -50,6 +58,47 @@ object SparkYarnApp extends Logging {
...
+
+  private val leakagedAppGCThread = new Thread() {
+    override def run(): Unit = {
+      while (true) {
+        if (!leakagedAppTags.isEmpty) {
+          // kill the app if found it and remove it if exceeding a threashold
+          val iter = leakagedAppTags.entrySet().iterator()
+          var isRemoved = false
+          val now = System.currentTimeMillis()
+          while(iter.hasNext) {
```
tc0312: Is it possible to batch the getApplications() calls? I think we can do it by using RMClient directly without going thru YarnClient.

zjffdu: I move `yarnClient.getApplications` to the outside of while loop so that we only call it one time.  LIVY-214 is only for hadoop-2.9, we can do it in another PR.

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
zjffdu: I tried it, but it looks like it is a little complicated than I expect. Because it may affect the code of batch session recovering. The remaining issue is that the leaked session can not be cleanup after recovery, I think this is not critical, so we can do it in a follow up ticket, what do you think ?

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
```diff
@@ -50,6 +59,47 @@ object SparkYarnApp extends Logging {
...
+        if (!leakedAppTags.isEmpty) {
+          // kill the app if found it and remove it if exceeding a threashold
+          val iter = leakedAppTags.entrySet().iterator()
+          var isRemoved = false
+          val now = System.currentTimeMillis()
+          val apps = yarnClient.getApplications(appType).asScala
+          while(iter.hasNext) {
+            val entry = iter.next()
+            apps.find(_.getApplicationTags.contains(entry.getKey))
+              .foreach(e => {
```
jerryshao: Mind if you change to

```
.foreach { e =>
...
...
}
```

server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
```diff
@@ -90,6 +90,7 @@ class LivyServer extends Logging {
 
     // Initialize YarnClient ASAP to save time.
     if (livyConf.isRunningOnYarn()) {
+      SparkYarnApp.init(livyConf)
```
jerryshao: It is not necessary to have a `init(...)` function to create and start the GC thread. We could directly initialize the values inside the `SparkYarnApp`. This values will be initialized when `SparkYarnApp` class is loaded, which is here calling `SparkYarnApp.yarnClient`.

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
jerryshao: Can you please create a JIRA to track this issue.

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
zjffdu: addressed

zjffdu:
@tc0312 Any more comments ?

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
tc0312: If we use RMClient directly, I think we can do it without depending on YARN 2.9.

conf/livy.conf
```diff
@@ -75,7 +75,13 @@
 # livy.server.recovery.state-store.url =
 
 # If Livy can't find the yarn app within this time, consider it lost.
-# livy.server.yarn.app-lookup-timeout = 30s
+# livy.server.yarn.app-lookup-timeout = 60s
+# When the cluster is busy, we may fail to launch yarn app in 30s, then it would cause session
```
tc0312: +# When the cluster is busy, we may fail to launch yarn app in app-lookup-timeout, then it would cause session

tc0312:
Mind if you add an integration test for this?

zjffdu:
It seems not easy to simulate the session leakage in integration test, I will try to do that in other ticket.

zjffdu:
Merge it and create LIVY-301 as followup

PR #261, opened 2016-12-22
LIVY-282. Fix fake_shell unexpectedly exit problem

jerryshao:
Due to the bug introduced in LIVY-233, pyspark interpreter will be failed to start in Spark 1.x, in which `SparkSession` is not supported. So here fix this problem.


jerryshao:
@tc0312 @zjffdu , this is a block issue, current pyspark session is not worked, please review. Thanks a lot.

zjffdu:
LGTM

PR #263, opened 2016-12-26
LIVY-284. Make Livy server request log configurable

jerryshao:


server/src/main/scala/com/cloudera/livy/LivyConf.scala
```diff
@@ -111,6 +111,9 @@ object LivyConf {
   // How often Livy polls YARN to refresh YARN app state.
   val YARN_POLL_INTERVAL = Entry("livy.server.yarn.poll-interval", "5s")
 
+  // Days to keep Livy web server request logs.
+  val REQUEST_LOG_RETAIN_DAYS = Entry("livy.server.request-log-retain.days", 5)
```
zjffdu: Need to add in livy.conf as well.

jerryshao: Done.

PR #266, opened 2016-12-30
Provide Jetty Http Request/Response Header Size Configuration

brockn:
Kerberos over http requires larger header sizes than the default,
or the HTTP Error 413 Request entity too large will be returned.
This patch increases the default for Livy to 128K and also allows
this to be configurable.

### DIFF
--- a/server/src/main/scala/com/cloudera/livy/LivyConf.scala
+++ b/server/src/main/scala/com/cloudera/livy/LivyConf.scala
@@ -58,6 +58,8 @@ object LivyConf {
 
   val SERVER_HOST = Entry("livy.server.host", "0.0.0.0")
   val SERVER_PORT = Entry("livy.server.port", 8998)
+  val HTTP_REQUEST_HEADER_SIZE = Entry("livy.http.request.header.size", 131072)
+  val HTTP_RESPONSE_HEADER_SIZE = Entry("livy.http.response.header.size", 131072)
   val CSRF_PROTECTION = LivyConf.Entry("livy.server.csrf_protection.enabled", false)
 
   val IMPERSONATION_ENABLED = Entry("livy.impersonation.enabled", false)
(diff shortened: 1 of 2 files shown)


server/src/main/scala/com/cloudera/livy/LivyConf.scala
```diff
@@ -58,6 +58,8 @@ object LivyConf {
 
   val SERVER_HOST = Entry("livy.server.host", "0.0.0.0")
   val SERVER_PORT = Entry("livy.server.port", 8998)
+  val HTTP_REQUEST_HEADER_SIZE = Entry("livy.http.request.header.size", 131072)
+  val HTTP_RESPONSE_HEADER_SIZE = Entry("livy.http.response.header.size", 131072)
```
jerryshao: Can we rename the configurations to "livy.server.request_header.size" and "livy.server.response_header.size", also it would be better to add some comments for these two configurations to describe why we have to change the default size.

server/src/main/scala/com/cloudera/livy/server/WebServer.scala
```diff
@@ -39,12 +39,20 @@ class WebServer(livyConf: LivyConf, var host: String, var port: Int) extends Log
...
-      (new ServerConnector(server), "http")
+      val http = new HttpConfiguration()
+      http.setRequestHeaderSize(requestHeaderSize)
+      http.setResponseHeaderSize(responseHeaderSize)
+      (new ServerConnector(server, new HttpConnectionFactory(http)), "http")
 
     case Some(keystore) =>
       val https = new HttpConfiguration()
+      https.setRequestHeaderSize(requestHeaderSize)
+      https.setResponseHeaderSize(responseHeaderSize)
```
jerryshao: Can we move this `HttpConfiguration` to above to remove the duplications.

server/src/main/scala/com/cloudera/livy/LivyConf.scala
tc0312: livy.server.request-header.size
livy.server.response-header.size

tc0312:
Mind if you create a JIRA for this?

jerryshao:
@brockn Would you mind addressing the comments mentioned above?

zjffdu:
@brockn Any update ?

PR #267, opened 2017-01-03
LIVY-286. Update the doc to align with 0.3 code

jerryshao:


jerryshao:
\cc @tc0312 @zjffdu

tc0312:
Thanks so much for this. LGTM.

README.rst
```diff
@@ -582,19 +639,60 @@ Response Body
...
+
+Cancel the specified statement in this session.
+
+Response Body
+^^^^^^^^^^^^^
+
++------+----------------------------+--------+
+| name | description                | type   |
++======+============================+========+
+| msg  | denote cancelled statement | string |
```
tc0312: "cancelled" if the statement has been sucessfully cancelled.

README.rst
```diff
@@ -582,19 +639,60 @@ Response Body
...
+Returns all the active batch sessions.
 
-Returns all the active batch jobs.
+Request Parameters
+^^^^^^^^^^^^^^^^^^
+
++------+-----------------------------------+------+
+| name | description                       | type |
++======+===================================+======+
+| from | The start session id to fetch     | int  |
```
tc0312: I don't think this's right. It's not a session id. It's the the index of the first shown session id in SessionServlet.sessions array.

However, I think what you are suggesting makes sense. We should change the code to match this behavior.

README.rst
```diff
@@ -582,19 +639,60 @@ Response Body
...
 
-+---------+---------------+------+
-| name    | description   | type |
-+=========+===============+======+
-|sessions | `batch`_ list | list |
-+---------+---------------+------+
++----------+-------------------------------+------+
+| name     | description                   | type |
++==========+===============================+======+
+| from     | The start session id fetched  | int  |
```
tc0312: Ditto

README.rst
```diff
@@ -783,15 +914,19 @@ A statement represents the result of an execution statement.
...
++------------+----------------------------------+
+| value      | description                      |
++============+==================================+
+| running    | Statement is currently running   |
++------------+----------------------------------+
+| available  | Statement has a response ready   |
++------------+----------------------------------+
+| error      | Statement failed                 |
++------------+----------------------------------+
+| cancelling | Statement is cancelling          |
```
tc0312: Statement is being cancelled

README.rst
jerryshao: In the current code, it will always return "cancelled" as the response of this request. User should query statement state to get the actual state of this statement. 

The code be improved, but I would like to make the doc correctly reflect the current code.

README.rst
jerryshao: I see, let me update the doc first. We could change the code late on.

README.rst
tc0312: Um.. how about 'is always cancelled'

jerryshao: Sure, will change it.

README.rst
tc0312: https://issues.cloudera.org/browse/LIVY-291

PR #268, opened 2017-01-04
LIYV-289. Fix NPE in ContextLauncher and expose state of RSCClient

jerryshao:
If `ContextLauncher#startDriver` doesn't start correctly, it will trigger the listener with `child` be empty, this will introduce NPE, so here check whether `child` is null or not. Also expose the state of `RSCClient`.
 
This is originated by @mridulm , please review. \cc @zjffdu .


zjffdu:
LGTM, could you update the title and description ? @jerryshao

jerryshao:
Ping @zjffdu .

PR #269, opened 2017-01-06
LIVY-292. Fix Python Job API issue and add Java/python Job API example

jerryshao:
Add two Java / Python Job API examples to Livy to demonstrate how to use.

Also Fix an issue of Python Job API where interpreter is not created when Job request is received in RSCDriver, this will lead to kryo deserialization error.

### DIFF
--- a/examples/src/main/java/com/cloudera/livy/examples/PiApp.java
+++ b/examples/src/main/java/com/cloudera/livy/examples/PiApp.java
@@ -1,0 +1,102 @@
+/*
+ * Licensed to Cloudera, Inc. under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  Cloudera, Inc. licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.cloudera.livy.examples;
+
+import java.io.File;
+import java.net.URI;
+import java.util.ArrayList;
+import java.util.List;
+
+import org.apache.spark.api.java.function.Function;
+import org.apache.spark.api.java.function.Function2;
+
+import com.cloudera.livy.*;
+
+class PiJob implements
+    Job<Double>,
+    Function<Integer, Integer>,
+    Function2<Integer, Integer, Integer> {
+
+  private final int slices;
+  private final int samples;
+
+  public PiJob(int slices) {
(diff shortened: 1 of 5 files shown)


repl/src/main/scala/com/cloudera/livy/repl/ReplDriver.scala
```diff
@@ -94,20 +94,23 @@ class ReplDriver(conf: SparkConf, livyConf: RSCConf)
   }
 
   override protected def createWrapper(msg: BaseProtocol.BypassJobRequest): BypassJobWrapper = {
-    interpreter match {
-      case pi: PythonInterpreter => pi.createWrapper(this, msg)
```
jerryshao: If Job is submitted before interpreter is fully launched, here `interpreter` will be null. This will offload `createWrapper` to `super.xxx`, and it will use kryo to deserialize pickle serialized job, this will lead to kryo exception.

jerryshao:
\CC @tc0312 @zjffdu please help to review, thanks a lot.

examples/src/main/python/pi_app.py
```diff
@@ -0,0 +1,48 @@
...
+from random import random
+from operator import add
+
+from livy.client import HttpClient
+
+if __name__ == "__main__":
+    """
+        Usage: PiJob [livy url] [slices]
+    """
+    slices = int(sys.argv[2])
```
zjffdu: it's better to check argument number. If it is not 2, then print usage.

examples/src/main/python/pi_app.py
```diff
@@ -0,0 +1,48 @@
+from __future__ import print_function
```
zjffdu: put this line after the licence header

examples/src/main/python/pi_app.py
```diff
@@ -0,0 +1,48 @@
...
+# limitations under the License.
+#
+
+import sys
+from random import random
+from operator import add
+
+from livy.client import HttpClient
+
+if __name__ == "__main__":
```
zjffdu: Add one comment about how to launch this python app like what you did in java example.

examples/src/main/python/pi_app.py
jerryshao: I referred this from Spark, not sure why Spark needs to put this in the top of file.

examples/src/main/java/com/cloudera/livy/examples/PiApp.java
```diff
@@ -0,0 +1,102 @@
...
+    return v1 + v2;
+  }
+}
+
+/**
+ * Example execution:
+ * java -cp /pathTo/spark-core_2.10-*version*.jar:/pathTo/livy-api-*version*.jar:
+ * /pathTo/livy-client-http-*version*.jar:/pathTo/livy-examples-*version*.jar
+ * com.cloudera.livy.examples.PiApp http://livy-host:8998 2
+ */
```
namitkabra: It will be nice to mention here (not sure if it is mentioned elsewhere) that the livy-examples-*version*.jar should be on hdfs.

jerryshao: livy example jar will be uploaded to hdfs inside the code, please see here (https://github.com/cloudera/livy/pull/269/files#diff-5dc2c9601d6afd761bf75b667d4f3abaR88).

PR #270, opened 2017-01-10
LIVY-294. HiveContext is always created instead of SQLContext for pyspark

zjffdu:
HiveContext is always created no matter whether we enable hiveContext through `spark.repl.enableHiveContext`. The root cause is that we depends on `shell.py` of spark. and unformatenlly HiveContext would not initialize itself when created, but defer its initialization until any methods is called. This PR would call `sqlContext.tables()` to check whether hiveContext can work properly.

Tested manully

### DIFF
--- a/repl/src/main/resources/fake_shell.py
+++ b/repl/src/main/resources/fake_shell.py
@@ -30,6 +30,7 @@ import threading
 import tempfile
 import shutil
 import pickle
+import textwrap
 
 if sys.version >= '3':
     unicode = str
@@ -541,12 +542,13 @@ def main():
                 # LIVY-294, need to check whether HiveContext can work properly,
                 # fallback to SQLContext if HiveContext can not be initialized successfully.
                 # Only for spark-1.
-                code = """import py4j
-from pyspark.sql import SQLContext
-try:
-  sqlContext.tables()
-except py4j.protocol.Py4JError:
-  sqlContext = SQLContext(sc)"""
+                code = textwrap.dedent("""
+                    import py4j
+                    from pyspark.sql import SQLContext
+                    try:
+                      sqlContext.tables()
+                    except py4j.protocol.Py4JError:
+                      sqlContext = SQLContext(sc)""")
                 exec(code, global_dict)
 
             #Start py4j callback server
--- a/repl/src/main/resources/fake_shell.py
+++ b/repl/src/main/resources/fake_shell.py
@@ -534,8 +534,20 @@ def main():
             exec('from pyspark.sql import HiveContext', global_dict)
             exec('from pyspark.streaming import StreamingContext', global_dict)
             exec('import pyspark.cloudpickle as cloudpickle', global_dict)
+
             if spark_major_version >= "2":
                 exec('from pyspark.shell import spark', global_dict)
+            else:
+                # LIVY-294, need to check whether HiveContext can work properly,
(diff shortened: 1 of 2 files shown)


zjffdu:
@tc0312 @jerryshao Please help review

jerryshao:
I'm wondering can we do this in fake_shell.py?

repl/src/main/scala/com/cloudera/livy/repl/PythonInterpreter.scala
```diff
@@ -205,13 +207,37 @@ private class PythonInterpreter(process: Process, gatewayServer: GatewayServer,
...
+    // LIVY-294, need to check whether HiveContext can work properly, fallback to SQLContext if
+    // HiveContext can not be initialized successfully. Only for spark-1.
+    val code =
+      """
+        |import py4j
+        |from pyspark.sql import SQLContext
+        |try:
+        |  sqlContext.tables()
+        |except py4j.protocol.Py4JError:
+        |    sqlContext = SQLContext(sc)
```
jerryshao: Can we handle this in fake_shell's main function, I think that place would be more proper to handle compared here.

jerryshao: Also we should check whether `LIVY_TEST` is configured, this should not be created if running in unit test.

zjffdu:
I tried in fake_shell.py, but got some weird issue. Ideally I think livy has its own pyspark bootstrap script rather than depends on shell.py which may break from version to version. But it could be done in a follow up ticket, this is only for quick fix.

jerryshao:
But with your fix, do you have conflicts with the one already created here in fake_shell.py:

```
exec('from pyspark.shell import sqlContext', global_dict)
```

I would suggest to make this fix self-completed, even for a quick fix. Putting this in Scala code is not so elegant to me.

zjffdu:
It won't as I only created sqlContext when `sqlContext.tables()` fails

repl/src/main/resources/fake_shell.py
```diff
@@ -534,8 +534,20 @@ def main():
...
+            else:
+                # LIVY-294, need to check whether HiveContext can work properly,
+                # fallback to SQLContext if HiveContext can not be initialized successfully.
+                # Only for spark-1.
+                code = """import py4j
+from pyspark.sql import SQLContext
+try:
+  sqlContext.tables()
+except py4j.protocol.Py4JError:
+  sqlContext = SQLContext(sc)"""
```
jerryshao: Do we have a way to handle this weird indention issue?

zjffdu: didn't find a right approach. Previous I thought the following approach would work, but unfortunately, it doesn't

```
                code = """import py4j
                             ...:from pyspark.sql import SQLContext
                             ...:try:
                             ...:  sqlContext.tables()
                             ...:except py4j.protocol.Py4JError:
                             ...:  sqlContext = SQLContext(sc)"""
```

jerryshao: How about using "\n" "\t" to handle this.

zjffdu: wouldn't this more readable than '\t', '\n' ?

jerryshao: But here we're not talking about readability, just indention issue.

jerryshao:
LGTM for the fix.

PR #273, opened 2017-01-11
LIVY-298 - hive-site.xml added in spark.jars. not spark.files

purechoc:
in currently livy,
when hive-site.xml in livy classpath, livy added hive-site.xml in SPARK_FILES.
ex) spark.files /home/livy/livy-server/conf/hive-site.xml
execute code,
executor find hive-site.xml in same classpath. (/home/livy/livy-server/conf/hive-site.xml)
but cluster node do not has "/home/livy/livy-server/conf/hive-site.xml)

so hive-site.xml is added in spark.jars. not spark.files

### DIFF
--- a/server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
+++ b/server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
@@ -288,7 +288,7 @@ object InteractiveSession extends Logging {
         case (Some(file), false) =>
           debug("Enable HiveContext because hive-site.xml is found under classpath, "
             + file.getAbsolutePath)
-          mergeConfList(List(file.getAbsolutePath), LivyConf.SPARK_FILES)
+          mergeConfList(List(file.getAbsolutePath), LivyConf.SPARK_JARS)
           mergeConfList(datanucleusJars(livyConf, sparkMajorVersion), LivyConf.SPARK_JARS)
         case (None, false) =>
           warn("Enable HiveContext but no hive-site.xml found under" +


server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
```diff
@@ -288,7 +288,7 @@ object InteractiveSession extends Logging {
         case (Some(file), false) =>
           debug("Enable HiveContext because hive-site.xml is found under classpath, "
             + file.getAbsolutePath)
-          mergeConfList(List(file.getAbsolutePath), LivyConf.SPARK_FILES)
+          mergeConfList(List(file.getAbsolutePath), LivyConf.SPARK_JARS)
```
jerryshao: Hmmm, I think your fix is not correct. It should be added spark.files rather than spark.jars, since this is a file, not a jar.

jerryshao:
Can you please elaborate the problem you met, how you run Livy and Spark. I suspect the problem you met is unrelated to `SPARK_FILES` or `SPARK_JARS`.

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
purechoc: spark.files is "Comma-separated list of files to be placed in the working directory of each executor."
this mean,  hive-site.xml is not deploy with livy jars when yarn-cluster mode

zjffdu:
This might due to https://issues.apache.org/jira/browse/SPARK-18160
Can you try spark 2.1.0 ?

purechoc:
all cluster has /opt/commonlib/abc.file
user can use SPARK_FILES /opt/commonlib/abc.file
then, spark does'nt deploy abc.file, because abc.file is already exist all cluster node.

but want to deploy abc.file,  need to using SPARK_JAR

purechoc:
i'm try spark 1.6.4

jerryshao:
Yeah, might be related to the issue mentioned by @zjffdu . Files added to spark.files will be distributed to each container's local dir, so each executor could fetch the files with no issue. This behavior is the same as spark.jars running on yarn.

purechoc:
@zjffdu 
yes. your right. this is spark bug.
i try again spark 2.0.2.  no problem.
i just misunderstanding spark option.

zjffdu:
@purechoc Thanks for confirmation, I will close this PR.

PR #274, opened 2017-01-11
LIVY-300. Fixed a race condition in rsc's RPC channel.

tc0312:
In `Rpc.call()`, RPC function call messages are enqueued to the send buffer of the Channel on `Channel.write()` call because it's called outside of the netty Channel IO event loop.
RPC replies are handled by `RpcDispatcher.channelRead0`. It calls `RpcDispatcher.writeMessage()` which calls `Channel.write()` to send RPC reply messages to the Channel. `RpcDispatcher.channelRead0` is executed in the event loop which means `Channel.write()` writes messages directly to the socket and bypassing the send buffer.
When one side of the RPC channel is making a RPC call but at the same time received another RPC call from the other side. Then this race condition might happen:
- Send call header
- Send reply header
- Send reply payload
- Send call payload
This means RPC reply messages might preempt the on going RPC call and messages will be sent out of order.

To fix this, we have to prevent`RpcDispatcher.channelRead0` from being called while `Rpc.call()` is sending messages. One of the solution is changing Rpc.call()` to write messages in the event loop.

### DIFF
--- a/rsc/src/main/java/com/cloudera/livy/rsc/rpc/Rpc.java
+++ b/rsc/src/main/java/com/cloudera/livy/rsc/rpc/Rpc.java
@@ -216,7 +216,6 @@ public class Rpc implements Closeable {
   private final AtomicLong rpcId;
   private final Channel channel;
   private final EventExecutorGroup egroup;
-  private final Object channelLock;
   private volatile RpcDispatcher dispatcher;
 
   private Rpc(RSCConf config, Channel channel, EventExecutorGroup egroup) {
@@ -224,7 +223,6 @@ public class Rpc implements Closeable {
     Utils.checkArgument(egroup != null);
     this.config = config;
     this.channel = channel;
-    this.channelLock = new Object();
     this.dispatcher = null;
     this.egroup = egroup;
     this.rpcClosed = new AtomicBoolean();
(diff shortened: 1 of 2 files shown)


jerryshao:
@tc0312 can you please elaborate what would be happened without your changes.

rsc/src/main/java/com/cloudera/livy/rsc/rpc/Rpc.java
```diff
@@ -275,10 +273,14 @@ public void operationComplete(ChannelFuture cf) {
       };
 
       dispatcher.registerRpc(id, promise, msg.getClass().getName());
-      synchronized (channelLock) {
```
jerryshao: From my understanding, we already have a lock to avoid race condition, do we need to put into `eventloop`?

tc0312: Behavior of `Channel.write()` is different when it's called inside or outside of the event loop. If it's called in the event loop, it writes to the socket immediately. When it's called outside, it enqueues the write to the send buffer. Event loop will write it to the socket later, 1 message at a time.

This lock doesn't prevent the race condition because `Rpc.call()` and `RpcDispatcher.channelRead0()`.

jerryshao: I think `Channel` is thread safe, it can be called in any thread. The problem here is that we divide the message into two writes, one for header and another for payload, which will mess the ordering of message if called in different threads.

I think there're two solutions to handle it:

1. the one you did, make all the `write` operations execute in one thread, which is channel's `eventloop`. 
2. Another is combining message header and payload together and invoke `write` once. so that we will not get broken message.

The current solution LGTM, thanks for the fix.

tc0312: You are totally right. I weighed the size of solution 1 and 2 and I think 1 is smaller but in the long term, 2 is better.

PR #276, opened 2017-01-12
LIVY-297. Code should be escaped again in R interpreter

jerryshao:
Otherwise the expression cannot be properly unescaped. See issuein LIVY-297.


jerryshao:
CC @zjffdu

zjffdu:
Thanks @jerryshao , Could you also verify the following multiple line R function definition ?, It doesn't work earlier,  not sure whether it is due to same bug. 
```
hello <- function( name ) {
    sprintf( "Hello, %s", name );
}
```

jerryshao:
Yes, it can be worked:

```R
hello <- function( name ) {
    sprintf( "Hello, %s", name );
}
hello("test")
```

Returns `ExecuteSuccess(JObject(List((text/plain,JString([1] "Hello, test")))))`.

zjffdu:
LGTM

PR #277, opened 2017-01-13
LIVY-240. Improve the session gc mechanism

jerryshao:
Current session gc mechanism has some issues:

1. Stopped session still needs to wait to timeout to gc-ed.
2. Batch session will be gc-ed unexpectedly in run-time when timing out, which makes long running application impossible.
3. Sometimes user doesn't want to stop idle sessions.

So here with this patch:

1. Never check the activity of batch session, which means batch session will only be gc-ed after stop.
2. Add a configuration to turn off activity check for interactive session, which meets some usage scenarios.
3. Add a configuration to control how long a finished session state will be kept in memory before cleaned out.

CC @tc0312

### DIFF
--- a/conf/livy.conf
+++ b/conf/livy.conf
@@ -21,6 +21,9 @@
 
 # Time in milliseconds on how long Livy will wait before timing out an idle session.
 # livy.server.session.timeout = 1h
+#
+# How long a finished session state should be kept in LivyServer for query.
+# livy.server.session.state-retain.sec = 600s
 
 # If livy should impersonate the requesting users when creating a new session.
 # livy.impersonation.enabled = true
--- a/conf/livy.conf
+++ b/conf/livy.conf
@@ -16,6 +16,9 @@
 # What spark deploy mode Livy sessions should use.
 # livy.spark.deployMode =
 
+# Enabled to check whether timeout Livy sessions should be stopped.
+# livy.server.session.timeout-check = true
+
 # Time in milliseconds on how long Livy will wait before timing out an idle session.
 # livy.server.session.timeout = 1h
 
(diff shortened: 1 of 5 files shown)


prabhu1984:
Hi @jerryshao,

After this change, are we not going to GC any of the batch session?

jerryshao:
With my change,batch sessions will only be gc-ed when they're finished, no matter it is failed or succeed.

>If yes, then all batch session success/error/dead metadata may sit indefinitely and affects performance.

Can you elaborate more?

tc0312:
> Session will be gc-ed immediately after stop.

Does that mean users can no longer check the batch job status after completion because they will be deleted immediately? How about changing it to start the GC timeout after the session goes to a completed state?

> Never check the activity of batch session, which means batch session will only be gc-ed after stop.

NIce

> Add a configuration to turn off activity check for interactive session, which meet some usage scenarios.

Nice

jerryshao:
>Does that mean users can no longer check the batch job status after completion because they will be deleted immediately? How about changing it to start the GC timeout after the session goes to a completed state?

You mean we don't delete the state immediately, still keep the session state a while after finished, so that user could query the state of session?

tc0312:
Yup.

tc0312:
I really like your change. I think auto cleanup for finished sessions are very useful.
What do you think about the original session cleanup? Deleting running sessions don't really make sense to me. Do you think we should keep it?

jerryshao:
I'm really not sure about user's scenario. From my understanding we could keep this GC mechanism for idle-running sessions but turn off by default, if user want this mechanism, they could set it by true to enable it. Then both sides could be taken care.

tc0312:
Code LGTM.
Mind if you update the commit message to match your latest change?

jerryshao:
@tc0312 which commit message are you pointing to?

tc0312:
Sorry, I mean the Github PR message.

PR #279, opened 2017-01-19
LIVY-303. Add statement retention mechanism

jerryshao:
To avoid OOM for long running sessions, introduce statement retention mechanism to remove old statements.

Also refactor the statement state code to make it more clear.

### DIFF
--- a/repl/src/main/scala/com/cloudera/livy/repl/Session.scala
+++ b/repl/src/main/scala/com/cloudera/livy/repl/Session.scala
@@ -108,17 +108,14 @@ class Session(
 
     Future {
       setJobGroup(statementId)
-      statement.compareAndTransition(StatementState.Waiting, StatementState.Running)
+      statement.compareAndTransit(StatementState.Waiting, StatementState.Running)
 
-      statement.checkStateAndExecute(StatementState.Running,
-        new Runnable {
-          override def run(): Unit = {
-            _statements(statementId).output = executeCode(statementId, code)
-          }
-        })
+      if (statement.state.get() == StatementState.Running) {
+        statement.output = executeCode(statementId, code)
+      }
 
-      statement.compareAndTransition(StatementState.Running, StatementState.Available)
-      statement.compareAndTransition(StatementState.Cancelling, StatementState.Cancelled)
+      statement.compareAndTransit(StatementState.Running, StatementState.Available)
+      statement.compareAndTransit(StatementState.Cancelling, StatementState.Cancelled)
     }(interpreterExecutor)
 
     statementId
@@ -139,27 +136,25 @@ class Session(
       // statement 2 then cancels statement 1. The 2nd cancel call will loop and block the 1st
       // cancel call since cancelExecutor is single threaded. To avoid this, set the statement
       // state to cancelled when cancelling a waiting statement.
-      statement.compareAndTransition(StatementState.Waiting, StatementState.Cancelled)
-      statement.compareAndTransition(StatementState.Running, StatementState.Cancelling)
+      statement.compareAndTransit(StatementState.Waiting, StatementState.Cancelled)
+      statement.compareAndTransit(StatementState.Running, StatementState.Cancelling)
     }
 
     info(s"Cancelling statement $statementId...")
 
     Future {
       val deadline = livyConf.getTimeAsMs(RSCConf.Entry.JOB_CANCEL_TIMEOUT).millis.fromNow
-      val runnable = new Runnable {
-        override def run(): Unit = {
-          if (deadline.isOverdue()) {
-            info(s"Failed to cancel statement $statementId.")
-            statement.compareAndTransition(StatementState.Cancelling, StatementState.Cancelled)
-          } else {
-            _sc.foreach(_.cancelJobGroup(statementId.toString))
-          }
-          Thread.sleep(livyConf.getTimeAsMs(RSCConf.Entry.JOB_CANCEL_TRIGGER_INTERVAL))
+
+      while (statement.state.get() == StatementState.Cancelling) {
+        if (deadline.isOverdue()) {
+          info(s"Failed to cancel statement $statementId.")
+          statement.compareAndTransit(StatementState.Cancelling, StatementState.Cancelled)
+        } else {
+          _sc.foreach(_.cancelJobGroup(statementId.toString))
         }
+        Thread.sleep(livyConf.getTimeAsMs(RSCConf.Entry.JOB_CANCEL_TRIGGER_INTERVAL))
       }
 
-      while (statement.checkStateAndExecute(StatementState.Cancelling, runnable)) { }
       if (statement.state.get() == StatementState.Cancelled) {
         info(s"Statement $statementId cancelled.")
       }
--- a/repl/src/main/scala/com/cloudera/livy/repl/Session.scala
+++ b/repl/src/main/scala/com/cloudera/livy/repl/Session.scala
@@ -151,7 +151,9 @@ class Session(
           statement.compareAndTransit(StatementState.Cancelling, StatementState.Cancelled)
         } else {
           _sc.foreach(_.cancelJobGroup(statementId.toString))
-          Thread.sleep(livyConf.getTimeAsMs(RSCConf.Entry.JOB_CANCEL_TRIGGER_INTERVAL))
+          if (statement.state.get() == StatementState.Cancelling) {
+            Thread.sleep(livyConf.getTimeAsMs(RSCConf.Entry.JOB_CANCEL_TRIGGER_INTERVAL))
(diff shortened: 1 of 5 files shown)


repl/src/main/scala/com/cloudera/livy/repl/Session.scala
```diff
@@ -245,4 +263,24 @@ class Session(
...
   }
+
+  private def cleanOldStatements(): Unit = {
+    if (_statements.size > numRetainedStatements) {
+      // Remove 10% of existing statements to avoid frequently calling this method when reaching
+      // threshold.
+      val toRemove = (numRetainedStatements * 0.1).toInt + 1
+      val stmtIdToRemove = math.max(newStatementId.get() - _statements.size, 0)
+      (stmtIdToRemove until stmtIdToRemove + toRemove).foreach { id =>
+        _statements.get(id).foreach { st =>
```
tc0312: This looks quite expensive. Is it better to replace the ConcurrentHashMap with LinkedHashMap? Then we can iterate the HashMap in `id` order. Also, statement submission doesn't really need high concurrency.

rsc/src/main/java/com/cloudera/livy/rsc/driver/StatementState.java
```diff
@@ -37,4 +44,43 @@
...
+  static {
+    final Map<StatementState, List<StatementState>> predecessors =
+      new EnumMap<>(StatementState.class);
+    put(Waiting, predecessors);
+    put(Running, predecessors, Waiting);
+    put(Available, predecessors, Running);
+    put(Cancelling, predecessors, Running);
+    put(Cancelled, predecessors, Waiting, Cancelling);
+
+    PREDECESSORS = Collections.unmodifiableMap(predecessors);
```
tc0312: Kudos for adding state check.

repl/src/main/scala/com/cloudera/livy/repl/Session.scala
```diff
@@ -117,38 +125,48 @@ class Session(
...
+            Thread.sleep(livyConf.getTimeAsMs(RSCConf.Entry.JOB_CANCEL_TRIGGER_INTERVAL))
+          } catch {
+            case _: InterruptedException => // Ignore interrupted exception.
+            case NonFatal(e) => throw e
+          }
         }
-      }(cancelExecutor)
+      }
+
+      while(_statements(statementId).checkStateAndExecute(StatementState.Cancelling, runnable)) { }
```
jerryshao: This `while` loop doesn't change the semantics compared to original code. It is because we introduce bunch of state transition code, so it is updated to keep align.

repl/src/main/scala/com/cloudera/livy/repl/Session.scala
jerryshao: It may not be so expensive as hashmap operation is `O(1)`, so the total cost would be `n * O(1)`. Yes `LinkedHashMap` is a better solution for this, let me try to solve this.

jerryshao: It is not high concurrency I'm concerned about, it is inconsistency I worried. If we use `LinkedHashMap` without lock, then we may potentially hit the threading issue (though occasionally). I cannot figure out a better solution like concurrent linked hash map.

tc0312: How bad is the lock contention if we use LinkedHashMap?

jerryshao: My concern is that we should add explicit lock in many different places, this makes the current code hard to read. My original purpose is not high concurrency, sorry about the confuse, let me try to replace with {{LinkedHashMap}}

rsc/src/main/java/com/cloudera/livy/rsc/RSCConf.java
```diff
@@ -75,7 +75,9 @@
     TEST_STUCK_START_DRIVER("test.do_not_use.stuck_start_driver", false),
 
     JOB_CANCEL_TRIGGER_INTERVAL("job_cancel.trigger_interval", "100ms"),
-    JOB_CANCEL_TIMEOUT("job_cancel.timeout", "30s");
+    JOB_CANCEL_TIMEOUT("job_cancel.timeout", "30s"),
+
+    RETAINED_STATEMENT_NUMBER("retained_statements", 100);
```
jerryshao: @tc0312 is it enough to set 100 as a default value, or should we increase the value like 1000?

repl/src/main/scala/com/cloudera/livy/repl/Session.scala
```diff
@@ -258,21 +258,17 @@ class Session(
     executeCode(statementId, cmd)
   }
 
-  private def cleanOldStatements(): Unit = {
+  private def cleanOldStatements(): Unit = _statements.synchronized {
```
tc0312: Since we are using LinkedHashMap, I can we can implement [removeEldestEntry()](https://docs.oracle.com/javase/8/docs/api/java/util/LinkedHashMap.html#removeEldestEntry-java.util.Map.Entry-).

jerryshao: Ohh, I see, will change it.

rsc/src/main/java/com/cloudera/livy/rsc/driver/Statement.java
```diff
@@ -36,4 +36,21 @@ public Statement(Integer id, StatementState state, String output) {
   public Statement() {
     this(null, null, null);
   }
+
+  public boolean compareAndTransition(final StatementState from, final StatementState to) {
```
tc0312: Nit: Rename it to compareAndTransit

rsc/src/main/java/com/cloudera/livy/rsc/driver/Statement.java
```diff
@@ -36,4 +36,21 @@ public Statement(Integer id, StatementState state, String output) {
...
+    if (state.compareAndSet(from, to)) {
+      StatementState.validate(from, to);
+      return true;
+    }
+    return false;
+  }
+
+  public boolean checkStateAndExecute(final StatementState s, final Runnable runnable) {
```
tc0312: This function creates a pretense that the function is atomic but it is't.
I would rather expose the state directly to Session. The code is shorter too.

repl/src/main/scala/com/cloudera/livy/repl/Session.scala
```diff
@@ -77,82 +87,83 @@ class Session(
...
+
+      while (statement.state.get() == StatementState.Cancelling) {
+        if (deadline.isOverdue()) {
+          info(s"Failed to cancel statement $statementId.")
+          statement.compareAndTransit(StatementState.Cancelling, StatementState.Cancelled)
+        } else {
+          _sc.foreach(_.cancelJobGroup(statementId.toString))
         }
-      }(cancelExecutor)
+        Thread.sleep(livyConf.getTimeAsMs(RSCConf.Entry.JOB_CANCEL_TRIGGER_INTERVAL))
```
tc0312: LGTM. One last thing, avoid sleeping if statement state has changed to Cancelled. Will merge after fixing this.

repl/src/main/scala/com/cloudera/livy/repl/Session.scala
```diff
@@ -151,8 +151,8 @@ class Session(
           statement.compareAndTransit(StatementState.Cancelling, StatementState.Cancelled)
         } else {
           _sc.foreach(_.cancelJobGroup(statementId.toString))
+          Thread.sleep(livyConf.getTimeAsMs(RSCConf.Entry.JOB_CANCEL_TRIGGER_INTERVAL))
```
alex-the-man:
```
if (statement.state.get() == StatementState.Cancelling) {
  Thread.sleep(livyConf.getTimeAsMs(RSCConf.Entry.JOB_CANCEL_TRIGGER_INTERVAL))
}
```

jerryshao: I think in most of the case code will jump into this `if` (statementState is `Cancelling`). Because `cancelJobGroup`  is an async operation, the return of this function doesn't mean job is cancelled, mostly it will still be `Cancelling`. So IMO it is not so necessary to add this `if` to check.

alex-the-man: Yes it's async but we are also calling it in a loop. Just trying to improve the best case latency here. If you feel strongly about this, I will not insist on the change.

jerryshao: OK, I see, sounds resonable. Let me update the code.

PR #283, opened 2017-01-30
Livy for Spark2.1

linchan-ms:
We hit https://issues.apache.org/jira/browse/SPARK-10872 while running unittests against Spark 2.1. At this point, Spark is not fixing it. In tests, we have to explicitly delete the derby db directory to workaround the issue. Also, need to increase JVM memory footprint with this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,11 @@ language: scala
 
 env:
   - MVN_FLAG='-Pspark-1.6 -DskipTests'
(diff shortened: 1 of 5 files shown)


zjffdu:
@linchan-ms I enabled hive in MiniCluster before, we might need to disable it now.

jerryshao:
Agree with @zjffdu , disabling hive would be a better fix.

 I could propose another PR if you don't have time to handle it.

PR #285, opened 2017-02-03
LIVY-310. Kinit should be called before doing recovery check

jerryshao:
Otherwise, the check will be failed with Kerberos exception.

CC @tc0312 @zjffdu .


tc0312:
Thanks. Are there plans to add integration test for Kerberos?

jerryshao:
@tc0312 it is not a trivial work to support kerberos for integration test, currently I don't have a plan on it.

@zjffdu , can you please take a look, thanks a lot.

zjffdu:
LGTM, yeah I don't think it is easy to test kerberos. @jerryshao did you manually test it ?

jerryshao:
Yes.

tc0312:
I'm merging this now but we really need integration test for Kerberos.

PR #286, opened 2017-02-08
LIVY-311. Set classpath in MiniCluster only when it is not in real cluster

zjffdu:
I am integrating livy MiniCluster to zeppelin to do integration test. So I am not using real cluster, the following assert will fail for me if it is outside the if block. In this PR, I move it into if block

```
assert(classPathFile.isFile,
        s"Cannot read MiniCluster classpath file: ${classPathFile.getCanonicalPath}")
```


zjffdu:
@tc0312 Please help review.

PR #287, opened 2017-02-13
LIVY-308. Avoid enabling Hive by default in Spark2

jerryshao:
In the current implementation of Livy Spark2 SparkSession, by default hive support will be enabled if not explicitly disabled, this is a default behavior in spark-shell. But in Livy interpreter we will disable it by default, otherwise it will introduce unit test failure.

This is another solution of issue #283 .

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1002,6 +1002,18 @@
         </property>
       </activation>
       <properties>
+        <spark.version>2.0.1</spark.version>
+      </properties>
+    </profile>
+
+    <profile>
+      <id>spark-2.1</id>
+      <activation>
+        <property>
+          <name>spark-2.1</name>
+        </property>
+      </activation>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -1002,7 +1002,7 @@
         </property>
       </activation>
       <properties>
-        <spark.version>2.0.1</spark.version>
+        <spark.version>2.1.0</spark.version>
```
tc0312: We should keep running test on Spark 2.0.
Would you mind adding a new profile for Spark 2.1, like #283?

jerryshao: So we will have 3 profiles, spark-1.6, spark-2.0, and spark-2.1?

tc0312:
LGTM. Will merge when test passes.

PR #289, opened 2017-02-14
MINOR. A small REST API typo in the README

ajbozarth:
README incorrectly said `logs` instead of `log` for the `sessions` REST API call, it is correct for the `batches` call below.


zjffdu:
Thanks for contribution. @ajbozarth

PR #292, opened 2017-02-15
LIVY-316. NPE when starting kinit thread

zjffdu:
Looks like it caused by #285, Fix it in this PR.

### DIFF
--- a/server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
+++ b/server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
@@ -101,7 +101,7 @@ class LivyServer extends Logging {
       )
       val launch_keytab = livyConf.get(LAUNCH_KERBEROS_KEYTAB)
       val launch_principal = SecurityUtil.getServerPrincipal(
-        livyConf.get(LAUNCH_KERBEROS_PRINCIPAL), server.host)
+        livyConf.get(LAUNCH_KERBEROS_PRINCIPAL), host)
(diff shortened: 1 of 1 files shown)


server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
```diff
@@ -101,7 +101,7 @@ class LivyServer extends Logging {
       )
       val launch_keytab = livyConf.get(LAUNCH_KERBEROS_KEYTAB)
       val launch_principal = SecurityUtil.getServerPrincipal(
-        livyConf.get(LAUNCH_KERBEROS_PRINCIPAL), server.host)
+        livyConf.get(LAUNCH_KERBEROS_PRINCIPAL), host)
```
jerryshao: AFAIK kerberos requires FQDN, and here host is "0.0.0.0" by default, should we use a hostname instead?

zjffdu: `SecurityUtil` will do this

jerryshao: Ohh, I see, just noticed this. Looks like this is introduced in my previous patch.

tc0312:
We really need some test coverage on Kerberos.

jerryshao:
Let me spend some time on Hadoop's MiniKdc, to see if we could incorporate this to Livy's integration test.

PR #295, opened 2017-02-17
LIVY-312. Fix missing parent poms related issues

jerryshao:
Due to support of different Scala versions with same source code, we introduced some parent poms without adding to main pom. This will introduce several issues described in [LIVY-312](https://issues.cloudera.org/browse/LIVY-312). The main issue is that when using Livy API as dependency, maven will be failed to resolve the parent modules.

So here by adding these missing parent poms to main pom to fix this issue. The tricky thing is that we don't want to build this parent pom (this will bring in compilation issue), so inspired by [here](http://stackoverflow.com/questions/1437718/configure-maven-plugins-to-stick-together/1437887#1437887) and [here](http://stackoverflow.com/questions/1501294/maven-skip-parent-project-build), I use the way mentioned to skip compile and test phase for these parent poms. 

These passed all the UTs and ITs in my local verification, also dependency resolving is correct.

CC @zjffdu .

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -68,20 +68,4 @@
     </dependency>
   </dependencies>
 
-  <build>
-    <plugins>
-      <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-jar-plugin</artifactId>
-        <executions>
-          <execution>
-            <goals>
-              <goal>test-jar</goal>
-            </goals>
-          </execution>
-        </executions>
-      </plugin>
-    </plugins>
-  </build>
-
 </project>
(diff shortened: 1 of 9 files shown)


zjffdu:
Thanks @jerryshao , Verify it with zeppelin, it works perfectly

jerryshao:
@zjffdu , are you going to make some modules publish-able to repo, like repl, server?

zjffdu:
Yeah, we can do that in another PR, mind to create ticket for it ? After livy goes into apache incubation, we need to do a lot change on pom file I think.

jerryshao:
OK, let me create a JIRA to track this.

PR #296, opened 2017-02-17
LIVY-182. Add SPARK_HOME/conf to classpath when SPARK_CONF_DIR is not…

jonalter:
… defined

[LIVY-182](https://issues.cloudera.org/browse/LIVY-182)


tc0312:
Mind if I ask what's this change for?
We are planning to add multiple SPARK_HOME supports. We might need to do this in another way so it works with #232.

jonalter:
You are correct. Also, #232 fixes this issue (see line 207 here: https://github.com/cloudera/livy/pull/232/files#diff-bc70dcf9c8f35cb4efd2f347bb004a9fR193)
Closing this PR and Ticket.

PR #297, opened 2017-02-18
LIVY-293: Redirect Spark Log to REST Response Log field for interactive sessions

praveenkanamarlapudi:
Redirect Spark Log to REST Response Log field 

Task Link: https://issues.cloudera.org/browse/LIVY-293

### DIFF
--- a/rsc/src/main/java/com/cloudera/livy/rsc/ContextLauncher.java
+++ b/rsc/src/main/java/com/cloudera/livy/rsc/ContextLauncher.java
@@ -69,10 +69,10 @@ class ContextLauncher {
   private static final String SPARK_ARCHIVES_KEY = "spark.yarn.dist.archives";
   private static final String SPARK_HOME_ENV = "SPARK_HOME";
 
-  static Promise<ContextInfo> create(RSCClientFactory factory, RSCConf conf)
+  static  DriverProcessInfo create(RSCClientFactory factory, RSCConf conf)
       throws IOException {
     ContextLauncher launcher = new ContextLauncher(factory, conf);
-    return launcher.promise;
+    return new DriverProcessInfo(launcher.promise, launcher.child.child);
   }
 
   private final Promise<ContextInfo> promise;
@@ -383,8 +383,6 @@ class ContextLauncher {
     private final Promise<?> promise;
     private final Process child;
     private final Thread monitor;
-    private final Thread stdout;
-    private final Thread stderr;
     private final File confFile;
 
     public ChildProcess(RSCConf conf, Promise<?> promise, Runnable child, File confFile) {
@@ -392,8 +390,6 @@ class ContextLauncher {
       this.promise = promise;
       this.monitor = monitor(child, CHILD_IDS.incrementAndGet());
       this.child = null;
-      this.stdout = null;
-      this.stderr = null;
       this.confFile = confFile;
     }
 
@@ -402,8 +398,6 @@ class ContextLauncher {
       this.conf = conf;
       this.promise = promise;
       this.child = childProc;
-      this.stdout = redirect("stdout-redir-" + childId, child.getInputStream());
-      this.stderr = redirect("stderr-redir-" + childId, child.getErrorStream());
       this.confFile = confFile;
 
       Runnable monitorTask = new Runnable() {
@@ -450,36 +444,11 @@ class ContextLauncher {
     }
 
     public void detach() {
-      if (stdout != null) {
-        stdout.interrupt();
-        try {
-          stdout.join(conf.getTimeAsMs(CLIENT_SHUTDOWN_TIMEOUT));
-        } catch (InterruptedException ie) {
-          LOG.info("Interrupted while waiting for child stdout to finish.");
-        }
-      }
-      if (stderr != null) {
-        stderr.interrupt();
-        try {
-          stderr.join(conf.getTimeAsMs(CLIENT_SHUTDOWN_TIMEOUT));
-        } catch (InterruptedException ie) {
-          LOG.info("Interrupted while waiting for child stderr to finish.");
-        }
-      }
-
       try {
         monitor.join(conf.getTimeAsMs(CLIENT_SHUTDOWN_TIMEOUT));
       } catch (InterruptedException ie) {
         LOG.debug("Interrupted before driver thread was finished.");
       }
-    }
-
-    private Thread redirect(String name, InputStream in) {
-      Thread thread = new Thread(new Redirector(in));
-      thread.setName(name);
-      thread.setDaemon(true);
-      thread.start();
-      return thread;
     }
 
     private Thread monitor(final Runnable task, int childId) {
--- a/rsc/src/main/java/com/cloudera/livy/rsc/ContextLauncher.java
+++ b/rsc/src/main/java/com/cloudera/livy/rsc/ContextLauncher.java
@@ -69,7 +69,7 @@ class ContextLauncher {
   private static final String SPARK_ARCHIVES_KEY = "spark.yarn.dist.archives";
   private static final String SPARK_HOME_ENV = "SPARK_HOME";
 
-  static  DriverProcessInfo create(RSCClientFactory factory, RSCConf conf)
+  static DriverProcessInfo create(RSCClientFactory factory, RSCConf conf)
       throws IOException {
     ContextLauncher launcher = new ContextLauncher(factory, conf);
     return new DriverProcessInfo(launcher.promise, launcher.child.child);
@@ -304,33 +304,6 @@ class ContextLauncher {
     return file;
   }
 
-  private static class Redirector implements Runnable {
-
-    private final BufferedReader in;
-
-    Redirector(InputStream in) {
-      this.in = new BufferedReader(new InputStreamReader(in));
-    }
-
-    @Override
-    public void run() {
-      try {
-        String line = null;
-        while ((line = in.readLine()) != null) {
-          LOG.info(line);
-        }
-      } catch (Exception e) {
-        LOG.warn("Error in redirector thread.", e);
-      }
-
-      try {
-        in.close();
-      } catch (IOException ioe) {
-        LOG.warn("Error closing child stream.", ioe);
-      }
-    }
-
-  }
 
   private class RegistrationHandler extends BaseProtocol
     implements RpcServer.ClientCallback {
--- a/rsc/src/main/java/com/cloudera/livy/rsc/DriverProcessInfo.java
+++ b/rsc/src/main/java/com/cloudera/livy/rsc/DriverProcessInfo.java
@@ -1,0 +1,43 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.cloudera.livy.rsc;
+
+import io.netty.util.concurrent.Promise;
+
+/**
+ * Information about driver process and @{@link ContextInfo}
+ */
+public class DriverProcessInfo {
+
+  private Promise<ContextInfo> contextInfo;
+  private transient Process driverProcess;
+
+
+  public DriverProcessInfo(Promise<ContextInfo> contextInfo, Process driverProcess) {
+    this.contextInfo = contextInfo;
+    this.driverProcess = driverProcess;
+  }
+
+  public Promise<ContextInfo> getContextInfo() {
+    return contextInfo;
+  }
+
+  public Process getDriverProcess() {
+    return driverProcess;
+  }
+}
--- a/rsc/src/main/java/com/cloudera/livy/rsc/DriverProcessInfo.java
+++ b/rsc/src/main/java/com/cloudera/livy/rsc/DriverProcessInfo.java
@@ -27,7 +27,6 @@ public class DriverProcessInfo {
   private Promise<ContextInfo> contextInfo;
   private transient Process driverProcess;
 
-
   public DriverProcessInfo(Promise<ContextInfo> contextInfo, Process driverProcess) {
     this.contextInfo = contextInfo;
     this.driverProcess = driverProcess;
--- a/rsc/src/main/java/com/cloudera/livy/rsc/RSCClient.java
+++ b/rsc/src/main/java/com/cloudera/livy/rsc/RSCClient.java
@@ -60,12 +60,14 @@ public class RSCClient implements LivyClient {
   private final Promise<URI> serverUriPromise;
 
   private ContextInfo contextInfo;
+  private transient Process driverProcess;
   private volatile boolean isAlive;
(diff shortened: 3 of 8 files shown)


jerryshao:
Is it possible to still keep the original log based way while adding this? I think the original way is still useful for debugging later on.

praveenkanamarlapudi:
Reading twice from the same buffer may not be possible. 
We may consider an approach, which can log both the interactive and batch.

zjffdu:
Agree with @jerryshao , persist the log to file would be useful. Otherwise it would be difficult to debug if the client didn't get any message due to some unknown issues.

praveenkanamarlapudi:
I agree with @jerryshao, but currently, we are not persisting the logs to file for batch. We may add logic to persist for both batch and interactive.

zjffdu:
@praveenkanamarlapudi If I read this PR correctly, it would apply to both interactive session and batch session, right ? AFAIK, the log of interactive session will be persisted to `livy-server.out`.

praveenkanamarlapudi:
@zjffdu Yes, currently interactive sessions log is persisted to `livy-server.out` through two threads ([stdout](https://github.com/cloudera/livy/blob/f5dea605bc53ed6ee420e0cc66a3e17457ef6613/rsc/src/main/java/com/cloudera/livy/rsc/ContextLauncher.java#L405) and [stderr](https://github.com/cloudera/livy/blob/f5dea605bc53ed6ee420e0cc66a3e17457ef6613/rsc/src/main/java/com/cloudera/livy/rsc/ContextLauncher.java#L406)) but log won't come in livy REST response. 

And this PR removes those threads and makes the interactive session's driveProcess handling more like batch.

Sorry for not being clear and missed to mention below point in earlier comments.

I thinks we can still persist the logs of both batch and interactive by enabling logs level to `trace` as we are logging in [trace level](https://github.com/cloudera/livy/blob/91058d29846a82b9d006390a6521bcba11bf56bd/server/src/main/scala/com/cloudera/livy/utils/LineBufferedStream.scala#L42).

rsc/src/main/java/com/cloudera/livy/rsc/ContextLauncher.java
```diff
@@ -69,10 +69,10 @@
   private static final String SPARK_ARCHIVES_KEY = "spark.yarn.dist.archives";
   private static final String SPARK_HOME_ENV = "SPARK_HOME";
 
-  static Promise<ContextInfo> create(RSCClientFactory factory, RSCConf conf)
+  static  DriverProcessInfo create(RSCClientFactory factory, RSCConf conf)
```
tc0312: Nit: Remove extra space between static and DriverProcessInfo .

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
```diff
@@ -376,9 +377,17 @@ class InteractiveSession(
 
   private val app = mockApp.orElse {
     if (livyConf.isRunningOnYarn()) {
+      val driverProcess = client.map { client =>
```
tc0312: val driverProcess = Option(client.map(_.getDriverProcess)).map(p => new LineBufferedProcess(p))

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
```diff
@@ -78,7 +78,8 @@ class SparkYarnApp private[utils] (
   private var yarnDiagnostics: IndexedSeq[String] = IndexedSeq.empty[String]
 
   override def log(): IndexedSeq[String] =
-    process.map(_.inputLines).getOrElse(ArrayBuffer.empty[String]) ++ yarnDiagnostics
+    process.map(_.inputLines).getOrElse(ArrayBuffer.empty[String]) ++ yarnDiagnostics ++
```
tc0312: I think we should print yarnDiagnostics after errorLines because it's changing. Can we also add some formatting to separate logs from stdout/stderr/yarn?

jerryshao:
@praveenkanamarlapudi  I would prefer to still keep the original behavior of log output while adding this. Using trace log is too verbose and usually don't enable this by default.

praveenkanamarlapudi:
@jerryshao I agree with you. How about changing the [logging level](https://github.com/cloudera/livy/blob/91058d29846a82b9d006390a6521bcba11bf56bd/server/src/main/scala/com/cloudera/livy/utils/LineBufferedStream.scala#L42) from `trace` to `debug`? This will enable logging for both batch and interactive sessions.

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
```diff
@@ -377,14 +377,8 @@ class InteractiveSession(
...
     if (livyConf.isRunningOnYarn()) {
-      val driverProcess = client.map { client =>
-        if (client.getDriverProcess == null) {
-          None
-        }
-        else {
-          Some(new LineBufferedProcess(client.getDriverProcess))
-        }
-      }.getOrElse(None)
+      val driverProcess = Option(client.map(_.getDriverProcess).get)
```
tc0312: My bad. Could you do `Option(client.flatMap(_.getDriverProcess))` instead?

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
```diff
@@ -78,8 +78,9 @@ class SparkYarnApp private[utils] (
   private var yarnDiagnostics: IndexedSeq[String] = IndexedSeq.empty[String]
 
   override def log(): IndexedSeq[String] =
-    process.map(_.inputLines).getOrElse(ArrayBuffer.empty[String]) ++ yarnDiagnostics ++
-      process.map(_.errorLines).getOrElse(ArrayBuffer.empty[String])
+    process.map(_.inputLines).getOrElse(ArrayBuffer.empty[String]) ++
```
tc0312: Unnecessary indentation?

```
  override def log(): IndexedSeq[String] =
    "stdout: " ++  process.map(_.inputLines).getOrElse(ArrayBuffer.empty[String]) ++
    "\nstderr: " ++ process.map(_.errorLines).getOrElse(ArrayBuffer.empty[String]) ++
    "\nYARN Diagnostics: " yarnDiagnostics
```

tc0312:
Are you suggesting changing [this](https://github.com/praveenkanamarlapudi/livy/blob/fac14c4973cc8e0bd4a03afcc0e51fd14a8d5d5a/server/src/main/scala/com/cloudera/livy/utils/LineBufferedStream.scala#L42) to `debug()`?
If @jerryshao wants to keep the current behaviour, we might have to change it to `info()`. I'm fine with either way.

server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala
praveenkanamarlapudi: My bad, forgot to add it.

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
praveenkanamarlapudi: Is it
`client.flatMap(c => Option(c.getDriverProcess)).map(new LineBufferedProcess(_))` ?

alex-the-man:
LGTM. @jerryshao?

rsc/src/main/java/com/cloudera/livy/rsc/DriverProcessInfo.java
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Information about driver process and @{@link ContextInfo}
+ */
+public class DriverProcessInfo {
+
+  private Promise<ContextInfo> contextInfo;
+  private transient Process driverProcess;
+
+
```
jerryshao: nit. additional line.

rsc/src/main/java/com/cloudera/livy/rsc/RSCClient.java
```diff
@@ -60,12 +60,14 @@
   private final Promise<URI> serverUriPromise;
 
   private ContextInfo contextInfo;
+  private transient Process driverProcess;
```
jerryshao: Wha't the purpose of transient here?

server/src/main/scala/com/cloudera/livy/server/interactive/InteractiveSession.scala
```diff
@@ -376,9 +377,11 @@ class InteractiveSession(
 
   private val app = mockApp.orElse {
     if (livyConf.isRunningOnYarn()) {
+      val driverProcess = client.flatMap(c => Option(c.getDriverProcess))
```
jerryshao: nit. `flatMap { xxx }`.

server/src/main/scala/com/cloudera/livy/utils/LineBufferedStream.scala
```diff
@@ -39,7 +39,7 @@ class LineBufferedStream(inputStream: InputStream) extends Logging {
       for (line <- lines) {
         _lock.lock()
         try {
-          trace("stdout: ", line)
+          info("stdout: ", line)
```
jerryshao: Can we copy this `line` and make the `info` out of lock. My concern is that log is too heavy inside the lock.

rsc/src/main/java/com/cloudera/livy/rsc/RSCClient.java
praveenkanamarlapudi: I think we can omit transient here.

rsc/src/main/java/com/cloudera/livy/rsc/DriverProcessInfo.java
praveenkanamarlapudi: 👍

server/src/main/scala/com/cloudera/livy/utils/LineBufferedStream.scala
praveenkanamarlapudi: I think it's a good idea to move logging out of lock.

jerryshao:
@praveenkanamarlapudi your change doesn't print out the message in yarn client or local mode, can you please fix it.

praveenkanamarlapudi:
@jerryshao Sure. Let me check it and update you.

jerryshao:
Another problem of this PR when running in local mode is that it will block the spark application, here is the stack of Spark application, the whole application is blocked by log4j appender.

This PR makes us very hard to debug on local mode, since the driver log are gone. More severely this will block driver to flush logs in local mode. So I would suggest to revert this PR until we properly fix it.

CC @alex-the-man  @zjffdu please help on it.

```
"driver-heartbeater" #69 daemon prio=5 os_prio=31 tid=0x00007fe19f27e000 nid=0xa203 waiting for monitor entry [0x0000700003c71000]
   java.lang.Thread.State: BLOCKED (on object monitor)
	at org.apache.log4j.Category.callAppenders(Category.java:204)
	- waiting to lock <0x000000078000fe80> (a org.apache.log4j.spi.RootLogger)
	at org.apache.log4j.Category.forcedLog(Category.java:391)
	at org.apache.log4j.Category.log(Category.java:856)
	at org.slf4j.impl.Log4jLoggerAdapter.warn(Log4jLoggerAdapter.java:479)
	at org.apache.spark.internal.Logging$class.logWarning(Logging.scala:87)
	at org.apache.spark.rpc.RpcEndpointRef.logWarning(RpcEndpointRef.scala:30)
	at org.apache.spark.rpc.RpcEndpointRef.askWithRetry(RpcEndpointRef.scala:111)
	at org.apache.spark.executor.Executor.org$apache$spark$executor$Executor$$reportHeartBeat(Executor.scala:538)
	at org.apache.spark.executor.Executor$$anon$1$$anonfun$run$1.apply$mcV$sp(Executor.scala:567)
	at org.apache.spark.executor.Executor$$anon$1$$anonfun$run$1.apply(Executor.scala:567)
	at org.apache.spark.executor.Executor$$anon$1$$anonfun$run$1.apply(Executor.scala:567)
	at org.apache.spark.util.Utils$.logUncaughtExceptions(Utils.scala:1951)
	at org.apache.spark.executor.Executor$$anon$1.run(Executor.scala:567)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:308)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(ScheduledThreadPoolExecutor.java:180)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:294)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)

  "heartbeat-receiver-event-loop-thread" #65 daemon prio=5 os_prio=31 tid=0x00007fe19d28d000 nid=0x9a03 waiting for monitor entry [0x0000700003866000]
     java.lang.Thread.State: BLOCKED (on object monitor)
  	at org.apache.spark.scheduler.TaskSchedulerImpl.executorHeartbeatReceived(TaskSchedulerImpl.scala:407)
  	- waiting to lock <0x0000000784caa648> (a org.apache.spark.scheduler.TaskSchedulerImpl)
  	at org.apache.spark.HeartbeatReceiver$$anonfun$receiveAndReply$1$$anon$2$$anonfun$run$2.apply$mcV$sp(HeartbeatReceiver.scala:129)
  	at org.apache.spark.util.Utils$.tryLogNonFatalError(Utils.scala:1283)
  	at org.apache.spark.HeartbeatReceiver$$anonfun$receiveAndReply$1$$anon$2.run(HeartbeatReceiver.scala:128)
  	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
  	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
  	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(ScheduledThreadPoolExecutor.java:180)
  	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:293)
  	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
  	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
  	at java.lang.Thread.run(Thread.java:745)
```

praveenkanamarlapudi:
I am checking now and will update you in few hours. Sorry for the trouble.

praveenkanamarlapudi:
@jerryshao I am working on fixing it. I can see logs missing for client and local mode for interactive. But I am not able to reproduce the above error. Can you please give more information about the error?

jerryshao:
If you create a Spark job with many tasks, for example:

```
data = {
  'code': textwrap.dedent("""
    val NUM_SAMPLES = 100000;
    val count = sc.parallelize(1 to NUM_SAMPLES, NUM_SAMPLES).map { i =>
      val x = Math.random();
      val y = Math.random();
      if (x*x + y*y < 1) 1 else 0
    }.reduce(_ + _);
    println(\"Pi is roughly \" + 4.0 * count / NUM_SAMPLES)
    """)
}
```

Then tasks will be blocked or hang, it is easy to reproduce if you're using local mode.

jerryshao:
Besides, do you really need so many spark logs redirect to REST APIs. I think for cluster mode the Spark process output is limited, but for client mode there will be many logs, cache so many logs in Livy for REST request is not a good idea, please be considered when you're improving it. We should not only considered yarn cluster mode (though it is the recommended one), but we should make sure other modes are not broken.

praveenkanamarlapudi:
Sure, let me check it out. Thanks for the detailed info. 👍

praveenkanamarlapudi:
@jerryshao Shall we make cacheing the logs configuration driven?

jerryshao:
I'm not clearly sure, would be better to discuss based on your PR.

PR #300, opened 2017-02-24
LIVY-287. Add Deprecation to Livy Configurations and update naming 

ajbozarth:
[LIVY-287](https://issues.cloudera.org/browse/LIVY-287)

### Updates to Livy configurations
- Added config deprecation with alternatives to `ClientConf`, `HTTPConf`, `RSCConf`, and `LivyConf`
- Added framework for deprecation without alternatives when the need arises
- Updated naming conventions in code and templates to use `-` instead of `_` or camelCase and deprecated previous configs
- Updated `TestClientConf` and added a new test

### Points of discussion
- I tried to optimize `ClientConf.get()` to be quick, not "clean" (was this a good idea?)
- Should I actually add code to deprecate without alternatives? (We don't have any cases yet)
- Are the new naming conventions what we want? They were the original reason for this pr.
- Should I add a test for logging the deprecation message? I don't think testing logs is appropriate for unit tests plus it would take a good chunk of new code to allow it. At the same time the logging is one of the key additions of this code.

### Tests
Passes tests. Tested manually, but manual testing by others is appreciated.


jerryshao:
@ajbozarth I'm afraid due to your changes lot's of things will be broken. Some vendor products already shipped with Livy, it requires lots of efforts to be compatible with next version of Livy. Personally I like this change, but it requires additional efforts to support it and relates to other projects like Ambari, so can we have a compromise solution to this change?

zjffdu:
agree with @jerryshao , keep compatibility is the first priority. I don't think it worth to do this at this stage to break the compatibility.

ajbozarth:
@jerryshao @zjffdu I did think about that when I was working on this, but given @alex-the-man created the JIRA I figured it was worth addressing now before Livy starts looking at an eventual v1.0. @alex-the-man what's your take since you created the JIRA?

On a separate note, what do you think of parts 2 and 3? Are those still worth it even if we abandon part 1?

jerryshao:
Part 2 or 3 LGTM. 

I do agree that conf name should be unified, but this change impacts lots of related projects. Can we have a grey update like Spark to mark them as deprecated, not remove them.

jerryshao:
Also IMHO can we separate 2 and 3 to another PR?

ajbozarth:
I'll split it apart when I have time later today, I'll leave this one for part 1 and open a new pr for part 2/3

As for part 1, your comments made me think about the amount of change (I picked camelCase out of personal preference) and I looked into which of the three options (camelCase, dash, and underscore) were currently used the most and switching the camelCase actually is the most code change by far. In LivyConf about half use dash, a third use underscore and a few use camelCase (not including confs that don't use any). Also all the livy client confs use underscore currently. So if would be best to switch to either dash or underscore instead of camelCase. 

On the above I will try out a grey update like you mentioned for this when I split apart the prs and will also switch away from camelCase

ajbozarth:
I've looked into it and I'm going to add the ability to deprecate confs in order to accomplish this, but since it will take me some time I'm going to close this for now and reopen it when I have a fix ready.

I've opened #301 to address parts 2 and 3

ajbozarth:
I'm reopening this for discussing my conf deprecation implementation.

@jerryshao and @zjffdu I'm looking for some feedback on a few things:
1. Do we need conf deprecation for both `LivyConf` and `ClientConf` (HTTP and RSC)?
2. Should the deprecation list be unified across all three or specific to each? 2b. If separate, would implementing separate scala (for LivyConf) and java (for HTTP and RSC) solutions be frowned upon?
3. What are your opinions on potentially making `ClientConf` a scala file (keeping current java compatibility of course)

I'll hold off any implementation until we get a consensus on this

jerryshao:
AFAIK, all the different confs inside Livy is based on `ClientConf` except python-api. I think we could add some common logics in `ClientConf`.

As for the detailed implement of `LivyConf` and `HttpConf`, we could put deprecation list separately belongs to different implementation.

ajbozarth:
@jerryshao that's what I was thinking, I'm currently looking at some sort of interface, static `Map` and access methods in `ClientConf` that each extension of `ClientConf` then overrides, each keeping their own deprecation list. It's just taking me some time to figure out a solid java implementation for this since I've been primarily writing in scala for the past few years.

ajbozarth:
@jerryshao @zjffdu I've pushed my WIP code and updated the PR description above with my current status. I'll be busy with meetings most of tomorrow but plan to start on my remaining TODO's Friday. I'd love feedback on my implementation choices and the points of discussion in the summary above.

ajbozarth:
Whoops, my line length bar on my editor was set wrong, also I found out style check doesn't running with `mvn package`

ajbozarth:
Pushed rest of the non-test updates, I will be writing new tests at the start of next week

ajbozarth:
A couple oddities I found I'd like input on:

1. On line 103 in `InteractiveSession.scala` the property `livy.client.sessionId` is set and seemly never referenced. Should this be updated to `livy.client.session-id` and if so where should it be deprecated? It doesn't match the naming conventions of `RSCConf.java` where its eventual stored.
2. On line 235 in `LivyServer.scala` `livy.server.serverUrl` is added to `sys.props` and in `MiniCluster.scala` on line 300 it's read. Also on line 179 it's saved to a separate conf file. Similarly should the naming be changed to `livy.server.server-url` and if so where would we deprecate it? It only stored in `sys.props`

In both cases I believe they should be left alone, but I want outside confirmation on that decision.

ajbozarth:
Thanks @alex-the-man I'll push those changes when I finish writing the tests.

Also @alex-the-man @zjffdu and @jerryshao While I'm finishing up writing the tests I'd love some feedback on the *Points of discussion* I listed in the summary above, thanks

ajbozarth:
I've pushed the remaining code and updated the summary, including new discussion points. This is a reasonably large code change and I'll very much like input on any decisions I made.

ajbozarth:
@alex-the-man @zjffdu @jerryshao just checking in if you got a chance to look at this, if not when do you think you may have time? Just don't want this to get buried in the backlog. (note I have a list of open questions in the summary)

client-http/src/main/java/com/cloudera/livy/client/http/HttpConf.java
```diff
@@ -80,4 +84,51 @@ private Entry(String key, Object dflt) {
   boolean isSpnegoEnabled() {
     return getBoolean(Entry.SPNEGO_ENABLED);
   }
+
+  private static final Map<String, DeprecatedConf> configsWithAlternatives
+    = new HashMap<String, DeprecatedConf>() {{
```
jerryshao: Can we make this map immutable using `Collections.unmodifiableMap(xxx)`?

client-common/src/main/java/com/cloudera/livy/client/common/ClientConf.java
```diff
@@ -193,4 +212,40 @@ private void check(boolean test, String message, Object... args) {
...
+    DeprecatedConf depConfs = getDeprecatedConfigs().get(key);
+    if (depConfs != null) {
+      LOG.warn("The configuration key " + depConfs.key() + " has been deprecated as of Livy "
+        + depConfs.version() + " and may be removed in the future. "
+        + depConfs.deprecationMessage());
+    }
+  }
+
+  /** Maps valid key to DeprecatedConf with the deprecated key. */
+  public abstract Map<String, DeprecatedConf> getConfigsWithAlternatives();
```
jerryshao: Do we need to expose these interfaces? I think `protected` should be enough.

client-common/src/main/java/com/cloudera/livy/client/common/ClientConf.java
```diff
@@ -193,4 +212,40 @@ private void check(boolean test, String message, Object... args) {
...
+
+  /** Maps deprecated key to DeprecatedConf with the same key. */
+  public abstract Map<String, DeprecatedConf> getDeprecatedConfigs();
```
jerryshao: I think we could combine `Alternatives` and `Deprecated` map together, using a special `DeprecatedConf` to make this configuration as removable without alternatives. That will make the code a little simpler.

client-http/src/main/java/com/cloudera/livy/client/http/HttpConf.java
ajbozarth: Good idea, will do

client-common/src/main/java/com/cloudera/livy/client/common/ClientConf.java
ajbozarth: You are correct, I'll fix it

client-common/src/main/java/com/cloudera/livy/client/common/ClientConf.java
ajbozarth: I'm not sure it would make it simpler, I think it would just move the complexity. Rather than two lists with different meanings you would instead have one list with two different types of content. I think the code required to combine the lists would be more complex than separate lists. The first indicator of this I see is that the lists are hashed on different values; one is the deprecated key the other is the new key. A second point would be their different usage, the alternates list is used to link old confs to their replacements, whereas the deprecated list is solely for logging warnings if set. If this was all scala code I would agree that combining would be simpler, but I can't see how in Java. I'm still open doing it if you could point out a simple code change to combine them, I admit I may just be too close to see it.

client-common/src/main/java/com/cloudera/livy/client/common/ClientConf.java
```diff
@@ -193,4 +212,40 @@ private void check(boolean test, String message, Object... args) {
...
+        + depConfs.version() + " and may be removed in the future. "
+        + depConfs.deprecationMessage());
+    }
+  }
+
+  /** Maps valid key to DeprecatedConf with the deprecated key. */
+  protected abstract Map<String, DeprecatedConf> getConfigsWithAlternatives();
+
+  /** Maps deprecated key to DeprecatedConf with the same key. */
+  protected abstract Map<String, DeprecatedConf> getDeprecatedConfigs();
```
jerryshao: Regarding this deprecated config. I'm just wondering can we have a special implementation like:

```java

abstract class RemovableConf implements DeprecatedConf {

String key() { return null; }

}
```

Also this makes it possible to combine two maps together, we can differentiate the type of value to differentiate whether it is removable or has alternatives.

client-http/src/main/java/com/cloudera/livy/client/http/HttpConf.java
```diff
@@ -80,4 +85,51 @@ private Entry(String key, Object dflt) {
...
+      put(HttpConf.Entry.JOB_INITIAL_POLL_INTERVAL.key, DepConf.JOB_INITIAL_POLL_INTERVAL);
+      put(HttpConf.Entry.JOB_MAX_POLL_INTERVAL.key, DepConf.JOB_MAX_POLL_INTERVAL);
+  }};
+
+  // Maps deprecated key to DeprecatedConf with the same key.
+  // There are no deprecated configs without alternatives currently.
+  private static final Map<String, DeprecatedConf> deprecatedConfigs = new HashMap<>();
+
+  public Map<String, DeprecatedConf> getConfigsWithAlternatives() {
+    return Collections.unmodifiableMap(configsWithAlternatives);
```
jerryshao: Wrap here in unnecessary, I think you should put this code in L93.

Also I think you should change the access modifier to `protected` here and below.

integration-test/src/main/scala/com/cloudera/livy/test/framework/MiniCluster.scala
```diff
@@ -176,7 +176,7 @@ object MiniLivyMain extends MiniClusterBase {
     // server. Do it atomically since it's used by MiniCluster to detect when the Livy server
     // is up and ready.
     eventually(timeout(30 seconds), interval(1 second)) {
-      val serverUrlConf = Map("livy.server.serverUrl" -> server.serverUrl())
```
jerryshao: Can you please change here to use `LivyConf.xxx.key` rather than string here.

server/src/main/scala/com/cloudera/livy/LivyConf.scala
```diff
@@ -166,6 +170,37 @@ object LivyConf {
     "spark.yarn.jars"
   )
 
+  // TODO: Add Conf Deprecation
+  case class DepConf(override val key: String,
+                     override val version: String,
```
jerryshao: 4 space indent here.

server/src/main/scala/com/cloudera/livy/LivyConf.scala
```diff
@@ -250,4 +285,12 @@ class LivyConf(loadDefaults: Boolean) extends ClientConf[LivyConf](null) {
     Option(get(entry)).map(_.split("[, ]+").toSeq).getOrElse(Nil)
   }
 
+  override def getConfigsWithAlternatives: JMap[String, DeprecatedConf] = {
+    JCollections.unmodifiableMap(configsWithAlternatives.asJava)
```
jerryshao: Scala `Map` is immutable, is it necessary to wrap again?

client-common/src/main/java/com/cloudera/livy/client/common/ClientConf.java
ajbozarth: I think I see where you're going, but `DeprecatedConf.key()` is the deprecated key, the new key is the string that maps to `DeprecatedConf` in the `configsWithAlternatives` HashMap. This would mean changing how the new key maps to the deprecated key in order to allow deprecated keys with no alternative to exist in the same map since the key in the map would be either `null` or `""` and you couldn't have multiple deprecated confs without alternatives in the Map (by the def of a map).
I'm not saying having one Map isn't possible, but based on how I wrote the `DeprecatedConf` interface it would either require more complex code, or a reworking of the interface. I originally based this implementation on Spark's, which had two separate lists, but if you think having a single list would be better in the long run and @zjffdu or @alex-the-man agree I'm willing to rework it from the bottom up with that goal. I'm not super attached to this particular implementation, I just want to make sure future deprecation is straight forward and simple.

client-http/src/main/java/com/cloudera/livy/client/http/HttpConf.java
ajbozarth: Thanks, I'll change this to `protected`and move the wrap to the map declaration (on all the files)

server/src/main/scala/com/cloudera/livy/LivyConf.scala
ajbozarth: Thanks, I found an example of how this indentation is supposed to work, we really need to clean up of style checker, it seems it misses a lot.

server/src/main/scala/com/cloudera/livy/LivyConf.scala
ajbozarth: you are correct, I misread a scala article and though `asJava` would make it mutable, but it correctly keeps the scala (un)mutability.

integration-test/src/main/scala/com/cloudera/livy/test/framework/MiniCluster.scala
ajbozarth: I left them as is based on @alex-the-man comment, but I agree that we should probable move these into `LivyConf`. I'll create a enum for them.

ajbozarth:
@jerryshao @zjffdu @alex-the-man This has been stalled for a few weeks, any chance you'd have time to review my latest code?

client-common/src/main/java/com/cloudera/livy/client/common/ClientConf.java
jerryshao: Sorry for the delay @ajbozarth . I think my purpose of this comment is to add an special abstract implementation to the removable configs, like I mentioned above. This is a separate question of merging two maps. And with this we could statically identify wether this config is replaceable or removable.

jerryshao:
Generally LGTM. Ping @alex-the-man @zjffdu to take a review.

alex-the-man:
Thank you so much for this. This looks really good. Merging.

PR #301, opened 2017-02-24
LIVY-90. Update conf dir to use templates

ajbozarth:
Split off of PR #300 
[LIVY-90](https://issues.cloudera.org/browse/LIVY-90)
I've also included code addressing comments I left in LIVY-90.

### Updates to Livy configurations
2. I've move all the files in conf/ to *.template files (like in Spark) and added the non-template files to .gitignore. This will help separate future dev changes to these files from personal conf changes.
3. Added a livy-client.conf.template to conf/ dir with some of the default values found in HTTPConf and RSCConf similar to livy.conf and LivyConf

### Points of discussion
2. Will switching to templates cause issue not found by tests? (I'm still new to Livy and need help with this)
3. What should livy-client.conf.template look like, I've filled it in as best as I understand the conf values and left out ones I think don't need to be in the template, but I'd like more input

### Tests
Passes current tests and no new tests seem required. Tested manually, but manual testing by others is appreciated.


ajbozarth:
@jerryshao @zjffdu I've split this off of #300 and it's no longer WIP, feedback is appreciated

jerryshao:
I think we should filter out some non-template files under conf when packaging (https://github.com/cloudera/livy/blob/master/assembly/assembly.xml#L21).

jerryshao:
Putting `client-conf` here will lead user to think that changing `client-conf` is enough for Livy client application, but AFAIK this is not done automatically, user should add this file to Livy client application's classpath and set `loadDefault` to true to make it work, I'm afraid this will bring in misleading to users.

ajbozarth:
@jerryshao I'm not sure we should filter the conf dir. Spark doesn't filter the conf dir and I've seen many Spark use cases where users will do a package build to include their custom confs in a jar, for instance to distribute to slave nodes in their cluster.

And for the client conf, from what I could tell from the code currently`loadDefault` defaults to true and is only set to false in certain tests, though since it is a public constructor that could change in future version.

As for the conf location, after checking the code you are correct, it looks for the file in the class path, not necessarily the conf dir. @zjffdu what exactly were you looking for with this JIRA? You seemed to want a `livi-client.conf.template` but as we've found it doesn't seem to make sense keeping it in the conf dir, where do you think it belongs?

jerryshao:
@ajbozarth , I'm fine with keeping some customized confs.

If we want to put `livy-client.conf.template` under the folder, I think we should write something to notify user to put this into classpath if they want to leverage it, not simply change the file here.  Also we should make sure our UT doesn't honor this file in the tests.

ajbozarth:
@jerryshao I updated `livy-client.conf.template` with a description and instructions about the classpath. As for tests, from what I can find, none of them load from `conf/` indiscriminately, they specify the conf file names or use `LivyConf`'s defaults.

jerryshao:
LGTM, cc @alex-the-man  and @zjffdu .

PR #304, opened 2017-03-07
LIVY-322. Catch JsonParseExceptions in PythonInterpreter on rawtext response from fake_shell.

ghost:
subprocess.call() commands in a PySpark snippet can potentially insert raw text into the sys_stdout in the fake_shell main(). This will then fail to be correctly parsed by PythonInterpreter in the sendRequest, as it will trigger a JsonParseException that is not caught. Added code to catch the JsonParseException and then retry reads of stdout until a valid line of JSON is reached, or 100 retries have been attempted.

### DIFF
--- a/repl/src/main/scala/com/cloudera/livy/repl/PythonInterpreter.scala
+++ b/repl/src/main/scala/com/cloudera/livy/repl/PythonInterpreter.scala
@@ -28,6 +28,7 @@ import scala.collection.mutable.ArrayBuffer
 import scala.collection.JavaConverters._
 import scala.util.control.NonFatal
 
+import com.fasterxml.jackson.core.JsonParseException
 import org.apache.spark.{SparkConf, SparkContext}
 import org.json4s.{DefaultFormats, JValue}
 import org.json4s.JsonAST.JObject
@@ -254,7 +255,35 @@ private class PythonInterpreter(process: Process, gatewayServer: GatewayServer, 
     stdin.flush()
 
     Option(stdout.readLine()).map { case line =>
-      parse(line)
+      try {
+        parse(line)
+      }
+      catch {
+        // LIVY-322 - If a statement puts raw text in the stdout without fake_shell parsing it,
+        // retry the readLine up to 100 times in an attempt to find the next parsable line of JSON.
+        case e: JsonParseException => retryRead(100)
+      }
+    }
+  }
+
+  // LIVY-322 - Method for retrying reads from stdout if a JsonParseException is encountered
+  // due to non-JSON formatted text in the stdout. Accepts a param of the number of times
+  // to retry the read, and recurses until either it finds a parsable line of JSON, or
+  // exhausts the maximum level of retries.  If the maximum number of retries is exhausted,
+  // an Exception is thrown.
+  private def retryRead(maxRetries: Int): JValue = {
+    if (maxRetries > 0) {
+      try {
+         parse(stdout.readLine())
+      } catch {
+        case e: JsonParseException =>
+          retryRead(maxRetries - 1)
+      }
+    }
+    else {
+      throw new Exception(
+         "Livy is unable to find valid JSON in the response from fake_shell. " +
+         "Please recreate the session.")
     }
   }
 


jerryshao:
@rickbernotas Can you please fix the title?

jerryshao:
I'm not sure if it is a thorough fix or bandaid for this specific issue, because your scenario is little different.

@zjffdu can you please take a look?

rickbernotas:
Title updated.

alex-the-man:
I can't help but think our current implementation in fake_shell.py should be revamped. I don't think fake_shell.py should use stdout/stderr to communicate with livy-sever.

@jerryshao @zjffdu Do you 2 happen to know how Zeppelin handles this?

alex-the-man:
@rickbernotas I'm so sorry for the late reply. 
Could you educate me on the issue? Do processes started by subprocess.call() inherit the original stdin/out/error? Not the fake one created by fake_shell.py?

rickbernotas:
@alex-the-man I suspect that is what is happening, although I don't know for sure the exact behavior of subprocess.call().  subprocess.call() will return a result (return code like 0 or 1 based on success or failure of the command) but then also starts dumping output into stdout.  The problem is that output from subprocess.call() ends up in this sys_stdout without ever getting caught and parsed to JSON:

https://github.com/cloudera/livy/blob/master/repl/src/main/resources/fake_shell.py#L631

...so you end up with non-JSON formatted rawtext in the flushed response to the PythonInterpreter.

The problem we were having was specific to "hadoop fs -rm" subprocess calls on Hadoop 2.7.  In Hadoop 2.7 the response from those commands includes control characters in the output (like tabs, newlines, etc) that I also suspect might have been a contributing factor.  If you do the same thing on Hadoop 2.8 (which does not have control chars in the response to hadoop fs commands) the subprocess.call() works fine with Livy.  It is as though the combination of how subprocess.call() handles stdout, and the inclusion of control characters in the stdout, creates the problem with the response.  I didn't have time to dig deeper on the issue with the stdout in the fake_shell, which is why I ended up just catching the JsonParseException in the PythonInterpreter.

rickbernotas:
Statement 1 (works fine):

import subprocess

Statement 2 (works fine):

print(1)

Statement 3 (works fine):

subprocess.call(["hadoop", "fs", "-touchz", "foo.tmp"])
print(subprocess.check_output(["hadoop", "fs", "-ls", "foo.tmp"]))

Statement 4 (JsonParseException):

subprocess.call(["hadoop", "fs", "-rm", "foo.tmp"])

Statement 5 (fails to return 1, instead returns nothing, so the responses are messed up going forward after statement 4):

print(1)

-----

This is only reproducible with Hadoop 2.7 due to the formatting of the response from hadoop fs -rm in Hadoop 2.7 (which includes control characters like tab and newline).  In Hadoop 2.8 the formatting of the response from hadoop fs commands changed and you will not get the JsonParseException.  Note that the hadoop fs -touchz also works fine on Hadoop 2.7, but that response is not breaking the fake_shell response like the -rm is.  subprocess.check_output() also works as expected, i.e. if you give the hadoop fs -rm to subprocess.check_output() on Hadoop 2.7, it's okay...but the whole response from subprocess.check_output() is returned as the result as a string, which is different than the behavior of subprocess.call() (which returns a return code as the result, and then dumps the text reponse straight into stdout).  

Theoretically, however, any command run by subprocess.call() that dumps text output straight to stdout, and also includes control characters in the output, will break the parser in the same fashion.

alex-the-man:
Can you change your code to the following just to root cause the problem?
```
subprocess.call(["hadoop", "fs", "-touchz", "foo.tmp"], stdout=open("out", "w"), stderr=open("err", "w"))
```

rickbernotas:
Yes, that works, the problematic line was the hadoop fs -rm, but it does work with that change to the call() and returns 0 as result.  No JsonParseException seen.

subprocess.call(["hadoop", "fs", "-rm", "foo.tmp"], stdout=open("out", "w"), stderr=open("err", "w"))

alex-the-man:
@rickbernotas Cool. Processes started by subprocess.call() inherit the original stdin/out/error then.

Is this workaround good enough for now?

rickbernotas:
Yes, the workarounds are sufficient for us, we were also using subprocess.check_output alternatively (as it is also a workaround).  I'm leaving the contents of the PR in place on my end just because it minimally catches users who go ahead and use subprocess.call anyways, and prevents the JsonParseException.  But if there is a change in how stdin/out/err is handled by the fake_shell in the future I'll definitely be interested in that.

alex-the-man:
I think the proper fix is:
- Modify PythonInterpreter to create a temp input/output files or 2 sets of mkfifo. Pass their path to [ProcessBuilder](https://github.com/rickbernotas/livy/blob/1796ccb6a6c28e334b4be71755778b262d216fd9/repl/src/main/scala/com/cloudera/livy/repl/PythonInterpreter.scala#L59) as a command line arugment.
- Modify fake_shell.py to use the temp input/output files instead of stdin/stdout.
- Modify  [ProcessInterpreter](https://github.com/rickbernotas/livy/blob/1796ccb6a6c28e334b4be71755778b262d216fd9/repl/src/main/scala/com/cloudera/livy/repl/ProcessInterpreter.scala#L48) to send requests and read output from temp input/output files rather than stdin/stdout.

johnchenghk01:
Any update on this issue?

PR #306, opened 2017-03-09
LIVY-326. Change Hadoop dependencies to Apache Hadoop

jerryshao:


pom.xml
```diff
@@ -42,7 +42,7 @@
   </organization>
 
   <properties>
-    <hadoop.version>2.6.0-cdh5.5.0</hadoop.version>
+    <hadoop.version>2.6.5</hadoop.version>
```
zjffdu: How about change it to 2.7.x which has been released for a while

jerryshao: Yes, personally I don't have a strong preference on which version of Hadoop to use, I just keep align with Spark's minimum supported version of Hadoop.

jerryshao:
@alex-the-man do you have any suggestion on which version of Hadoop should be used?

jerryshao:
Looks there's a unrelated flaky test.

alex-the-man:
Since we don't have multiple profiles like Spark has in their pom.xml, I think it's better to use 2.7 so we are running our integration test with the most popular YARN version.

zjffdu:
LGTM, wait for CI pass

PR #307, opened 2017-03-09
LIVY-313. Fixed SparkRInterpreter always returning success.

jonalter:
- Stopped redirecting stderr to stdout.
- Continue to read ErrorStream (it was only being read once).
- Checking for any errors returned by stderr before returning success.

[LIVY-313](https://issues.cloudera.org/browse/LIVY-313)

### DIFF
--- a/repl/src/main/scala/com/cloudera/livy/repl/ProcessInterpreter.scala
+++ b/repl/src/main/scala/com/cloudera/livy/repl/ProcessInterpreter.scala
@@ -109,7 +109,7 @@ abstract class ProcessInterpreter(process: Process)
 
   private[this] val stderrThread = new Thread("process stderr thread") {
     override def run() = {
-      while(true) {
+      while (true) {
         val lines = Source.fromInputStream(process.getErrorStream).getLines()
 
         for (line <- lines) {
--- a/repl/src/main/scala/com/cloudera/livy/repl/ProcessInterpreter.scala
+++ b/repl/src/main/scala/com/cloudera/livy/repl/ProcessInterpreter.scala
@@ -109,14 +109,16 @@ abstract class ProcessInterpreter(process: Process)
 
   private[this] val stderrThread = new Thread("process stderr thread") {
     override def run() = {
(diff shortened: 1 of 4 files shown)


alex-the-man:
Mind if you fix the style issue?

alex-the-man:
There used to be a race condition in the code I fixed in #251. To solve it, I modified the code to redirect stderr to stdout so SparkRInterpreter doesn't need to read from multiple streams. To repo the issue, run  [this unit test](https://github.com/jerryshao/livy/blob/dcd26436e3101945c47cd7c22c0a95b3b1b03fbc/repl/src/test/scala/com/cloudera/livy/repl/SparkRInterpreterSpec.scala#L85) in a loop.



I think this fix undoes my fix.

jonalter:
@alex-the-man thanks for the feedback. Could you tell me how you run a single test or single test suite? I have been doing it by building my own class path and running `org.scalatest.tools.Runner` manually. I imagine there has to be an easier way to do it.

alex-the-man:
@jonalter Do you use ideaj?

jonalter:
@alex-the-man yes I do, but I am pretty new to using it.

alex-the-man:
You can start a specific scalatest test case using the test runner integrated in IntelliJ. You will hit some errors if your environment variables are not set proper. You can set it referencing to [pom.xml](https://github.com/cloudera/livy/blob/master/pom.xml#L707).

However, for your particular case, you can just add a infinite loop around [this block](https://github.com/jerryshao/livy/blob/dcd26436e3101945c47cd7c22c0a95b3b1b03fbc/repl/src/test/scala/com/cloudera/livy/repl/SparkRInterpreterSpec.scala#L86). Let it run overnight to verify the race condition is gone.

jonalter:
@alex-the-man, thanks for the info!

jonalter:
I ran that one test overnight like you suggested with no failure.

jonalter:
PR ready for review

alex-the-man:
LGTM. @zjffdu

zjffdu:
LGTM, thanks @jonalter  will merge it.

PR #308, opened 2017-03-13
LIVY-329. Fix two SSL issues

jerryshao:
1. Livy server url is exposed as http URL even https is enabled, so we should handle this.
2. Livy server SSL keystore password and key password currently set to same configurations, which should be separated.

Verified locally.


jerryshao:
@zjffdu @alex-the-man , please help to review, just simple fix.

PR #310, opened 2017-03-14
LIVY-331. Fix flaky test (interactive session should not gc-ed if session timeout check is off)

jerryshao:
The fix is inspired by [here](http://stackoverflow.com/questions/11121772/when-i-run-mockito-test-occurs-wrongtypeofreturnvalue-exception).

Tried several times locally without the previous issue, will see if this issue still exist in travis test.


jerryshao:
Jacoco failure, retest.

jerryshao:
test again to see if this flaky test still occurs.

jerryshao:
Ran the test on traivs several rounds, I think this flakiness should be gone.

PR #311, opened 2017-03-17
LIVY-249. Update livy-server script to include status command and return 1 on error

ajbozarth:
[LIVY-249](https://issues.cloudera.org/browse/LIVY-249)

Currently if `livy-server` fails to start it will still return `0` (success), this has been fixed. I also added a `livy-server status` command that will return a status output similar to that of `livy-server start` when it's already running or `livy-server stop` when it's already stopped. Lastly I updated all output text to use `livy-server` instead of `livy_server` since that's the actual name of the script.

Tested manually


ajbozarth:
@zjffdu since you wrote most of the code I'm editing

zjffdu:
LGTM

PR #312, opened 2017-04-07
LIVY-323. Fixed livy-repl doesn't start with some spark packages.

alex-the-man:
Some Spark packages are depending on scala-reflect 2.11.0 and it conflicts with Spark's scala version 2.11.8.
This's not Livy's fault actually but doesn't hurt to make Livy more fault tolerant.
Since the scala-reflect jar with the correct version must already be in CLASSPATH, fixed livy-repl to not load user supplied scala-reflect jars.


alex-the-man:
This's a small fix. I will merge if there's no further comments.

zjffdu:
LGTM

PR #313, opened 2017-04-11
Fix repo url in README to be available without login.

tlahn:
Change git clone URL in README so that it is available without being logged in to github.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -72,7 +72,7 @@ Livy is built using `Apache Maven`_. To check out and build Livy, run:
 
 .. code:: shell
 
-    git clone git@github.com:cloudera/livy.git
+    git clone https://github.com/cloudera/livy.git
     cd livy
     mvn package
 


jerryshao:
LTGM, thanks for the fix.

zjffdu:
@tlahn Thanks for contribution, did you sign CCLA.

https://github.com/cloudera/livy/wiki/Contributing-to-Livy

tlahn:
@zjffdu Hi, no I did not sign that.

I do understand that nowadays legal issues are a big concern. But reading, understanding and signing a legal contract in a foreign language with a company from a foreign country (which means I should additionally know that country's laws) and disclosing personal information is really too much risk and effort in comparison to changing a URL. So I will not do that at this moment.

Feel free to use the change and let someone else provide it as a patch. I can also offer to create an issue and add the change as a written proposal in there if that is fine for you. Just let me know.

zjffdu:
Understood, @tlahn , still appreciate your contribution.

jerryshao:
@tlahn I think @zjffdu may not clarify this thing very clearly. 

For lots of open source project like tensorflow, it requires agreement for the individual contributor. If you're not a company paid open source contributor, I think ICLA is enough, not CCLA.

If you have any concern about this, you could send mail to Cloudera, or ping @ksakellis .

tlahn:
@jerryshao Thank you for clarification. I already assumed that ICLA would be the right thing in my case, so that is what I was talking about. For me this is just too much legal stuff for a 10 second patch.

Sorry, I did not intend to make this a big thing. These are your rules and that's fine. So is there another way to get this in easily? Like someone who already signed provides the patch.

jerryshao:
Yes someone else already signed this could help to get this in. But the owner of this commit will be that guy. Is that acceptable for you?

Also since this is a quite tiny fix, we could fix it together next time when we submit other patches.

tlahn:
Yes, that's fine for me. I don't care about who is the author, just wanted it to be fixed.

alex-the-man:
#315 Closing.

PR #316, opened 2017-04-21
Expose session id and job id to the httpclient

xjodoin:
When we use the http client we need to have access to the session id and job id in case of a unexpected shutdown.


jerryshao:
@xjodoin very sorry to be so delay. Can you please resubmit this PR against https://github.com/apache/incubator-livy, so we could have a discussion there. Thanks!

PR #317, opened 2017-04-24
Logging Queries submitted to Livy

pranayhasan:
This PR adds support to Log Queries coming to Livy. It can be enabled using the configuration **livy.server.query-logging.enabled** .

### DIFF
--- a/server/src/main/scala/com/cloudera/livy/LivyConf.scala
+++ b/server/src/main/scala/com/cloudera/livy/LivyConf.scala
@@ -62,9 +62,13 @@ object LivyConf {
 
   val SERVER_HOST = Entry("livy.server.host", "0.0.0.0")
   val SERVER_PORT = Entry("livy.server.port", 8998)
-  val CSRF_PROTECTION = LivyConf.Entry("livy.server.csrf-protection.enabled", false)
 
   val UI_ENABLED = Entry("livy.ui.enabled", true)
+
+  val REQUEST_HEADER_SIZE = Entry("livy.server.request-header.size", 131072)
+  val RESPONSE_HEADER_SIZE = Entry("livy.server.response-header.size", 131072)
+
+  val CSRF_PROTECTION = Entry("livy.server.csrf-protection.enabled", false)
 
   val IMPERSONATION_ENABLED = Entry("livy.impersonation.enabled", false)
   val SUPERUSERS = Entry("livy.superusers", null)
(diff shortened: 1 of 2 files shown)


jerryshao:
I think it is already supported in Livy. Livy could log the incoming request like below, what is the difference of your PR?

```
0:0:0:0:0:0:0:1 - - [01/Mar/2017:10:31:30 +0000] "POST /sessions HTTP/1.1" 201 -
0:0:0:0:0:0:0:1 - - [01/Mar/2017:10:33:42 +0000] "GET /sessions/0 HTTP/1.1" 200 -
0:0:0:0:0:0:0:1 - - [01/Mar/2017:10:33:48 +0000] "POST /sessions/0/statements HTTP/1.1" 201 -
0:0:0:0:0:0:0:1 - - [01/Mar/2017:10:33:55 +0000] "GET /sessions/0/statements/0 HTTP/1.1" 200 -
```

jerryshao:
CC @zjffdu

pranayhasan:
With this PR, when users enable Query Logging they can log the content of the Statements hitting Livy Endpoint. In the above, only the request-response headers are visible. In here, we log the POST requests with IP, remoteUser (if kerberos is enabled), Timestamp in UTC and the code snippet.

Attaching a screenshot of a sample view:
screenshot

jerryshao:
I don't think it is a good idea to dump such detailed request contents to the log, especially in the access controlled environments.

pranayhasan:
Isn't it good to have a log of requests if Livy acts an endpoint serving a lot of users? It shall be helpful in usage monitoring who's accessing what datasets etc in the Hadoop cluster.

pranayhasan:
Also, its a configuration change which enables Livy admin to log queries and shall be false by default.

jerryshao:
1. If you want to know who's accessing Livy, I think previous access log with http head should be enough for most of the scenarios, dumping content is not necessary.
2. If you want to track who's accessing which dataset, then hdfs audit log and CallerContext is much better and accurate. It's quite vague to know the real things only based on request contents.

pranayhasan:
Your recommendation is true just for HDFS logs. We have a requirement to know what queries users ran in case of a security breach on the shared cluster. This requires tracing back what queries were run from Livy by looking at the query logs.

jerryshao:
In the current Livy, only the user who created this session or Livy super user has the permission to get the statements of this session, not to say all sessions. You PR will be a security hole.

zjffdu:
@pranayhasan Are you trying to add audit log for livy like hdfs ?

pranayhasan:
@jerryshao I agree to your point that the Applications logs are world readable and opens a security hole. Is it acceptable to log queries to a new log file where only Livy superusers have access to? I would update the PR incase that works.

@zjffdu Yes, we want Livy to have something similar to HDFS audit log.

jerryshao:
Since we already have access log, adding a new log which has similar functionality seems unnecessary. I would suggest to enrich current `RequestLogHandler` rather than create a new filter (like what you did currently).

First of all you should sign ICLA (if not) and create a JIRA.

pranayhasan:
Thanks @jerryshao and @zjffdu for taking a look. I would sign ICLA and create a JIRA and post it here. I would have to implement a custom RequestLog similar to NCSARequestLog.

PR #318, opened 2017-04-25
LIVY-246 Support multiple Spark home in runtime

ayushiagarwal:
Made code changes  for enabling multiple spark versions on Livy for spark batch jobs. User can pass sparkVersion on run time for batch jobs and SPARK_HOME and SPARK_CONF_DIR would be set  according to given sparkVersion.

Need to set path for livy.server.spark-home_$version in Livy.conf and those versions which are added would be supported. 

Assumption:
* bin/livy-env.sh will continue to have SPARK_HOME and SPARK_CONF_DIR which will be used as default spark version for batch and interactive

### DIFF
--- a/conf/livy-env.sh.template
+++ b/conf/livy-env.sh.template
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+ #!/usr/bin/env bash
 #
 # Licensed to Cloudera, Inc. under one
 # or more contributor license agreements.  See the NOTICE file
--- a/conf/livy-env.sh.template
+++ b/conf/livy-env.sh.template
@@ -1,4 +1,4 @@
- #!/usr/bin/env bash
+#!/usr/bin/env bash
 #
 # Licensed to Cloudera, Inc. under one
 # or more contributor license agreements.  See the NOTICE file
--- a/server/src/main/scala/com/cloudera/livy/LivyConf.scala
+++ b/server/src/main/scala/com/cloudera/livy/LivyConf.scala
@@ -246,14 +246,20 @@ class LivyConf(loadDefaults: Boolean) extends ClientConf[LivyConf](null) {
   def sparkDeployMode(): Option[String] = Option(get(LIVY_SPARK_DEPLOY_MODE)).filterNot(_.isEmpty)
 
   /** Return the location of the spark home directory */
-  def sparkHome(): Option[String] = Option(get(SPARK_HOME)).orElse(sys.env.get("SPARK_HOME"))
+  def sparkHome(version: Option[String] = None): Option[String] = {
+    version.map {version => Option(get(s"livy.server.spark-home_$version"))
+      .orElse(throw new IllegalArgumentException(
+        s"Spark version: $version is not supported"))
+    }.getOrElse(Option(get(s"livy.server.spark-home")).orElse(sys.env.get("SPARK_HOME")))
+  }
 
   /** Return the spark master Livy sessions should use. */
   def sparkMaster(): String = get(LIVY_SPARK_MASTER)
 
   /** Return the path to the spark-submit executable. */
-  def sparkSubmit(): String = {
-    sparkHome().map { _ + File.separator + "bin" + File.separator + "spark-submit" }.get
+  def sparkSubmit(version: Option[String] = None): String = {
+    sparkHome(version).map { _ + File.separator + "bin" + File.separator +
+      "spark-submit" }.get
   }
 
   /** Return the list of superusers. */
--- a/server/src/main/scala/com/cloudera/livy/LivyConf.scala
+++ b/server/src/main/scala/com/cloudera/livy/LivyConf.scala
@@ -245,12 +245,16 @@ class LivyConf(loadDefaults: Boolean) extends ClientConf[LivyConf](null) {
   /** Return the spark deploy mode Livy sessions should use. */
   def sparkDeployMode(): Option[String] = Option(get(LIVY_SPARK_DEPLOY_MODE)).filterNot(_.isEmpty)
 
-  /** Return the location of the spark home directory */
-  def sparkHome(version: Option[String] = None): Option[String] = {
-    version.map {version => Option(get(s"livy.server.spark-home_$version"))
+  /** Return the  spark home version */
+  def SPARK_HOME_VER(version: String): Option[String] = {
+    Option(get(s"livy.server.spark-home-$version"))
       .orElse(throw new IllegalArgumentException(
         s"Spark version: $version is not supported"))
-    }.getOrElse(Option(get(s"livy.server.spark-home")).orElse(sys.env.get("SPARK_HOME")))
+  }
+  /** Return the location of the spark home directory */
+  def sparkHome(version: Option[String] = None): Option[String] = {
+    version.map {version => SPARK_HOME_VER(version)
+    }.getOrElse(Option(get(SPARK_HOME)).orElse(sys.env.get("SPARK_HOME")))
   }
 
   /** Return the spark master Livy sessions should use. */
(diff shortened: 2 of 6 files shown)


server/src/main/scala/com/cloudera/livy/LivyConf.scala
```diff
@@ -246,14 +246,20 @@ class LivyConf(loadDefaults: Boolean) extends ClientConf[LivyConf](null) {
   def sparkDeployMode(): Option[String] = Option(get(LIVY_SPARK_DEPLOY_MODE)).filterNot(_.isEmpty)
 
   /** Return the location of the spark home directory */
-  def sparkHome(): Option[String] = Option(get(SPARK_HOME)).orElse(sys.env.get("SPARK_HOME"))
+  def sparkHome(version: Option[String] = None): Option[String] = {
+    version.map {version => Option(get(s"livy.server.spark-home_$version"))
```
ajbozarth: I think we should abstract out `livy.server.spark-home_$version` here into a def like the val's above. Something along the lines of:
`def SPARK_HOME_VER(version: String) = Entry(s"livy.server.spark-home-$version", null)`

ajbozarth (commented):
I'm not the best to review how well this accomplishes it's purpose, but I've given config related feedback where I can

ayushiagarwal:
Fixed the changes suggested. Please review

server/src/main/scala/com/cloudera/livy/LivyConf.scala
```diff
@@ -245,15 +245,25 @@ class LivyConf(loadDefaults: Boolean) extends ClientConf[LivyConf](null) {
   /** Return the spark deploy mode Livy sessions should use. */
   def sparkDeployMode(): Option[String] = Option(get(LIVY_SPARK_DEPLOY_MODE)).filterNot(_.isEmpty)
 
+  /** Return the  spark home version */
+  def SPARK_HOME_VER(version: String): Option[String] = {
```
ajbozarth: I was thinking more along the lines of adding
`def SPARK_HOME_VER(version: String) = Entry(s"livy.server.spark-home-$version", null)`
to the LivyConf object above as a `def` similar to the `val`'s, then doing `Option(get(SPARK_HOME_VER(version)))` which you could then do an orElse throw Exception on inside the sparkHome def as before.

server/src/main/scala/com/cloudera/livy/LivyConf.scala
```diff
@@ -245,15 +245,25 @@ class LivyConf(loadDefaults: Boolean) extends ClientConf[LivyConf](null) {
...
+    Option(get(s"livy.server.spark-home-$version"))
+      .orElse(throw new IllegalArgumentException(
+        s"Spark version: $version is not supported"))
+  }
   /** Return the location of the spark home directory */
-  def sparkHome(): Option[String] = Option(get(SPARK_HOME)).orElse(sys.env.get("SPARK_HOME"))
+  def sparkHome(version: Option[String] = None): Option[String] = {
+    version.map {version => SPARK_HOME_VER(version)
```
ajbozarth: On a second look it may be better to use a `case` here, assuming the changes I recommend above something like:
```
case Some(version) => Option(get(SPARK_HOME_VER(version))).orElse(throw ...)
case None => Option(get(SPARK_HOME)).orElse(sys.env.get("SPARK_HOME"))
```
where the None case is just the old code. This would be easier to read overall.

ajbozarth (commented):
Thanks for the updates, added some follow up comments

server/src/main/scala/com/cloudera/livy/LivyConf.scala
ayushiagarwal: If I declare  def SPARK_HOME_VER(version: String) = Entry(s"livy.server.spark-home-$version", null) then I cant pass it in def sparkHome(version: Option[String] = None): Option[String] = {
+    version.map {version=>Option(get(SPARK_HOME_VER(version))).orElse(throw ...)} as it is a map and it will not accept null values

ajbozarth: If I'm understanding you correctly, that's where my comment below would fix it. I'll edit the comment below to clarify it.

ayushiagarwal:
Fixed the suggested changes. Please review

ayushiagarwal:
Moved the version with rest of the Entry's and also added code to set up  SPARK_CONF_DIR. We need to set up SPARK_CONF_DIR otherwise sparksubmit will take the default environment set up SPAR_CONF_DIR - so setting it up according to spark version.

conf/livy-env.sh.template
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+ #!/usr/bin/env bash
```
ajbozarth: I think this was added by accident

server/src/main/scala/com/cloudera/livy/server/batch/BatchSession.scala
```diff
@@ -62,7 +63,11 @@ object BatchSession {
           request.conf, request.jars, request.files, request.archives, request.pyFiles, livyConf))
       require(request.file != null, "File is required.")
 
-      val builder = new SparkProcessBuilder(livyConf)
+      val builder = new SparkProcessBuilder(livyConf, request.sparkVersion)
+      request.sparkVersion.map({ value =>
```
ajbozarth: You don't need maps here, this should work fine as:
```
val sparkConfDir = livyConf.sparkHome(request.sparkVersion).map(_ + File.separator + "conf")
builder.env("SPARK_CONF_DIR", sparkConfDir)
```

ajbozarth (commented):
Thanks for the fixes. I found one nit and made a comment on the SPARK_CONF_DIR code you added (I understand why it's needed now)

server/src/main/scala/com/cloudera/livy/server/batch/BatchSession.scala
```diff
@@ -62,7 +63,11 @@ object BatchSession {
...
+        builder.env("SPARK_CONF_DIR", livyConf.sparkHome(request.sparkVersion)
```
prabhu1984: Hi @ajbozarth, if we are not setting up SPARK_CONF_DIR, then if SPARK_CONF_DIR is set on environment already, then the older one gets precedence and hence spark version change might not get effect. 
Example on Hortonworks Distribution: in SPARK_CONF_DIR/spark-defaults.conf, then "spark.yarn.jar" is set to "hdfs:///hdp/spark-assembly.1.6.2.jar" and hence if we are not overwriting the SPARK_CONF_DIR, then always this spark-assembly jar will be used for any spark version.

server/src/main/scala/com/cloudera/livy/server/batch/BatchSession.scala
ayushiagarwal: I think we will need map in builder.env as sparkConfDir is optional variable and builder.env declaration needs a string.  Agreed we can remove the first map of  request.sparkVersion. Will make the changes and deploy.

ayushiagarwal:
Fixed the suggested changes in SPARK_CONF_DIR. Please review.

ajbozarth (approved):
LGTM

jerryshao:
What's difference compared to #232 ?

ajbozarth:
@ayushiagarwal want to reopen this on apache/incubator-livy and address @jerryshao concern with his similar PR?

karthikknnatarajan:
@ayushiagarwal @ajbozarth 
We are looking for such a feature in Livy. Any ideas around when this will be merged into master? I can help getting this change merged in.

ajbozarth:
If @ayushiagarwal lets this continue to sit idle for another week I think it would be ok to reopen this on apache/incubator-livy yourself @karthikknnatarajan 
No promises on whether it gets merged though, @jerryshao has an alternate implementation ( #232 ) that he hasn't moved over either

karthikknnatarajan:
Sounds good. I will drop a note here before I create a new PR on apache/incubator-livy.

PR #321, opened 2017-04-28
LIVY-351. Remove the use of guava api in Livy

jerryshao:
This will lead to ClassNotFound issue when running in client mode. Since Spark doesn't have guava dependency.

### DIFF
--- a/rsc/src/main/java/com/cloudera/livy/rsc/driver/StatementState.java
+++ b/rsc/src/main/java/com/cloudera/livy/rsc/driver/StatementState.java
@@ -20,7 +20,6 @@ package com.cloudera.livy.rsc.driver;
 import java.util.*;
 
 import com.fasterxml.jackson.annotation.JsonValue;
-import com.google.common.base.Preconditions;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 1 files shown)


jerryshao:
CC @zjffdu , this blocks the use of local mode in Livy, can you please review it.

zjffdu:
LGTM. I see guava is in hadoop's dependency. yarn client don't run driver in yarn container, that's why we can not use it. Looks like we need to be careful about the different classpath between running driver in yarn-container and non yarn container.

```
[INFO] +- org.apache.hadoop:hadoop-common:jar:2.7.3:provided
[INFO] |  +- org.apache.hadoop:hadoop-annotations:jar:2.7.3:provided
[INFO] |  |  \- jdk.tools:jdk.tools:jar:1.8:system
[INFO] |  +- com.google.guava:guava:jar:11.0.2:provided
[INFO] |  +- commons-cli:commons-cli:jar:1.2:provided
[INFO] |  +- xmlenc:xmlenc:jar:0.52:provided
```

jerryshao:
Yes, that's why it is OK in yarn cluster mode, because guava is in Hadoop's classpath, whereas in local mode this is not true.

PR #322, opened 2017-04-28
LIVY-349. Fix two SparkR test issue regarding to different versions of R

jerryshao:
This is caused by the difference of error output with different versions of R.


jerryshao:
@zjffdu , please review, SparkR UT failure should be fixed now.

zjffdu:
Thanks @jerryshao , will merge it.

PR #323, opened 2017-05-03
LIVY-355. Refactor statement progress tracker to fix binary compatible issue

jerryshao:
Current Livy's `StatementProgressListener` is an implementation of `SparkListener`, while the signature of `SparkListener` is different in Spark1 and Spark2, which means the Livy code compiled with Spark1 cannot be run on Spark2+, vice versa.

So here propose to avoid directly using `SparkListener`, and instead using `SparkStatusTracker` to measure the progress of statement. `SparkStatusTracker` is a stable interface both in Spark1 and Spark2, using this will solve binary compatible issue.


repl/src/main/scala/com/cloudera/livy/repl/Session.scala
```diff
@@ -238,25 +237,4 @@ class Session(
 
     compact(render(resultInJson))
   }
-
-  private def setJobGroup(statementId: Int): String = {
```
jerryshao: Instead of using String statement to set job group we use underlying SparkContext in each interpreter to directly set job group.

jerryshao:
@zjffdu , can you please help to review it.

zjffdu:
LGTM, will merge it

PR #324, opened 2017-05-03
LIVY-356 Enabling LDAP authentication for Client to Server.

jakhani:
Currently, Livy doesn't support LDAP Authentication from client(sparkmagic) to server(livy). We need to add LDAP authentication as that's preferable method due to security reasons. We won't be able to use Knox for this purpose. That is why I am raising this PR which contains LDAP authentication. I have upgraded hadoop.version in livy-main to 2.8.0 as this version contains LivyAuthenticationHandler. Below I have mentioned link for the same:
https://insight.io/github.com/apache/hadoop/blob/HEAD/hadoop-common-project/hadoop-auth/src/main/java/org/apache/hadoop/security/authentication/server/LdapAuthenticationHandler.java


jerryshao:
Can you please create a JIRA and change the title like other PRs.

Also did you sign ICLA? You need to sign a ICLA before contributing, please see the wiki about the details.

jakhani:
@jerryshao I have created jira ticket for this change: https://issues.cloudera.org/browse/LIVY-356

I have already signed CCLA when I joined LinkedIn.

jerryshao:
@jakhani , did you sign CCLA / ICLA and send to cloudera? (https://github.com/cloudera/livy/wiki/Individual-Contributor-License-Agreement-(ICLA)).

Also can you update the PR title to add JIRA number like other PRs.

pom.xml
```diff
@@ -42,7 +42,7 @@
   </organization>
 
   <properties>
-    <hadoop.version>2.7.3</hadoop.version>
+    <hadoop.version>2.8.0</hadoop.version>
```
jerryshao: What is the purpose of using Hadoop 2.8.0, is it mainly for LDAP support?

server/src/main/scala/com/cloudera/livy/LivyConf.scala
```diff
@@ -90,6 +90,13 @@ object LivyConf {
   val KINIT_FAIL_THRESHOLD =
     LivyConf.Entry("livy.server.launch.kerberos.kinit-fail-threshold", 5)
 
+  // Ldap properties
+  val AUTH_LDAP_URL = Entry("livy.server.auth.ldap.url", null)
+  val AUTH_LDAP_BASE_DN = Entry("livy.server.auth.ldap.base.dn", null)
```
jerryshao: I would suggest to change the configuration name to `livy.server.auth.ldap.base-dn`, as well as for the following configurations, would be better to change to `livy.server.auth.ldap.xx-xx-xx`.

server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
```diff
@@ -197,7 +197,27 @@ class LivyServer extends Logging {
           livyConf.get(AUTH_KERBEROS_NAME_RULES))
         server.context.addFilter(holder, "/*", EnumSet.allOf(classOf[DispatcherType]))
         info(s"SPNEGO auth enabled (principal = $principal)")
-
+      case authType@LdapAuthenticationHandler.TYPE =>
```
jerryshao: nit: space between "@".

server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
```diff
@@ -197,7 +197,27 @@ class LivyServer extends Logging {
...
+          holder.setInitParameter(LdapAuthenticationHandler.PROVIDER_URL, url)
+        }
+        val baseDN = livyConf.get(LivyConf.AUTH_LDAP_BASE_DN)
+        if (baseDN != null) {
+          holder.setInitParameter(LdapAuthenticationHandler.BASE_DN, baseDN)
+        }
+        val domain = livyConf.get(LivyConf.AUTH_LDAP_USERNAME_DOMAIN)
+        if (domain != null) {
+          holder.setInitParameter(LdapAuthenticationHandler.LDAP_BIND_DOMAIN, domain)
+        }
```
jerryshao: Does `url`, `domain` and `baseDN` are all required to enable LDAP, or either of them can be null?

Also the code can be simplified as below, no need to do null check like Java.

```
Option(livyConf.get(LivyConf.AUTH_LDAP_USERNAME_DOMAIN)).foreach(holder.setInitParameter(LdapAuthenticationHandler.LDAP_BIND_DOMAIN, _))
```

jerryshao (commented):
I would not upgrade Hadoop version to 2.8.0, since lots of users as well as vendors are not using this version.

If LDAP is the feature in 2.8.0, then probably you'd better using reflection to handle this. Also since `AuthenticationFilter` is more like a container for different authentication methods, we can avoid directly refer to LDAP specific code.

pom.xml
jakhani: Yes it was mainly for LDAP support. I can write my custom class for the same rather than using Hadoop-auth LdapAuthenticationHandler.

server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
jakhani: Either of them can be null but I can't put then in holder if any is null because it internally uses HashTable for env variables. It will throw NPE in this case. But I can simplify code a bit.

pom.xml
jerryshao: It is OK to use Hadoop's implementation, just try to not directly refer to any object in `LdapAuthenticationHandler `, all you need to do is to set proper parameters to `FilterHolder`. And you can build Livy against Hadoop 2.8 your own, then you will have Ldap supported transparently.

PR #328, opened 2017-05-10
Minor. Fix deprecated conf warning log issue

jerryshao:
```
./livy-sshao-server.out.5:17/05/08 16:50:44 WARN LivyConf: The configuration key livy.spark.deployMode has been deprecated as of Livy 0.4 and may be removed in the future. Please use the new key livy.spark.deploy-mode instead.
./livy-sshao-server.out.5:17/05/08 16:50:45 WARN LivyConf: The configuration key livy.spark.scalaVersion has been deprecated as of Livy 0.4 and may be removed in the future. Please use the new key livy.spark.scala-version instead.
./livy-sshao-server.out.5:17/05/08 16:51:04 WARN RSCConf: The configuration key livy.rsc.driver_class has been deprecated as of Livy 0.4 and may be removed in the future. Please use the new key livy.rsc.driver-class instead.
```

This log is incorrect even if we use new configuration key. This is mainly because the logic in `logDeprecationWarning` to check alternative configurations is not correct.

CC @ajbozarth .

### DIFF
--- a/client-common/src/main/java/com/cloudera/livy/client/common/ClientConf.java
+++ b/client-common/src/main/java/com/cloudera/livy/client/common/ClientConf.java
@@ -211,12 +211,11 @@ public abstract class ClientConf<T extends ClientConf>
 
   /** Logs a warning message if the given config key is deprecated. */
   private void logDeprecationWarning(String key) {
-    String altConfKey = allAlternativeKeys().get(key);
-    if (altConfKey != null) {
-      DeprecatedConf altConfs = getConfigsWithAlternatives().get(altConfKey);
-      LOG.warn("The configuration key " + altConfs.key() + " has been deprecated as of Livy "
-        + altConfs.version() + " and may be removed in the future. Please use the new key "
-        + altConfKey + " instead.");
+    ConfPair altConf = allAlternativeKeys().get(key);
+    if (altConf != null) {
+      LOG.warn("The configuration key " + key + " has been deprecated as of Livy "
+        + altConf.depConf.version() + " and may be removed in the future. Please use the new key "
+        + altConf.newKey + " instead.");
       return;
     }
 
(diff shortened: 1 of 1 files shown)


client-common/src/main/java/com/cloudera/livy/client/common/ClientConf.java
```diff
@@ -236,6 +234,18 @@ private void logDeprecationWarning(String key) {
   /** Maps deprecated key to DeprecatedConf with the same key. */
   protected abstract Map<String, DeprecatedConf> getDeprecatedConfigs();
 
+  private Map<String, String> altToNewKeyMap = null;
+  private Map<String, String> allAlternativeKeys() {
```
ajbozarth: If you could add a comment above this to explain its purpose, it took me some time to figure out what it was doing and it will save future devs that same time.

ajbozarth (approved):
Thanks for catching this, I missed that Spark had a function like this that inverted the map for the warning lookup. It might be messier in java but this is a good implementation, just one nit. LGTM

client-common/src/main/java/com/cloudera/livy/client/common/ClientConf.java
```diff
@@ -214,11 +211,12 @@ private void check(boolean test, String message, Object... args) {
 
   /** Logs a warning message if the given config key is deprecated. */
   private void logDeprecationWarning(String key) {
-    DeprecatedConf altConfs = getConfigsWithAlternatives().get(key);
-    if (altConfs != null) {
+    String altConfKey = allAlternativeKeys().get(key);
+    if (altConfKey != null) {
+      DeprecatedConf altConfs = getConfigsWithAlternatives().get(altConfKey);
```
ajbozarth: Actually on second look this line is unnecessary and `altConfs.key()` on the line below can be replaced with `key` now. This function was previously testing the inverse of what it was supposed to.

jerryshao:
@ajbozarth I slightly changed the code, can you review again.

ajbozarth:
I feel like the static class is a bit overkill, but its cleaner than the alternative of `Map<String, Map<String, DepConf>>` so I'll say LGTM

jerryshao:
@ajbozarth , actually the structure is not `Map<String, Map<String, DepConf>>`, it is `Map<String, Tuple<String, DepConf>>`, because Java doesn't have built-in tuple, so I have to write a class instead.

jerryshao:
Moving to incubator-livy repo.

PR #329, opened 2017-05-12
LIVY-358. Make JettyServer Http request and response header size configurable

jerryshao:
This patch is based on #266 , mainly to fix header size too large problem in spnego enabled scenario. 

1. Increase the default header size.
2. Make header size configurable.


zjffdu:
LGTM

PR #331, opened 2017-05-15
LIVY-358. [Follow-up] Add unit test to verify header size configurations

jerryshao:
This is a followup of #329 to add unit test to verify header size configurations. CC @zjffdu .

### DIFF
--- a/client-http/src/test/scala/com/cloudera/livy/client/http/LivyConnectionSpec.scala
+++ b/client-http/src/test/scala/com/cloudera/livy/client/http/LivyConnectionSpec.scala
@@ -18,20 +18,20 @@
 
 package com.cloudera.livy.client.http
 
+import java.io.IOException
 import java.net.URLEncoder
 import java.nio.charset.StandardCharsets.UTF_8
-import javax.servlet.http.{HttpServlet, HttpServletRequest, HttpServletResponse}
 
+import scala.util.Random
 import org.apache.http.client.utils.URIBuilder
 import org.eclipse.jetty.security._
 import org.eclipse.jetty.security.authentication.BasicAuthenticator
-import org.eclipse.jetty.server.Server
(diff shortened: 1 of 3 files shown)


zjffdu:
Thanks @jerryshao , sorry I miss to review one thing in #329,  could add those 2 configuration in livy.conf.template ?

jerryshao:
Thanks, I will update the conf file.

client-http/src/test/scala/com/cloudera/livy/client/http/LivyConnectionSpec.scala
```diff
@@ -60,31 +60,29 @@ class LivyConnectionSpec extends FunSpecLike with BeforeAndAfterAll with LivyBas
...
+        .setScheme(server.protocol)
+        .setHost(server.host)
+        .setPort(server.port)
         .setUserInfo(URLEncoder.encode(username, utf8Name), URLEncoder.encode(password, utf8Name))
         .build()
       info(uri.toString)
       val conn = new LivyConnection(uri, new HttpConf(null))
       try {
-        conn.get(classOf[Boolean], "/") shouldBe true
+        conn.get(classOf[Object], "/") should not be (null)
```
zjffdu: why is this changed ?

jerryshao: Because previously we use this [servlet](https://github.com/cloudera/livy/pull/331/files#diff-6ca014ee7958eb376596cb519952a394L63) which will return `Boolean`. Now we change to use `InteractiveSessionServlet`, so using `Boolean` will meet JSON parse error.

zjffdu:
LGTM

jerryshao:
@zjffdu any further comment?

PR #337, opened 2017-05-20
[LIVY-364] livy-server writes one log file when set LIVY_MAX_LOG_FILES to 0

davidxdh:
[https://issues.cloudera.org/browse/LIVY-364](https://issues.cloudera.org/browse/LIVY-364)
When set LIVY_MAX_LOG_FILES to 0, livy-server renames old log file to livy-root-server.out.0 while startup. So there has one log file which is inconsistent with LIVY_MAX_LOG_FILES value 0.

### DIFF
--- a/bin/livy-server
+++ b/bin/livy-server
@@ -49,7 +49,7 @@ pid="$LIVY_PID_DIR/livy-$LIVY_IDENT_STRING-server.pid"
 livy_rotate_log() {
   log=$1
   num=$LIVY_MAX_LOG_FILES
-  if [ -f "$log" ]; then # rotate logs
+  if [ -f "$log" ] && [ $num -gt 0 ]; then # rotate logs
 	while [ $num -gt 1 ]; do
 	  prev=`expr $num - 1`
 	  [ -f "$log.$prev" ] && mv "$log.$prev" "$log.$num"
@@ -57,6 +57,7 @@ livy_rotate_log() {
 	done
     mv "$log" "$log.$num"
   fi
+  rm -f $log
 }
 
 create_dir() {


davidxdh:
@ajbozarth @jerryshao please review this PR. Thanks.

ajbozarth:
From my understanding of the code, `LIVY_MAX_LOG_FILES` is the number of old log files to keep, so the total logs would actually be `LIVY_MAX_LOG_FILES + 1` so if you add a `[ $num -gt 0 ]` check to the first `if` statement here, you could instead set `LIVY_MAX_LOG_FILES=0` to get your desired result.

If we would rather change this so `LIVY_MAX_LOG_FILES` includes all logs not just old logs, then I am ok with this implementation as is. I'll differ to @jerryshao @zjffdu or @alex-the-man on whether that's what we want though.

jerryshao:
I agree with @ajbozarth , I checked Spark code, Spark also follows the same behavior, the actual file number is `MAX_LOG_FILES + 1`. So I think it is not necessary to change this behavior.

davidxdh:
@ajbozarth. I think your advice is reasonable. According to your opinion, I want to do the following things:
a: change content of issue LIVY-362 to the situation when LIVY_MAX_LOG_FILES set to 0.
b: change this PR by adding a [ $num -gt 0 ] check to the first if statement.

ajbozarth (commented):
This LGTM, adding the ability to set it to 0 is nice

davidxdh:
@zjffdu or @alex-the-man please review this PR. Thanks.

jerryshao:
Hi @davidxdh sorry for delay, can you please submit this PR against incubator-livy (https://github.com/apache/incubator-livy/pulls)?

PR #339, opened 2017-05-24
Livy 365. RSC. Include statement code in statement object

ajbozarth:
[LIVY-365](https://issues.cloudera.org/browse/LIVY-365)

Based on a user list request and my work on the Livy Web UI including the code that was run in the statement return is useful. Currently there is no way for a user to check what code was run through the Livy REST API. This will be helpful when the Session Page in the Web UI is added in [LIVY-343](https://issues.cloudera.org/browse/LIVY-343) which will include a table of statements. Currently such a table could only display output without context.

I choose to insert `code` between `id` and `state` because it does not change value during execution like `state` and `output`. Since it can also makes sense to include between `state` and `output` given the potential length of its value, I can make such a change if requested.

Tested manually and passes current Livy UTs and ITs

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -914,6 +914,8 @@ A statement represents the result of an execution statement.
 +========+======================+=====================+
 | id     | The statement id     | integer             |
 +--------+----------------------+---------------------+
+| code   | The execution code   | string              |
++--------+----------------------+---------------------+
 | state  | The execution state  | statement state     |
 +--------+----------------------+---------------------+
 | output | The execution output | statement output    |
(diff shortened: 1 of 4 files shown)


ajbozarth:
@alex-the-man and @jerryshao could you take a look at this? You two wrote most of the surrounding code.

jerryshao:
Can you please fix the compilation error.

ajbozarth:
pushed a fix, going on vacation for the next 3 days so I'll be afk

jerryshao:
LGTM.

jerryshao:
Hi @ajbozarth , sorry for the delay, can you submit this PR against https://github.com/apache/incubator-livy, thanks.

ajbozarth:
moved to https://github.com/apache/incubator-livy/pull/5

PR #343, opened 2017-06-15
LIVY-356 Enabling LDAP authentication for Client to Server.

jakhani:
Currently, Livy doesn't support LDAP Authentication from client(sparkmagic) to server(livy). We need to add LDAP authentication as that's preferable method due to security reasons. We won't be able to use Knox for this purpose. That is why I am raising this PR which contains LDAP authentication.

I am also adding config to disable batch endpoints. Due to security reasons, we don't want to keep batch endpoint open. I am adding config so that We can disable it.

### DIFF
--- a/server/src/main/scala/com/cloudera/livy/LivyConf.scala
+++ b/server/src/main/scala/com/cloudera/livy/LivyConf.scala
@@ -64,6 +64,7 @@ object LivyConf {
   val SERVER_PORT = Entry("livy.server.port", 8998)
 
   val UI_ENABLED = Entry("livy.ui.enabled", true)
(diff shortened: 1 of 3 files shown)


jakhani:
@jerryshao Can you please review this PR? I have discarded old PR for LDAP as this is better solution.

jerryshao:
Sorry @jakhani for the delay, we're working on migrating Livy project to ASF, It would be better to wait for while and send PR against in new repo. Really sorry about the delay.

jerryshao:
Hi @jakhani can you please submit this PR against incubator-livy (https://github.com/apache/incubator-livy/pulls), thanks!

jakhani:
@jerryshao I will raise PR and let you know. Thanks.

